PR #42, opened 2014-12-06
Add a default timeout of 80 seconds for API requests.

metcalf:
The `http.DefaultClient` has no timeout on requests. In our other bindings, we set an 80 second timeout. AFAIK this particular value is tied to a considered funnel of timeouts that works back through our systems.

r? @cosn

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -65,7 +65,7 @@ func SetDebug(value bool) {
 // GetBackend returns the currently used backend in the binding.
 func GetBackend() Backend {
 	if backend == nil {
-		backend = NewInternalBackend(&http.Client{Timeout: 80}, "")
+		backend = NewInternalBackend(&http.Client{Timeout: 80 * time.Second}, "")
 	}
 
 	return backend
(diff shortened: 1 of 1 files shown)


josegrad:
Hi, I got here from this other case

"Lower the value of api request timeout" - https://github.com/stripe/stripe-ruby/issues/46

In the last month I've experienced a couple of timeouts from Stripe which on Heroku take the users to a general error page. In some cases users pay again and we need to refund + manually fix some things.

If there is a default timeout now, can we also specify a custom one? I would set it to 25 or so to avoid the 30 seconds timeout from Heroku.

I kind of guess the custom timeout is not possible, but just making sure...

cosn:
You can build your own `httpClient` which includes any timeout value desired:

https://github.com/stripe/stripe-go/blob/master/stripe.go#L74

josegrad:
Sorry I didn't notice his wasn't ruby, though it was referenced in a ruby case. I'll delete this and my previous comment from this issue later.

PR #55, opened 2015-01-07
Added Bitcoin Receiver & Transaction objects; support multiple charge sources

maxwells:
Added Bitcoin Receiver & Transaction objects

### DIFF
--- a/bitcoin_receiver.go
+++ b/bitcoin_receiver.go
@@ -1,0 +1,64 @@
+package stripe
+
+import "encoding/json"
+
+// BitcoinReceiverListParams is the set of parameters that can be used when listing BitcoinReceivers.
+// For more details see https://stripe.com/docs/api/#list_bitcoin_receivers.
+type BitcoinReceiverListParams struct {
+	ListParams
+	NotFilled  bool
+	NotActive  bool
+	Uncaptured bool
+}
+
+// BitcoinReceiverParams is the set of parameters that can be used when creating a BitcoinReceiver.
+// For more details see https://stripe.com/docs/api/#create_bitcoin_receiver.
+type BitcoinReceiverParams struct {
+	Params
+	Amount   uint64
+	Currency Currency
+	Desc     string
+	Email    string
+	Meta     map[string]string `json:"metadata"`
+}
+
+// BitcoinReceiver is the resource representing a Stripe bitcoin receiver.
+// For more details see https://stripe.com/docs/api/#bitcoin_receivers
+type BitcoinReceiver struct {
+	ID                    string                  `json:"id"`
+	Created               int64                   `json:"created"`
+	Currency              Currency                `json:"currency"`
+	Amount                uint64                  `json:"amount"`
+	AmountReceived        uint64                  `json:"amount_received"`
+	BitcoinAmount         uint64                  `json:"bitcoin_amount"`
+	BitcoinAmountReceived uint64                  `json:"bitcoin_amount_received"`
+	Filled                bool                    `json:"filled"`
+	Active                bool                    `json:"active"`
+	RejectTransactions    bool                    `json:"reject_transactions"`
+	Desc                  string                  `json:"description"`
+	InboundAddress        string                  `json:"inbound_address"`
+	RefundAddress         string                  `json:"refund_address"`
+	BitcoinUri            string                  `json:"bitcoin_uri"`
+	Meta                  map[string]string       `json:"metadata"`
+	Email                 string                  `json:"email"`
+	Payment               string                  `json:"payment"`
+	Customer              string                  `json:"customer"`
+	Transactions          *BitcoinTransactionList `json:"transactions"`
+}
+
+// UnmarshalJSON handles deserialization of a BitcoinReceiver.
+// This custom unmarshaling is needed because the resulting
+// property may be an id or the full struct if it was expanded.
+func (br *BitcoinReceiver) UnmarshalJSON(data []byte) error {
+	type bitcoinReceiver BitcoinReceiver
+	var r bitcoinReceiver
+	err := json.Unmarshal(data, &r)
+	if err == nil {
+		*br = BitcoinReceiver(r)
+	} else {
+		// the id is surrounded by "\" characters, so strip them
+		br.ID = string(data[1 : len(data)-1])
+	}
+
+	return nil
+}
--- a/bitcoin_receiver.go
+++ b/bitcoin_receiver.go
@@ -19,7 +19,6 @@ type BitcoinReceiverParams struct {
 	Currency Currency
 	Desc     string
 	Email    string
-	Meta     map[string]string `json:"metadata"`
 }
 
 // BitcoinReceiver is the resource representing a Stripe bitcoin receiver.
@@ -44,10 +43,6 @@ type BitcoinReceiver struct {
 	Payment               string                  `json:"payment"`
 	Customer              string                  `json:"customer"`
 	Transactions          *BitcoinTransactionList `json:"transactions"`
-}
-
-func (br BitcoinReceiver) PaymentType() string {
-	return "bitcoin_receiver"
 }
 
 // UnmarshalJSON handles deserialization of a BitcoinReceiver.
--- a/bitcoin_receiver.go
+++ b/bitcoin_receiver.go
@@ -46,6 +46,10 @@ type BitcoinReceiver struct {
 	Transactions          *BitcoinTransactionList `json:"transactions"`
 }
 
+func (br BitcoinReceiver) PaymentType() string {
+	return "bitcoin_receiver"
+}
+
 // UnmarshalJSON handles deserialization of a BitcoinReceiver.
 // This custom unmarshaling is needed because the resulting
 // property may be an id or the full struct if it was expanded.
--- a/bitcoin_receiver.go
+++ b/bitcoin_receiver.go
@@ -9,19 +9,16 @@ import (
 // For more details see https://stripe.com/docs/api/#list_bitcoin_receivers.
 type BitcoinReceiverListParams struct {
 	ListParams
-	NotFilled  bool
-	NotActive  bool
-	Uncaptured bool
+	NotFilled, NotActive, Uncaptured bool
 }
 
 // BitcoinReceiverParams is the set of parameters that can be used when creating a BitcoinReceiver.
 // For more details see https://stripe.com/docs/api/#create_bitcoin_receiver.
 type BitcoinReceiverParams struct {
 	Params
-	Amount   uint64
-	Currency Currency
-	Desc     string
-	Email    string
+	Amount      uint64
+	Currency    Currency
+	Desc, Email string
 }
 
 // BitcoinReceiver is the resource representing a Stripe bitcoin receiver.
@@ -48,7 +45,7 @@ type BitcoinReceiver struct {
 	Transactions          *BitcoinTransactionList `json:"transactions"`
 }
 
-// Human readable/displayable way of inspecting a BitcoinReceiver
+// Display human readable representation of a BitcoinReceiver.
 func (br *BitcoinReceiver) Display() string {
 	var filled string
 	if br.Filled {
--- a/bitcoin_receiver.go
+++ b/bitcoin_receiver.go
@@ -1,6 +1,9 @@
 package stripe
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"fmt"
+)
 
 // BitcoinReceiverListParams is the set of parameters that can be used when listing BitcoinReceivers.
 // For more details see https://stripe.com/docs/api/#list_bitcoin_receivers.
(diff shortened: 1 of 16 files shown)


maxwells:
WIP

maxwells:
r? @zenazn 
cc @cosn @raycmorgan 

Happy to chat context/concerns IRL!

bitcoin_receiver.go
```diff
@@ -0,0 +1,68 @@
...
+
+// BitcoinReceiverParams is the set of parameters that can be used when creating a BitcoinReceiver.
+// For more details see https://stripe.com/docs/api/#create_bitcoin_receiver.
+type BitcoinReceiverParams struct {
+	Params
+	Amount   uint64
+	Currency Currency
+	Desc     string
+	Email    string
+	Meta     map[string]string `json:"metadata"`
```
cosn: I think this is already on Params

bitcoin_receiver.go
```diff
@@ -0,0 +1,68 @@
...
+	RefundAddress         string                  `json:"refund_address"`
+	BitcoinUri            string                  `json:"bitcoin_uri"`
+	Meta                  map[string]string       `json:"metadata"`
+	Email                 string                  `json:"email"`
+	Payment               string                  `json:"payment"`
+	Customer              string                  `json:"customer"`
+	Transactions          *BitcoinTransactionList `json:"transactions"`
+}
+
+func (br BitcoinReceiver) PaymentType() string {
```
cosn: Comment?

card.go
```diff
@@ -142,6 +142,10 @@ func (c *CardParams) AppendDetails(values *url.Values, creating bool) {
 	}
 }
 
+func (c Card) PaymentType() string {
```
cosn: Comment?

payment_source.go
```diff
@@ -0,0 +1,17 @@
+package stripe
```
cosn: Can we please add comments for this whole file?

payment_source.go
```diff
@@ -0,0 +1,17 @@
...
+	Card  *CardParams
+}
+
+type PaymentSource struct {
+	ID     string `json:"id"`
+	Object string `json:"object"`
+}
+
+type Chargeable interface {
+	PaymentType() string
```
cosn: This isn't being invoked for anywhere ... who is this meant for?

payment_source.go
```diff
@@ -0,0 +1,17 @@
+package stripe
+
+type SourceParams struct {
+	Params
+	ID    string
+	Token string
+	Card  *CardParams
+}
+
+type PaymentSource struct {
```
cosn: I don't see this being used anywhere

cosn:
Also I'm assuming `make test` passed after these changes?

Version should be bumped as well (4.4.0) and we should add an entry in the changelog

zenazn:
@cosn — okay if you're the primary reviewer on this? You know this code a lot better than me.

bitcoin_receiver.go
```diff
@@ -0,0 +1,79 @@
...
+// property may be an id or the full struct if it was expanded.
+func (br *BitcoinReceiver) UnmarshalJSON(data []byte) error {
+	type bitcoinReceiver BitcoinReceiver
+	var r bitcoinReceiver
+	err := json.Unmarshal(data, &r)
+	if err == nil {
+		*br = BitcoinReceiver(r)
+	} else {
+		// the id is surrounded by "\" characters, so strip them
+		br.ID = string(data[1 : len(data)-1])
```
zenazn: This isn't right—the part between the `"`'s might contain escape characters or similar. It's better to `json.Unmarshal` into a string type and let the JSON package do all the hard work for you.

I get the feeling that this code was dutifully cargo-culted from somewhere else, though, so perhaps this deserves to be fixed in a separate PR.

cosn:
yeah happily

payment_source.go
```diff
@@ -0,0 +1,67 @@
...
+	Display() string
+}
+
+// PaymentSource describes the payment source used to make a Charge.
+// The Type should indicate which object is fleshed out (eg. BitcoinReceiver or Card)
+// For more details see https://stripe.com/docs/api#retrieve_charge
+type PaymentSource struct {
+	Type            string `json:"object"`
+	ID              string `json:"id"`
+	Card            *Card
```
zenazn: should these have `json:"-"`?

payment_source.go
```diff
@@ -0,0 +1,67 @@
...
+func (s *PaymentSource) UnmarshalJSON(data []byte) error {
+	type source PaymentSource
+	var ss source
+	err := json.Unmarshal(data, &ss)
+	if err == nil {
+		*s = PaymentSource(ss)
+
+		switch s.Type {
+		case paymentsource.BitcoinReceiver:
+			s.BitcoinReceiver = &BitcoinReceiver{}
```
zenazn: I'm pretty sure `encoding/json` will initialize these for you—no need to do it by hand.

paymentsource/types.go
```diff
@@ -0,0 +1,6 @@
+package paymentsource
+
+const (
+	BitcoinReceiver = "bitcoin_receiver"
```
zenazn: should these be typed constants? (do they need to be in a separate package?)

payment_source.go
```diff
@@ -0,0 +1,67 @@
...
+
+// Interface to provide a human readable representation of a struct
+type Displayer interface {
+	Display() string
+}
+
+// PaymentSource describes the payment source used to make a Charge.
+// The Type should indicate which object is fleshed out (eg. BitcoinReceiver or Card)
+// For more details see https://stripe.com/docs/api#retrieve_charge
+type PaymentSource struct {
```
zenazn: I take it you ended up doing this over the `interface` strategy?

maxwells: Yeah -- I didn't really love that in order to access the common properties ("id" & "object"), end users would need a type assertion or we would provide a bunch of redundant getters to satisfy the interface. I also got to move the unmarshaling of the Card/BitcoinReceiver structs into the PaymentSource itself out of Charge, which ~broke some level of encapsulation (https://github.com/stripe/stripe-go/commit/20d58c60d5bac016bb3a13d5775443d3f1c2ad06#diff-04f5f707872b10c0ab2a591bceddbf5eL79). I may well have missed a better approach to unmarshaling into an interface property, though.

payment_source.go
maxwells: It's definitely redundant data, but allows the bindings user to get the common information (ID, Type) without having to inspect the child property of that type. (in `func (s *PaymentSource) UnmarshalJSON(data []byte) error` it unmarshals to the PaymentSource itself and then the appropriate child property)

paymentsource/types.go
maxwells: yes, they should be typed!

My thinking for putting these into the paymentsource package was that the terms I wanted were already defined in the top level stripe package. I could certainly do `BitcoinReceiverType` or `CardSourceType` or something to that effect and define them in the top level `payment_source.go` file -- is that a more reasonable approach?

Thanks much for looking this over, btw.

payment_source.go
maxwells: you're correct - thanks for the tip

payment_source.go
zenazn: err, as in the following JSON blob will deserialize in an unexpected way:

```
{
    "type": "card",
    "card": {
        "number": "12345"
    },
    "expiry": "1/23"
}
```

(a similar thing will be true for the case of JSON marshaling)

maxwells:
@cosn Yeah all tests pass provided I use test keys that have a number of charges that fit into a `uint16`.

```go
type ListMeta struct {
	Count uint16 `json:"total_count"` // <-- Unmarshaling fails otherwise
	More  bool   `json:"has_more"`
	URL   string `json:"url"`
}
```

payment_source.go
maxwells: Gotcha. Yes. Forgot how github line comments work and thought you were referring to L28-29. I'll get that patched up

cosn:
Cool, let me know when all the udpates are done and I'll go through it again

bitcoin_receiver.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+)
+
+// BitcoinReceiverListParams is the set of parameters that can be used when listing BitcoinReceivers.
+// For more details see https://stripe.com/docs/api/#list_bitcoin_receivers.
+type BitcoinReceiverListParams struct {
+	ListParams
+	NotFilled  bool
+	NotActive  bool
+	Uncaptured bool
```
cosn: `NotFilled, NotActive, Uncaptured bool`

bitcoin_receiver.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// BitcoinReceiverParams is the set of parameters that can be used when creating a BitcoinReceiver.
+// For more details see https://stripe.com/docs/api/#create_bitcoin_receiver.
+type BitcoinReceiverParams struct {
+	Params
+	Amount   uint64
+	Currency Currency
+	Desc     string
+	Email    string
```
cosn: `Desc, Email string`

bitcoin_transaction.go
```diff
@@ -0,0 +1,47 @@
+package stripe
+
+import "encoding/json"
+
+// BitcoinTransactionListParams is the set of parameters that can be used when listing BitcoinTransactions.
+type BitcoinTransactionListParams struct {
+	ListParams
+	Receiver string
+	Customer string
```
cosn: `Receiver, Customer string`

bitcointransaction/client.go
```diff
@@ -0,0 +1,71 @@
+// Package bitcointransaction provides the /bitcoin/transactions APIs
+package bitcointransaction
+
+import (
+	"net/url"
+
```
cosn: why newline here?

card.go
```diff
@@ -142,6 +143,11 @@ func (c *CardParams) AppendDetails(values *url.Values, creating bool) {
 	}
 }
 
+// Human readable/displayable way of inspecting a Card
+func (c *Card) Display() string {
+	return fmt.Sprintf("%s (Last Four: %s)", c.Brand, c.LastFour)
```
cosn: `%s ending in %`

card.go
```diff
@@ -142,6 +143,11 @@ func (c *CardParams) AppendDetails(values *url.Values, creating bool) {
 	}
 }
 
+// Human readable/displayable way of inspecting a Card
```
cosn: This should start with "Display". Also period at the end

payment_source.go
```diff
@@ -0,0 +1,65 @@
...
+// can be described by SourceParams.
+// For more details see https://stripe.com/docs/api#create_charge
+type SourceParams struct {
+	Params
+	ID    string
+	Token string
+	Card  *CardParams
+}
+
+// Interface to provide a human readable representation of a struct
```
cosn: Should begin with "Displayer"

paymentsource/types.go
```diff
@@ -0,0 +1,6 @@
+package paymentsource
```
cosn: rather than putting this in its own package, can we put it in `payment_source.go` and name it `PaymentSource{BitcoinReceiver,Card}`

payment_source.go
```diff
@@ -0,0 +1,65 @@
...
+type Displayer interface {
+	Display() string
+}
+
+// PaymentSource describes the payment source used to make a Charge.
+// The Type should indicate which object is fleshed out (eg. BitcoinReceiver or Card)
+// For more details see https://stripe.com/docs/api#retrieve_charge
+type PaymentSource struct {
+	Type            string           `json:"object"`
```
cosn: Rather than `string`, can we make this a `PaymentSourceType` that just has the two consts below?

bitcointransaction/client.go
```diff
@@ -0,0 +1,71 @@
+// Package bitcointransaction provides the /bitcoin/transactions APIs
```
cosn: period at the end

bitcoinreceiver/client.go
```diff
@@ -0,0 +1,120 @@
+// Package bitcoinreceiver provides the /bitcoin/receivers APIs
```
cosn: period at the end

bitcoin_receiver.go
```diff
@@ -0,0 +1,79 @@
...
+	RefundAddress         string                  `json:"refund_address"`
+	BitcoinUri            string                  `json:"bitcoin_uri"`
+	Meta                  map[string]string       `json:"metadata"`
+	Email                 string                  `json:"email"`
+	Payment               string                  `json:"payment"`
+	Customer              string                  `json:"customer"`
+	Transactions          *BitcoinTransactionList `json:"transactions"`
+}
+
+// Human readable/displayable way of inspecting a BitcoinReceiver
```
cosn: This should begin with "Display". Also period at the end

cosn:
Generally looks good, just some minor nits. Once we get those cleared I'm happy to merge this (or feel free to merge yourself)

PR #59, opened 2015-01-12
Adding functionality for listing file uploads.

wangjohn:
This should allow you to list file uploads via the go bindings.

r? @cosn

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,7 @@
+4.3.0 2015-01-13
+  Added support for listing FileUploads
+  Mime parameter on FileUpload has been changed to Type
+
 4.2.1 2014-12-28
 	Handle charges with customer card tokens
 
--- a/fileupload/client_test.go
+++ b/fileupload/client_test.go
@@ -10,7 +10,7 @@ import (
 
 const (
 	expectedSize = 734
-	expectedMime = "application/pdf"
+	expectedType = "pdf"
 )
 
 func init() {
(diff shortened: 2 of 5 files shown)


fileupload/client_test.go
```diff
@@ -55,3 +55,40 @@ func TestFileUploadNewThenGet(t *testing.T) {
 		t.Errorf("(GET) File upload id %q does not match expected id %q\n", target.ID, res.ID)
 	}
 }
+
+func TestFileUploadList(t *testing.T) {
+	f, err := os.Open("test_data.pdf")
```
cosn: is this going to be available locally? Is it checked in already?

cosn:
:+1: 

we should also bump the version in `stripe.go`, and add a changelog entry. I can take care of merging and tagging

fileupload/client_test.go
wangjohn: Yep, it was added for tests for file uploads when they were originally added to the Go bindings (here: https://github.com/stripe/stripe-go/blob/master/fileupload/test_data.pdf)

wangjohn:
Cool, thanks! I've updated the version and changelog.

CHANGELOG
```diff
@@ -1,3 +1,6 @@
+4.2.2 2015-01-13
```
cosn: 4.2.0

stripe.go
```diff
@@ -20,10 +20,10 @@ const (
 )
 
 // apiversion is the currently supported API version
-const apiversion = "2014-12-17"
+const apiversion = "2015-01-11"
```
cosn: any other changes between these two versions?

stripe.go
```diff
@@ -20,10 +20,10 @@ const (
...
 
 // clientversion is the binding version
-const clientversion = "4.2.1"
+const clientversion = "4.2.2"
```
cosn: 4.3.0

stripe.go
wangjohn: Looks like these are the only changes made to the go bindings since the last release.

PR #63, opened 2015-01-19
Add Token support to BankAccountParams

### LINKED ISSUE #62: BankAccountParams Token Missing
According to [Create Recipient Documentation](https://stripe.com/docs/api/go#create_recipient), I should be able to pass a Stripe.js bank account token like so:

```go
Bank:  &stripe.BankAccountParams{Token: token}
```

in the same manner as:

```go
Card:  &stripe.CardParams{Token: token},
```

However, `Token` is not a defined field in `stripe.BankAccountParams`. Has it simply not been added yet?

tylerb:
This commit adds `Token` support to BankAccountParams.

Fix #62

### DIFF
--- a/recipient.go
+++ b/recipient.go
@@ -34,7 +34,7 @@ type RecipientListParams struct {
 
 // BankAccountParams is the set of parameters that can be used when creating or updating a bank account.
 type BankAccountParams struct {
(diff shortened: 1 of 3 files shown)


cosn:
Thanks!

PR #66, opened 2015-01-22
Support adding arbitrary key-value params to Stripe requests.

metcalf:
This adds the ability to include arbitrary key-value parameters in requests using the new `AddExtra` method on the `Params` type. This allows for cases where we're beta testing new parameters in the API but haven't documented them for public release.

A couple explicit decisions that are worth calling out:
* I have no idea if `Extra` is the right name here -- totally open to suggestions.
* We had discussed using a `map[string]interface{}` type for this at one point but I actually don't think this works well. There is no sense of type in URL parameters so short of writing our own serialization (ala https://github.com/google/go-querystring) we can't accept arbitrary types.
* There are no functional tests of this capability since that would either (a) depend on a transient beta feature in the API or (b) confusingly duplicate a parameter like `amount` that's already available as a static type.

r? @cosn

### DIFF
--- a/params.go
+++ b/params.go
@@ -18,6 +18,7 @@ const (
 type Params struct {
 	Exp            []string
 	Meta           map[string]string
+	Extra          url.Values
 	IdempotencyKey string
 }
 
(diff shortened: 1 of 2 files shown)


metcalf:
I realized that I don't actually need this immediately for the project I'm working on since I already do unholy things with the tokenization request. I do still think there's value in an approach like this and we have discussed other applications for it.

params.go
```diff
@@ -18,6 +18,7 @@ const (
 type Params struct {
 	Exp            []string
 	Meta           map[string]string
+	Extra          url.Values
```
cosn: `map[string]string`? Or `map[string]interface{}`?

metcalf: See my comment in the PR description for why we don't want a `map[string]interface{}`. A `map[string]string` doesn't support an array of multiple values for the same key which is a valid form. `url.Values` seems like it matches the semantics that we're using this for.

cosn:
Are we getting this merged at some point?

metcalf:
Happy to get it merged. Do you have further thoughts on the question around the type for `Extra`?

cosn:
:+1:

PR #75, opened 2015-02-18
added structure to support for unified customer sources api

maxwells:
cc @canderson, @raycmorgan

### DIFF
--- a/bitcoinreceiver/client.go
+++ b/bitcoinreceiver/client.go
@@ -2,9 +2,9 @@
 package bitcoinreceiver
 
 import (
+	"fmt"
 	"net/url"
 	"strconv"
-	"fmt"
 
 	stripe "github.com/stripe/stripe-go"
 )
--- a/bitcoinreceiver/client.go
+++ b/bitcoinreceiver/client.go
@@ -4,6 +4,7 @@ package bitcoinreceiver
 import (
 	"net/url"
 	"strconv"
+	"fmt"
 
 	stripe "github.com/stripe/stripe-go"
 )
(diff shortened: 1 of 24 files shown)


maxwells:
r? @cosn 

Updated to reflect changes in API version that shipped yesterday (card, cards, default_card => source, sources, default_source). Haven't moved everything over on the request parameter side yet -- thinking to clean that up in a second version, though it leaves the current state to be a little confusing (can pass Source{Card} or Card to mean the same thing in ChargeParams, CustomerParams, but can only access response as Source.Card)

payment_source.go
```diff
@@ -10,9 +11,10 @@ import (
 // For more details see https://stripe.com/docs/api#create_charge
 type SourceParams struct {
```
maxwells: so I have concerns about how this approach: If we continue to add sources that the customers or charges API accept as hashes (eg. Card here), then this could get pretty bloated. My reasoning for allowing a Card struct here was to not lose the clarity and explicitness of knowing exactly whats needed to create a Card source

bitcoinreceiver/client.go
```diff
@@ -63,6 +64,35 @@ func (c Client) Get(id string, params *stripe.BitcoinReceiverParams) (*stripe.Bi
 	return bitcoinReceiver, err
 }
 
+// Update updates a bitcoin receiver's properties.
+// For more details see	https://stripe.com/docs/api#update_bitcoin_receiver.
```
cosn: spacing fix?

customer/client.go
```diff
@@ -31,9 +31,11 @@ func (c Client) New(params *stripe.CustomerParams) (*stripe.Customer, error) {
 		}
 
 		if len(params.Token) > 0 {
-			body.Add("card", params.Token)
+			body.Add("source", params.Token)
 		} else if params.Card != nil {
 			params.Card.AppendDetails(body, true)
```
cosn: does this need to change as well?

paymentsource/client.go
```diff
@@ -0,0 +1,154 @@
...
+	if len(params.Customer) > 0 {
+		err = s.B.Call("POST", fmt.Sprintf("/customers/%v/sources/%v", params.Customer, id), s.Key, body, &params.Params, source)
+	} else {
+		err = errors.New("Invalid source params: customer needs to be set")
+	}
+
+	return source, err
+}
+
+// Del remotes a source.
```
cosn: removes

cosn:
We should also bump the major version and add a changelog entry.

From an input perspective, not a huge fan of having to do
```
Source: &stripe.SourceParams{
 		Card: &stripe.CardParams{
```

I wonder if it should become a method where you can just set the payment source or something *shrug*

customer/client.go
maxwells: if I understand what you're asking: it doesn't have to --  the backend is set up to take `card` or `source` for backwards compatibility so users can continue to use Card if they're keen on it

maxwells:
I like the method idea a lot. Thinking something like:

```
params := &stripe.ChargeParams{
		Amount:   1000,
		Currency: currency.USD,
}
params.SetSource(&CardParams{...})
```

which may involve a type assertion or something to that effect.

stripe.go
```diff
@@ -20,7 +20,7 @@ const (
 )
 
 // apiversion is the currently supported API version
-const apiversion = "2015-01-11"
+const apiversion = "2015-02-18"
 
 // clientversion is the binding version
 const clientversion = "4.4.3"
```
cosn: we should bump this to 5.0.0

cosn:
we'll also need a changelog entry and bumping the version

also this change implies doc changes on stripe.com/docs/api

:+1:

PR #77, opened 2015-02-25
modified invoice client to properly use GetNext upcoming url

pgrunde:
Upcoming was missing from the end of the URL


cosn:
good catch, thanks!

PR #79, opened 2015-02-27
Add Event.EventData() function - returns interface{} that represents the correct type of the event data object.

nemothekid:
Currently, when receiving events you have to either unmarshal the Raw data for yourself or use the `Event.GetObjValue(keys...)` function. For us it was just easier to have the library handle this for us.

Assumptions:

1.) If the `EventData.UnmarshalJSON` does not return an error, `Event.EventData()` will not return a JSON failure. (If it does there is a mismatch between the client library and the server which likely means the client library has to be upgraded - but I'm assuming if this isn't the case, you have some malformed data that  `EventData.UnmarshalJSON` would have caught). This assumption provides the motivation for not returning `error` in `Event.EventData()`.

2.) Events follow the format of `resource.action`, exceptions are `ping` (not sure how to handle this), and the `bitcoin.receiver` set of events (it seems these should be `bitcoin_receiver`, like `application_fee`). The `parseEventType` should be kept up to date, but if it isn't, it will return a `map[string]interface{}` much like the `EventData.Obj` (A user should assume that if the go type exists, it won't return `map[string]interface{}` I guess).

Usage:

```go
for i.Next() {
	e := i.Event()
	// You can only care about Charge events, ok will be false for other events.
	if charge, ok := e.EventData().(*stripe.Charge); ok {
		// do things with Charge Object.
	} else if card, ok := e.EventData().(*stripe.Card); ok {
		// do things with Card
	}
}
```

The changes should be backwards compatible - and tests are included under event that extend the current ones.

### DIFF
--- a/event.go
+++ b/event.go
@@ -19,9 +19,10 @@ type Event struct {
 
 // EventData is the unmarshalled object as a map.
 type EventData struct {
-	Raw  json.RawMessage        `json:"object"`
-	Prev map[string]interface{} `json:"previous_attributes"`
-	Obj  map[string]interface{}
+	Raw     json.RawMessage        `json:"object"`
+	PrevRaw json.RawMessage        `json:"previous_attributes"`
+	Prev    map[string]interface{} `json:"-"`
+	Obj     map[string]interface{} `json:"-"`
 }
 
 // EventListParams is the set of parameters that can be used when listing events.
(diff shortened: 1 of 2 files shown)


cosn:
I agree that this can work, but the downside is the long case statement which has to be maintained each time we add new things. If everyone cares about every event, then I agree it makes sense to put in the library, but I'm not sure that assumption holds. Will I'll do some digging to see if that's the case.

Thanks!

PR #85, opened 2015-03-11
New Accounts style

cosn:
* [x] stripe-account header
* [x] /v1/accounts
* [x] /v1/accounts/{id}/bank_accounts
* [x] /v/1/transfers/{id}/reversals
* [x] comments

r? @bkrausz 

Will ping when everything's done, but figured this way you can review at your leisure

### DIFF
--- a/account.go
+++ b/account.go
@@ -8,7 +8,7 @@ import (
 )
 
 // LegalEntityType describes the types for a legal entity.
-// Current values are "individual", "corporation".
+// Current values are "individual", "company".
 type LegalEntityType string
 
 // IdentityVerificationStatus describes the different statuses for identity verification.
@@ -21,7 +21,7 @@ type Interval string
 
 const (
 	Individual LegalEntityType = "individual"
-	Corp       LegalEntityType = "corporation"
+	Company    LegalEntityType = "company"
 
 	IdentityVerificationPending    IdentityVerificationStatus = "pending"
 	IdentityVerificationVerified   IdentityVerificationStatus = "verified"
--- a/account.go
+++ b/account.go
@@ -7,10 +7,16 @@ import (
 	"strconv"
 )
 
+// LegalEntityType describes the types for a legal entity.
+// Current values are "individual", "corporation".
 type LegalEntityType string
 
+// IdentityVerificationStatus describes the different statuses for identity verification.
+// Current values are "pending", "verified", "unverified".
 type IdentityVerificationStatus string
 
+// Interval describes the payout interval.
+// Current values are "manual", "daily", "weekly", "monthly".
 type Interval string
 
 const (
@@ -27,6 +33,7 @@ const (
 	Month  Interval = "monthly"
 )
 
+// AccountParams are the parameters allowed during account creation/updates.
 type AccountParams struct {
 	Params
 	Country, Email, DefaultCurrency, Statement, BusinessName, SupportPhone string
@@ -34,6 +41,7 @@ type AccountParams struct {
 	Managed                                                                bool
 }
 
+// AccountListParams are the parameters allowed during account listing.
 type AccountListParams struct {
 	ListParams
 }
@@ -71,6 +79,7 @@ type Account struct {
 	TransferSchedule *TransferSchedule `json:"transfer_schedule"`
 }
 
+// LegalEntity is the structure for properties related to an account's legal state.
 type LegalEntity struct {
 	Type             LegalEntityType      `json:"type"`
 	BusinessName     string               `json:"business_name"`
@@ -87,6 +96,7 @@ type LegalEntity struct {
 	BusinessVatID    string               `json:"business_vat_id"`
 }
 
+// Address is the structure for an account address.
 type Address struct {
 	Line1   string `json:"line1"`
 	Line2   string `json:"line2"`
@@ -96,12 +106,14 @@ type Address struct {
 	Country string `json:"country"`
 }
 
+// DOB is a structure for an account owner's date of birth.
 type DOB struct {
 	Day   int `json:"day"`
 	Month int `json:"month"`
 	Year  int `json:"year"`
 }
 
+// Owner is the structure for an account owner.
 type Owner struct {
 	First        string               `json:"first_name"`
 	Last         string               `json:"last_name"`
@@ -110,6 +122,7 @@ type Owner struct {
 	Verification IdentityVerification `json:"verification"`
 }
 
+// IdentityVerification is the structure for an account's verification.
 type IdentityVerification struct {
 	Status   IdentityVerificationStatus `json:"status"`
 	Document *struct {
@@ -120,16 +133,13 @@ type IdentityVerification struct {
 	Details *string `json:"details"`
 }
 
+// TransferSchedule is the structure for an account's transfer schedule.
 type TransferSchedule struct {
 	Delay    int      `json:"delay_days"`
 	Interval Interval `json:"interval"`
 }
 
-type AccountList struct {
-	ListMeta
-	Values []*Account `json:"data"`
-}
-
+// AppendDetails adds the legal entity to the query string.
 func (a *AccountParams) AppendDetails(values *url.Values) {
 	if a.LegalEntity != nil {
 		values.Add("legal_entity[type]", string(a.LegalEntity.Type))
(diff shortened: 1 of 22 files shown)


account.go
```diff
@@ -1,22 +1,260 @@
...
+)
+
+type LegalEntityType string
+
+type IdentityVerificationStatus string
+
+type Interval string
+
+const (
+	Individual  LegalEntityType = "individual"
```
bkrausz: These are actually only 'company' and 'individual' (we accept more, but that's all we're exposing)

account.go
```diff
@@ -1,22 +1,260 @@
...
+	SoleProp    LegalEntityType = "sole_prop"
+	LLC         LegalEntityType = "llc"
+	Partnership LegalEntityType = "partnership"
+	Corp        LegalEntityType = "corporation"
+
+	IdentityVerificationWaiting  IdentityVerificationStatus = "waiting"
+	IdentityVerificationPending  IdentityVerificationStatus = "pending"
+	IdentityVerificationVerified IdentityVerificationStatus = "verified"
+	IdentityVerificationFailed   IdentityVerificationStatus = "failed"
```
bkrausz: `waiting` and `failed` have been merged into just `unverified`

account.go
```diff
@@ -1,22 +1,260 @@
...
+		Secret  string `json:"secret"`
+		Publish string `json:"publishable"`
+	} `json:"keys"`
+	Verification *struct {
+		Fields    []string `json:"fields_needed"`
+		Due       *int64   `json:"due_by"`
+		Contacted bool     `json:"contacted"`
+	} `json:"verification"`
+	LegalEntity      *LegalEntity      `json:"legal_entity"`
+	TransferSchedule *TransferSchedule `json:"transfer_schedule"`
```
bkrausz: There's also a top-level `debit_negative_balances` bool

account.go
cosn: OK, I grabbed the list from https://stripe.com/docs/marketplaces/accounts#legal-entity

account.go
cosn: ack

account.go
cosn: ack (not documented yet)

account.go
bkrausz: Yea, it's probably better to go with the AccountAPIResource directly.

account.go
bkrausz: https://qa.stripe.com/docs/connect/accounts has some of it. We're pushing most doc changes into the new format for now to avoid needing to double-write.

account.go
cosn: left `individual` and `corporation`

account.go
cosn: added

cosn:
OK @bkrausz I think I'm done. Wanna take a look? Given that it's hard to test these things at the moment, a close look at resources, structs, JSON keys, URLs would be most valuble

account.go
```diff
@@ -1,22 +1,267 @@
...
+// Current values are "pending", "verified", "unverified".
+type IdentityVerificationStatus string
+
+// Interval describes the payout interval.
+// Current values are "manual", "daily", "weekly", "monthly".
+type Interval string
+
+const (
+	Individual LegalEntityType = "individual"
+	Corp       LegalEntityType = "corporation"
```
bkrausz: Company, not corporation (though we technically accept both)

account.go
```diff
@@ -1,22 +1,267 @@
...
+	Last         string               `json:"last_name"`
+	DOB          DOB                  `json:"dob"`
+	Address      Address              `json:"address"`
+	Verification IdentityVerification `json:"verification"`
+}
+
+// IdentityVerification is the structure for an account's verification.
+type IdentityVerification struct {
+	Status   IdentityVerificationStatus `json:"status"`
+	Document *struct {
```
bkrausz: I don't think we inline expand this...

`present :document, :string, depth: ...`

account.go
```diff
@@ -1,22 +1,267 @@
...
+		Created int64  `json:"created"`
+		Size    int64  `json:"size"`
+	} `json:"document"`
+	Details *string `json:"details"`
+}
+
+// TransferSchedule is the structure for an account's transfer schedule.
+type TransferSchedule struct {
+	Delay    int      `json:"delay_days"`
+	Interval Interval `json:"interval"`
```
bkrausz: Also may include weekly_anchor string and monthly_anchor int (only there if interval is Weekly or Monthly, respectively).

account.go
```diff
@@ -1,22 +1,267 @@
...
+		ID      string `json:"id"`
+		Created int64  `json:"created"`
+		Size    int64  `json:"size"`
+	} `json:"document"`
+	Details *string `json:"details"`
+}
+
+// TransferSchedule is the structure for an account's transfer schedule.
+type TransferSchedule struct {
+	Delay    int      `json:"delay_days"`
```
bkrausz: This can be passed up as, optionally, "minimum" to select the minimum value. Any way to support this in Go?

account/client.go
```diff
@@ -11,8 +14,50 @@ type Client struct {
...
+
+	if len(params.BusinessName) > 0 {
+		body.Add("business_name", params.BusinessName)
+	}
+
+	if len(params.SupportPhone) > 0 {
+		body.Add("support_phone", params.SupportPhone)
+	}
+
+	params.AppendDetails(body)
```
bkrausz: Do we need to do the same for transfer_schedule?

account.go
```diff
@@ -1,22 +1,267 @@
...
+	Interval Interval `json:"interval"`
+}
+
+// AppendDetails adds the legal entity to the query string.
+func (a *AccountParams) AppendDetails(values *url.Values) {
+	if a.LegalEntity != nil {
```
bkrausz: We also support verification[document] passing for both the legal_entity and the additional_owners.

account.go
cosn: updated

account.go
cosn: iow you have to manually expand or that it's not expandable?

account.go
cosn: sorry, not following

reversal/client.go
```diff
@@ -0,0 +1,113 @@
...
+	B   stripe.Backend
+	Key string
+}
+
+// New POSTs a new transfer reversal.
+func New(params *stripe.ReversalParams) (*stripe.Reversal, error) {
+	return getC().New(params)
+}
+
+func (c Client) New(params *stripe.ReversalParams) (*stripe.Reversal, error) {
```
bkrausz: Reversals have a "refund_application_fee" bool

bankaccount.go
```diff
@@ -0,0 +1,42 @@
...
+	ListParams
+	AccountID string
+}
+
+// BankAccount represents a Stripe bank account.
+type BankAccount struct {
+	ID          string            `json:"id"`
+	Name        string            `json:"bank_name"`
+	Country     string            `json:"country"`
+	Currency    Currency          `json:"currency"`
```
bkrausz: This now also has routing_number exposed

bkrausz:
Do we support/want to support:

1) bank_accounts list when retrieving the Account?
2) bank_account param when updating the Account?

account.go
bkrausz: I'm 80% sure manually expand

account.go
bkrausz: Instead of setting a number on this param, you could set the string "minimum".

cosn: ah, gotcha ... yeah we'd have to have a different parameter. Will add

account.go
cosn: I'll add an `Anchor` param, and based on the presence of Weekly or Monthly will populate the right value over the wire

account.go
cosn: changed

account.go
cosn: OK, since the types were different (string vs. int), I added them as two separate properties.

Just to be clear, the TransferSchedule is just gettable on an account, or can it also be set via the API?

bkrausz: Settable as well

account/client.go
cosn: added

reversal/client.go
cosn: added

bankaccount.go
cosn: added

PR #89, opened 2015-03-27
Add custom MarshalJSON method to PaymentSource, so it accurately represents the JSON rep. of its type

### LINKED ISSUE #88: PaymentSource truncates all values on the Source when Marshalled
Hi,

Currently we currently proxy requests to Stripe (for example, if the web application need to access the logged in customers' `Customer` object it hits our API, which hits Stripe).

Unfortunately, when the `Customer` object is marshalled, it loses all `Source` information, which is somewhat unexpected. 

Perhaps `PaymentSource` can be transformed into an interface with `Type()`,`ID()`,`Card()` and `BitcoinReciever()` methods, and have `Card` or `BitcoinReciever` implement these methods? (Not backwards compat?)

Or maybe add a custom `MarshalJson` function that makes sure `PaymentSource` marshals itself appropriately?

nemothekid:
This PR adds a custom `MarshalJSON` method to `PaymentSource`. Now when the `PaymentSource` object is marshaled, it will represent the same JSON object that it was presented with.

Code:
```
//...
var customer Customer
//...
return json.Marshal(customer)
```

Before Output:
```json
"customer": {
  "id": "djndfjkngknjgnkjfnkg",
  "sources": {
    "data": [
      {
        "object": "card",
        "id": "card_5wey6SaMFjuRx6"
      }
    ]
  }
}
```

After Output:
```json
"customer": {
  "id": "djndfjkngknjgnkjfnkg",
  "sources": {
    "data": [
      {
        "object": "card",
        "customer": "djndfjkngknjgnkjfnkg",
        "id": "dfjvkjfnknf",
        "exp_month": 10,
        "exp_year": 2016,
        "more_keys": "more_values"
      }
    ]
  }
}
```

This fixes #88 

**Edit:** I did some work handling the third case of the switch statement, where the type isn't a `PaymentSourceBitcoinReceiver` or `PaymentSourceCard`. I thought it was alright if we just fell back to the default, but that didn't really produce a clean output either. However, looking at the way `UnmarshalJSON` works, it seems pretty confident it's type is either `PaymentSourceBitcoinReceiver`, `PaymentSourceCard` or an id string. Based on this, I felt that just marshaling the raw string is the best option. Internally, you guys might have other ideas.

### DIFF
--- a/payment_source.go
+++ b/payment_source.go
@@ -137,3 +137,33 @@ func (s *PaymentSource) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// MarshalJSON handles serialization of a PaymentSource.
+// This custom marshaling is needed because the specific type
+// of payment instrument it represents is specified by the PaymentSourceType
+func (s *PaymentSource) MarshalJSON() ([]byte, error) {
+	var target interface{}
+
+	switch s.Type {
+	case PaymentSourceBitcoinReceiver:
+		target = struct {
+			Type PaymentSourceType `json:"object"`
+			*BitcoinReceiver
+		}{
+			Type:            s.Type,
+			BitcoinReceiver: s.BitcoinReceiver,
+		}
+	case PaymentSourceCard:
+		target = struct {
+			Type     PaymentSourceType `json:"object"`
+			Customer string            `json:"customer"`
+			*Card
+		}{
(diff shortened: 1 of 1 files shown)


cosn:
Thanks!

PR #90, opened 2015-03-30
Add balance transaction to reversals

ryanlower:
Adds Tx field to Reversal struct

I wasn't sure if/how you'd want tests for this - it doesn't look like you have equivalent tests on charges or transfers - happy to add if you want though


cosn:
Thanks!

PR #94, opened 2015-04-24
Added terms of service acceptance for managed accounts

stephenmuss:


account.go
```diff
@@ -154,6 +160,12 @@ type TransferSchedule struct {
 	MonthAnchor uint64   `json:"monthly_anchor"`
 }
 
+type TOSAcceptanceParams struct {
```
cosn: can you please add a comment for this? All public structs and methods should have comments

cosn:
I'll take care of it :) Thanks for submitting!

Issue #95, opened 2015-04-24
Customer struct does not support 'deleted' property


peterkellyonline:
According to Stripe API, when [retrieving a customer](https://stripe.com/docs/api#retrieve_customer) that has been deleted, Stripe returns a subset of that customer's data and an extra property is returned called 'deleted' with a value of 'true'. 

Without that being supported in Customer struct it is difficult to handle deleted customers.

PR #96, opened 2015-04-24
Adding Deleted field to Customer

peterkellyonline:
When you retrieve a customer that has been deleted, the Stripe API returns a property 'deleted' along with a subset of the Customer details. That is missing from Customer struct so I am adding it here along with a test. 

I also fixed an existing test (TestCustomerUpdate) that was failing because no CVC was supplied with CardParams. 

Some formatting fixes too. 

This fixes the issue I raised - https://github.com/stripe/stripe-go/issues/95

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -47,6 +47,7 @@ type Customer struct {
 	Email         string            `json:"email"`
 	Meta          map[string]string `json:"metadata"`
 	Subs          *SubList          `json:"subscriptions"`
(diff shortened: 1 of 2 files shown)


peterkellyonline:
Hey @cosn, is this pull request okay? Not sure how long it takes to get reviewed and merged in. Let me know if I am missing something that I need to fix/add. Thanks

cosn:
Thanks!

PR #98, opened 2015-05-15
Add DynamicLastFour to Card struct.

SimonRichardson:
According to the following documentation[1] it should be possible to
retrieve the dynamic last 4 digits. The documentation states that it
could potentially be `null`. The pull request makes the unmarshalling
of the json optional as well, with the usage of `json:"xxx,omitempty"`.

[1]https://stripe.com/docs/api#retrieve_token

### DIFF
--- a/card.go
+++ b/card.go
@@ -40,26 +40,27 @@ type CardListParams struct {
 // Card is the resource representing a Stripe credit/debit card.
 // For more details see https://stripe.com/docs/api#cards.
 type Card struct {
-	ID            string       `json:"id"`
-	Month         uint8        `json:"exp_month"`
-	Year          uint16       `json:"exp_year"`
-	Fingerprint   string       `json:"fingerprint"`
-	Funding       CardFunding  `json:"funding"`
-	LastFour      string       `json:"last4"`
-	Brand         CardBrand    `json:"brand"`
-	City          string       `json:"address_city"`
-	Country       string       `json:"address_country"`
-	Address1      string       `json:"address_line1"`
-	Address1Check Verification `json:"address_line1_check"`
-	Address2      string       `json:"address_line2"`
-	State         string       `json:"address_state"`
-	Zip           string       `json:"address_zip"`
-	ZipCheck      Verification `json:"address_zip_check"`
-	CardCountry   string       `json:"country"`
-	Customer      *Customer    `json:"customer"`
-	CVCCheck      Verification `json:"cvc_check"`
-	Name          string       `json:"name"`
-	Recipient     *Recipient   `json:"recipient"`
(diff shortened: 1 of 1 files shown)


abh:
Unmarshalling "missing" fields works already. "Omitempty" is only used when converting (marshaling) _to_ JSON.

SimonRichardson:
Do you have a example of unmarshalling missing fields?

abh:
I'm not sure what you mean. The most basic of example will show that "extra" fields in the JSON data are just ignored: http://play.golang.org/p/BECR8TRw1f

SimonRichardson:
I think we've got our wires crossed, I want to use the `stripe.Card` struct to gain access to `dynamic_last4` property found in the stripe api reference.

Below is an extract from the link I pasted. If you're saying that I have to read in the post myself, then I don't see the point of utilising the library, which has all the typed contracts to the stripe api json. 

There is currently no way to inject your own struct into replace the `stripe.Token`[1] during the unmarshalling phase. I see two possibilities:

 # I roll my own POST and use my own unmarshalling
 # I override the `func (c Client) Get` so that I can inject a new version of the token struct so that I can pass in my own card.

```json
&stripe.Token JSON: {
  "id": "tok_163faf2eZvKYlo2CXEPOkuAK",
  "livemode": false,
  "created": 1431922257,
  "used": false,
  "object": "token",
  "type": "card",
  "card": {
    "id": "card_163faf2eZvKYlo2CPOj4TYdE",
    "object": "card",
    "last4": "4242",
    "brand": "Visa",
    "funding": "credit",
    "exp_month": 8,
    "exp_year": 2016,
    "country": "US",
    "name": null,
    "address_line1": null,
    "address_line2": null,
    "address_city": null,
    "address_state": null,
    "address_zip": null,
    "address_country": null,
    "cvc_check": null,
    "address_line1_check": null,
    "address_zip_check": null,
    "dynamic_last4": null,
    "metadata": {
    }
  }
}
```

Also note: I added `omitempty` in case `Card` was used as consumable and didn't want that to cause failures. If that's not the case I can remove it.

[1] https://github.com/stripe/stripe-go/blob/master/token/client.go#L73

iceydee:
@cosn what's your take on this?

cosn:
it's fine to add the `DynamicLastFour` but as @abh mentions, the `omitempty` is not used on demarshaling, so we can leave it off.

Also, can you please update tests for this new property?

Thanks!

cosn:
took care of it. Thanks for submitting!

SimonRichardson:
@cosn fyi it's `dynamic_last4` not `dynamic_last_four`

cosn:
good catch, fixed!

Issue #101, opened 2015-05-21
Is DefaultSource missing on the CustomerParams?


kytrinyx:
I'm trying to do the following:

* create a new card (using the card creation API endpoint)
* make the new card the default (by using the customer update API epndoint).

According to the documentation (https://stripe.com/docs/api#update_customer) there is an optional `default_source` parameter, however don't see it on the customer params: https://github.com/stripe/stripe-go/blob/master/customer.go#L7-L17

Is this named something else? Is there another preferred way of doing this?

UPDATE: It looks like this used to be `default_card`, but that this was deprecated and then removed in the API change of [2015-02-18](https://stripe.com/docs/upgrades#2015-02-18). I'll submit a PR to change DefaultCard to DefaultSource, since it seems like it is not there intentionally.

PR #102, opened 2015-05-21
Rename deprecated customer param value

kytrinyx:
The API was updated on [2015-02-18](https://stripe.com/docs/upgrades#2015-02-18) to remove the old Card and DefaultCard
parameters, in favour of Source and DefaultSource.

The codebase was updated in https://github.com/stripe/stripe-go/commit/43e56d28015faea5136409a86cad1664fb321c23 to reflect the new
changes to the API, but the customer params did not get updated at that
time.

This relates to #101.

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -13,7 +13,7 @@ type CustomerParams struct {
 	Plan          string
 	Quantity      uint64
 	TrialEnd      int64
-	DefaultCard   string
+	DefaultSource string
 }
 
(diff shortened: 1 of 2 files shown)


kytrinyx:
OK, I've clearly missed something. It looks like the client() is expecting something to do with DefaultCard.

cosn:
Thanks!

PR #104, opened 2015-05-23
Exposed logging object

### LINKED ISSUE #97: Expose logger
I have a use case where I need to track transactions going across my application. Each request has a unique ID which I attach to the log entry. I propose that instead of using the `log` package as it is seen [here](https://github.com/stripe/stripe-go/blob/3e087c3123b07601e1409d0cd660f62db0819195/stripe.go#L258), that a `Logger` variable be added as a package level variable to `stripe.go` where the whole package could log to instead. This would allow runtime customization of the logging prefix, and level to a much higher degree then the `stripe.LogLevel` variable gives.

Could create a new logger in an `init` function as:

```go
var Logger *log.Logger

func init() {
    ...
    Logger = log.New(os.Stderr, "", log.LstdFlags)
    ...
}
```

Then customizations could be done with `stripe.Logger.SetPrefix` or `stripe.Logger.SetFlags`.

wyattjoh:
Should resolve #97 to expose logger. Retained `LogLevel` for backwards compatibility.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"os"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


cosn:
thanks! will take a look next week

shazow:
:+1:

cosn:
thanks!

PR #110, opened 2015-06-05
Add missing fields to invoice line

kytrinyx:
The documentation for the invoice object
https://stripe.com/docs/api#invoice_object
shows that a line may have a subscription (string) and a discountable (bool).

I named the subscription Sub to match the equivalent field on other types in
this package.


cosn:
thanks!

PR #111, opened 2015-06-23
account: Adds new support/business fields

mduvall:
We recently documented a few support/business fields at `/v1/account` (see [here](https://stripe.com/docs/api#update_account)) and this adds a few new fields: `business_primary_color`, `business_url`, `support_email`, and `support_url`. Let me know if any changes need to be made.

r? @cosn


cosn:
:+1:

PR #113, opened 2015-07-01
Add destination, application fee, and transfer to Charge struct

ryanlower:
Charges created with a destination (connect) have these properties.

Named `Dest`, `Fee`, & `Transfer` for consistency with keys used in charge creation.


cosn:
thanks!

PR #114, opened 2015-07-04
Update bank account creation API.

zkirill:
Pass **token** or **dictionary** containing a user’s bank account details as
value for `external_account`.

Please see https://stripe.com/docs/api/go#create_bank_account for reference.


cosn:
thanks!

PR #115, opened 2015-07-06
Add request ID to Stripe errors

kyleconroy:
While request IDs are never in the Stripe error JSON, I added the
`json:"request_id"` so that the ID will show up when the error is printed.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -244,6 +244,7 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
 				Type:           ErrorType(root["type"].(string)),
 				Msg:            root["message"].(string),
 				HTTPStatusCode: res.StatusCode,
+				RequestID:      res.Header.Get("Request-Id"),
 			}
 
 			if code, found := root["code"]; found {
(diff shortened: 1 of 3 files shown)


stripe.go
```diff
@@ -244,6 +244,7 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
 				Type:           ErrorType(root["type"].(string)),
 				Msg:            root["message"].(string),
 				HTTPStatusCode: res.StatusCode,
+				RequestID:      res.Header.Get("Request-Id"),
```
cosn: i'm assuming this is always present

cosn:
wanna add a test for this?

also, don't forget to bump CHANGELOG and versions and cut a new tag after merging in master

stripe.go
kyleconroy: Yes. Even in the case that it isn't, this will just return an empty string. Since we serialize the error to JSON when printing, and we have `omitempty` in the json tag, an empty value won't show.

kyleconroy:
What tests would you like me to add beyond the one I added to `error_test`. I'll make sure to bump and tag after I merge.

cosn:
ah nm forgot about that ... all good, thanks!

:+1:

PR #116, opened 2015-07-07
Move proration date to subscription update

kytrinyx:
Originally the proration date got appended to the body of the POST request to create a subscription,
which makes no sense at all.

What we really need, is to add the proration date on the update call.


cosn:
thanks!

PR #117, opened 2015-07-08
Send "country" when creating or updating an account.

zkirill:
Pass `country` argument, if provided. This is an optional argument to **POST** `https://api.stripe.com/v1/accounts`.

Please see https://stripe.com/docs/api#create_account.


cosn:
thanks!

PR #122, opened 2015-07-16
Added support for verify bank account endpoint

jackdent:


jackdent:
I'm unclear on whether `bankaccount/client.go` should be polymorphic on the `/accounts` and the `/customers/:id/sources` endpoints, as it is in the other client libraries. For now, the verify method is lurking on top of `paymentsource/client.go`

cosn:
cool ... can we add a test for this?

jackdent:
Following a discussion with @bkrausz , I just planned an API review session on Monday to discuss the design of the `.../bank_accounts/...` endpoint going forward. This PR may become redundant, pending that.

PR #124, opened 2015-07-21
total_count should be uint32

dclausen:
in my experience, some stripe accounts have more than 65535 charges.


kyleconroy:
Agreed! Thanks for the PR

PR #125, opened 2015-07-29
Added the ability to delete an account

shalecraig:
R? @cosn


cosn:
:+1: 

We should also have a CHANGELOG entry, bump the version in `stripe.go` and then once merged to master, tag it (happy to take any/all of that if useful)

shalecraig:
merging - can you take that?

cosn:
ack

PR #127, opened 2015-08-10
TOSAcceptance are not updated when account is updated

lisardo:
Account is correctly updated when I set AccountParams.
However, TOSAcceptance is not updated no matter what is set.

It happens because TOSAcceptance.AppendDetails is not called in the account update.


kyleconroy:
@lisardo thanks for the PR

PR #128, opened 2015-08-11
Allow TOSAccept on Account update

telecoda:
Updating terms of service, was only supported on Account create, NOT Account update.

Fixes incorrect parameter name on TransferSchedules


kyleconroy:
:see_no_evil: That typo...

Thanks for the change.

PR #131, opened 2015-08-13
Use top level refund endpoints.

cupcait:
TODO: change `beta_refunds` endpoint before merging, once non-beta endpoints have shipped.

r? @cosn 
cc @kyleconroy 

Note: I added a bunch of error raising because we noticed during debugging that a lot of things wouldn't be caught. If this is too much, let me know and I'm happy to remove.

Also - I wouldn't mind having a test to make sure that the list method works without the charge filter, but this returns a large set of results and I'm not sure verifying the size is straightforward with the returned iterator. Open to suggestions! Or I can leave as is.

### DIFF
--- a/refund.go
+++ b/refund.go
@@ -20,7 +20,6 @@ type RefundParams struct {
 // For more details see https://stripe.com/docs/api#list_refunds.
 type RefundListParams struct {
 	ListParams
-	Charge string
 }
 
 // Refund is the resource representing a Stripe refund.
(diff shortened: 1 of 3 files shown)


kyleconroy:
@cupcait you could add a `limit` filter to the list call to prevent fetching __all__ the refunds for the test account.

cosn:
:+1: 

(you'll also want to update `CHANGEDOC`, bump the version in `stripe.go` and then we'll need to tag once in master; i'm happy to take any/all of these)

cupcait:
Thanks! I've added another test for listing (with limit param). 

I'm happy to take the changelog/version tasks, but I don't remember what happens at release time vs now, in the PR - should I update the changelog now? (I think @kyleconroy knows this better than I do and should be able to help with the release when this goes out.)

kyleconroy:
No need to update the `CHANGEDOC` now. Whomever releases the new version will make that a separate commit.

PR #132, opened 2015-08-14
Detect nil `c.Card.Customer` during `PaymentSource.MarshalJSON()`

### LINKED ISSUE #118: Invalid memory address when json.Marshalling a charge response
I'm trying to save the charge response as an audit log in postgresql as a JSON field but whenever I try to `json.Marshal` the object I get an nil pointer reference.

The trace output shows the error as originating here: `github.com/stripe/stripe-go/payment_source.go:165` which is the `MarshalJSON()` method on the [`PaymentSource`](https://github.com/stripe/stripe-go/blob/master/payment_source.go#L144-L171)

It seems to be caused because `c.Card.Customer` is `nil` [here](https://github.com/stripe/stripe-go/blob/master/payment_source.go#L163).

I've created a work around for it that satisfies my use-case in which in the `MarshalJSON` method I check for the `c.Card.Customer` object and only assign the Customer if it's set:

```go
func (s *PaymentSource) MarshalJSON() ([]byte, error) {
	var target interface{}

	switch s.Type {
	case PaymentSourceBitcoinReceiver:
		target = struct {
			Type PaymentSourceType `json:"object"`
			*BitcoinReceiver
		}{
			Type:            s.Type,
			BitcoinReceiver: s.BitcoinReceiver,
		}
	case PaymentSourceCard:
		var customerID *string
		if s.Card.Customer != nil {
			customerID = &s.Card.Customer.ID
		}

		target = struct {
			Type     PaymentSourceType `json:"object"`
			Customer *string           `json:"customer"`
			*Card
		}{
			Type:     s.Type,
			Customer: customerID,
			Card:     s.Card,
		}
	case "":
		target = s.ID
	}

	return json.Marshal(target)
}
```

My code:

```go
sc := &client.API{}
sc.Init(stripeCredentials.TestKey, nil)

	Amount:    uint64(order.TotalIncTax * 100),
	Currency:  currency.GBP,
	Desc:      "Payment To " + store.Name + " for Your Order #" + strconv.FormatInt(order.ExternalID, 10),
	NoCapture: true,
}

chargeParams.Params.IdempotencyKey = stripe.NewIdempotencyKey()

chargeParams.SetSource(&stripe.CardParams{
	Number:   cardDetails.CardNumber,
	Month:    cardDetails.ExpiryMonth,
	Year:     cardDetails.ExpiryYear,
	CVC:      cardDetails.CVC,
	Address1: order.BillingAddress.Line1,
	Address2: order.BillingAddress.Line2,
	City:     order.BillingAddress.City,
	State:    order.BillingAddress.State,
	Zip:      order.BillingAddress.PostalCode,
	Country:  order.BillingAddress.CountryISO,
})

charge, err := sc.Charges.New(chargeParams)

// This method appends the passed value to the transaction AuditLog field which is then json.Marshalled()
// This call only fails if the charge object is passed in. Anything else works
transaction.AddAuditLog(&TransactionAuditLog{
    Error: err,
    Response: charge,
})
```

chakrit:
Fix #118 (code by @imjoshholloway)

### DIFF
--- a/payment_source.go
+++ b/payment_source.go
@@ -154,15 +154,21 @@ func (s *PaymentSource) MarshalJSON() ([]byte, error) {
 			BitcoinReceiver: s.BitcoinReceiver,
 		}
 	case PaymentSourceCard:
+		var customerID *string
+		if s.Card.Customer != nil {
+			customerID = &s.Card.Customer.ID
+		}
+
 		target = struct {
 			Type     PaymentSourceType `json:"object"`
-			Customer string            `json:"customer"`
+			Customer *string           `json:"customer"`
 			*Card
 		}{
 			Type:     s.Type,
-			Customer: s.Card.Customer.ID,
+			Customer: customerID,
 			Card:     s.Card,
 		}
+
 	case "":
 		target = s.ID
 	}


imjoshholloway:
Thanks @chakrit! Much appreciated :+1:

kyleconroy:
\o/ Thanks for the change :+1:

PR #134, opened 2015-08-25
Add Metadata When Creating a Charge

coreymgilmore:
The official Stripe golang library does not allow for adding metadata when creating a charge.  This ability is available in other languages.  This update just allows the golang library to do this as well.

It is a pretty simple update: it allows a user to add a map of type map[string]string fields to the "Meta" struct field in the ChargeParams struct.  This map is then broken out into individual key/value pairs that are appended to the POST request body.  The important part to note is that each key is appended to the "metadata" field like so: metadata[key].


kyleconroy:
@coreymgilmore This change looks good, but can you please back out the README deletion change? That seems unrelated.

kyleconroy:
Actually, it looks like metadata is already supported. `ChargeParams` embeds the `Params` struct which has support for metadata. The following will work:

```go
params := ChargeParams{}
params.AddMeta("key", "value)
```

coreymgilmore:
@kyleconroy Thanks for the `params.AddMeta()` hint.  That makes this wayyyyy easier.

On an aside: I must say, I think the Golang documentation is on the worse side of okay.  I end up digging through the source code more than the api reference on the Stripe website!  This should definitely be something documented in the official docs.

kyleconroy:
I'm not sure I agree with you. This is documented in the metadata section on the Stripe website. https://stripe.com/docs/api#metadata

coreymgilmore:
@kyleconroy disregard that....it just is in the "metadata" category.  Ha!  I was looking around in the charges, customer,... categories in the docs.  Thanks!

PR #138, opened 2015-08-31
Add expandable parameters to ListParams

### LINKED ISSUE #119: Customer Email and Desc params aren't loaded
I can load a customer via customer.Get(customerId) directly and see the Email and Desc correctly. 

However, if I load all charges for that same customer using charge.List(), and then try to access charge.Customer.Desc and charge.Customer.Email, they are both always blank.

I was expecting charge.Customer to display all attributes from the Customer struct.

In the ruby client, I'm able to pass expand: ['data.customer'] to load the details of the customer. However, I don't see how to do so using ChargeListParams or ListParams.

kyleconroy:
r? @cosn 

Fixes #119

### DIFF
--- a/params.go
+++ b/params.go
@@ -26,6 +26,7 @@ type Params struct {
 // ListParams is the structure that contains the common properties
 // of any *ListParams structure.
 type ListParams struct {
+	Exp        []string
 	Start, End string
 	Limit      int
 	Filters    Filters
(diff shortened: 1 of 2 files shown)


cosn:
Should we add an actual API test rather than unit test for this?

:+1: otherwise

kyleconroy:
I think the unit test is fine.

PR #140, opened 2015-09-10
reveal HTTP status code json

jxu577:
r? @kyleconroy 

cc @cos we want to show the HTTP status code because we now have different ones such as 401, 404, 429, etc. - was there a reason we didn't show this in the first place?


kyleconroy:
I'm :+1: but would like to hear what @cos has to say.

cosn:
:+1:

cos:
you might be cc-ing the wrong @cos ;-)

kyleconroy:
@cos sorry! @cosn thanks :)

PR #141, opened 2015-09-16
Add easier client initialization from HTTP client

### LINKED ISSUE #86: App Engine: "Call error 3: invalid security ticket"
The documented approach for using this package on App Engine works for the first HTTP request, but never again.  Specifically, the documentation says to call

```go
stripe.SetHTTPClient(urlfetch.Client(appengine.NewContext(r)))
```

That creates and uses an HTTP client which is only valid for the life of the current HTTP request (`r`).  On the next request, one calls `stripe.SetHTTPClient` again.  That sets the default HTTP client but doesn't change the API backend, which still references the HTTP client from the first request.

A second HTTP request might call something like `customer.New(...)` which calls `stripe.GetBackend(stripe.APIBackend)` which returns a backend referring to an HTTP client from the first request.  Using that HTTP client gives an error like:

```
Post https://api.stripe.com/v1/customers: Call error 3: invalid security ticket: 6e2752a700a25c35
```

A better incantation for App Engine is something like:

```go
stripe.SetBackend(stripe.APIBackend, nil)
stripe.SetHTTPClient(urlfetch.Client(appengine.NewContext(r)))
```

I'd submit a pull request, but I'm not sure if this should be corrected in the documentation, in `SetHTTPClient` or in `GetBackend`

Somewhat related, using a global HTTP client in App Engine is bound to cause trouble when simultaneous requests want to call Stripe APIs.  Both goroutines will contend over stripe.httpClient and at most one goroutine can win.  I wish I could recommend a good solution to that problem.

kyleconroy:
r? @cosn Fixes #86

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -135,3 +135,11 @@ func (a *API) Init(key string, backends *Backends) {
 	a.Reversals = &reversal.Client{B: backends.API, Key: key}
 	a.BankAccounts = &bankaccount.Client{B: backends.API, Key: key}
 }
+
+// New creates a new Stripe client with the appropriate secret key
+// as well as providing the ability to override the backends as needed.
+func New(key string, backends *Backends) *API {
+	api := API{}
+	api.Init(key, backends)
+	return &api
+}
--- a/stripe.go
+++ b/stripe.go
@@ -92,6 +92,17 @@ func SetHTTPClient(client *http.Client) {
 	httpClient = client
 }
 
+// NewBackends creates a new set of backends with the given HTTP client. You
+// should only need to use this for testing purposes or on App Engine.
+func NewBackends(httpClient *http.Client) *Backends {
+	return &Backends{
+		API: BackendConfiguration{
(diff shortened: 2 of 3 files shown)


client/api.go
```diff
@@ -135,3 +135,9 @@ func (a *API) Init(key string, backends *Backends) {
 	a.Reversals = &reversal.Client{B: backends.API, Key: key}
 	a.BankAccounts = &bankaccount.Client{B: backends.API, Key: key}
 }
+
+func New(key string, backends *Backends) *API {
```
cosn: can we add some docs since it's a public func?

stripe.go
```diff
@@ -92,6 +92,15 @@ func SetHTTPClient(client *http.Client) {
 	httpClient = client
 }
 
+func NewBackends(httpClient *http.Client) *Backends {
```
cosn: Ditto

cosn:
:+1:

PR #145, opened 2015-10-21
Add `external_account` parameter for account update requests

### LINKED ISSUE #144: External Account functionality missing
I'd like to use stripe connect, but I need to be able to set the `external_account` param in the `AccountParams` in order to attach a bank account to the `Account` object using the token retrieved from Stripe.js.

brandur:
Allows accounts to be updated with an `external_account` as defined in
the API.

I've tried to follow the pattern demonstrated in the "recipient"
resource here, but I'm not totally sure if it's fully correct.

Fixes #144.

/cc @kyleconroy Mind taking a look at this? Thanks!

### DIFF
--- a/account.go
+++ b/account.go
@@ -38,6 +38,7 @@ type AccountParams struct {
 	Params
 	Country, Email, DefaultCurrency, Statement, BusinessName, BusinessUrl,
 	BusinessPrimaryColor, SupportPhone, SupportEmail, SupportUrl string
+	ExternalAccount  *BankAccountParams
 	LegalEntity      *LegalEntity
 	TransferSchedule *TransferScheduleParams
 	Managed          bool
--- a/account.go
+++ b/account.go
@@ -38,17 +38,24 @@ type AccountParams struct {
 	Params
 	Country, Email, DefaultCurrency, Statement, BusinessName, BusinessUrl,
 	BusinessPrimaryColor, SupportPhone, SupportEmail, SupportUrl string
-	ExternalAccount  *BankAccountParams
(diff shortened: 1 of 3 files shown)


kyleconroy:
:+1: Looks good to me

brandur:
Ugh. Just realized that the existing `BankAccount` parameters will [use the wrong parameter name when `AppendDetails` is called](https://github.com/stripe/stripe-go/blob/brandur-add-account-external-account/bankaccount.go#L43,L52).

kyleconroy:
I guess we'll need to pass a parameter name to `AppendDetails` :(

brandur:
Hey @kyleconroy, mind taking a look at this one once more?

As mentioned at lunch, I believe that a `BankAccount` was incorrectly added to `Account` instead of the correct `ExternalAccount`. If passed in, it would result in the serialization of `bank_account` cache, which neither the create nor update actions [on the account resource](https://stripe.com/docs/api/curl#create_account) seem to have.

This pull removes that and adds a custom set of parameters called somewhat awkwardly `AccountExternalAccountParams` (logic being that it's "external account" under "account"). My other option is to re-use the `BankAccountParams` struct for this, but this doesn't seem to be particularly advantageous. Any opinion either way? Thanks!

kyleconroy:
I looked over the documentation more closer and believe that you are correct. This change will now send the correct parameters. The name of the struct is also an improvement, as external accounts can also be debit cards.

kyleconroy:
:+1:

brandur:
@kyleconroy Thanks!

PR #146, opened 2015-10-22
Add debit_negative_balances to Account object

tylerb:


brandur:
@tylerb Sweet!! Thanks for sending this in!

It seems that in general most of this library tends to avoid abbreviations by convention (or as far as I can tell anyway). What do you think about expanding `DebitNegativeBal` to `DebitNegativeBalances` even if is a little verbose.

tylerb:
I specifically used `DebitNegativeBal` so it would be consistent with https://github.com/stripe/stripe-go/blob/master/account.go#L92. I prefer the full name as well, but I figured it would be best to stick with what already exists.

brandur:
Oh man, I hadn't noticed that. +1 for keeping that naming consistent then.

Thanks again!

PR #148, opened 2015-11-10
Customer shipping support, Minor Fixes

seenickcode:
- ability to associate Customer Shipping during Customer creation and updating
- more tests to verify Customer Shipping fields
- ** important fix for updating Charge Shipping Address zip code


brandur:
Thank-you Nick! This looks good to me. Great catch on the `zip` vs. `postal_code` discrepancy.

seenickcode:
Thanks @brandur .  Happy to contribute.

PR #151, opened 2015-11-12
Return the resource when deleting it

### LINKED ISSUE #150: Subscription.Cancel does not return the subscription object (per spec)
I'm referring specifically to [`func sub.Cancel(id string, params *stripe.SubParams) error`](https://github.com/stripe/stripe-go/blob/master/sub/client.go#L164)

The [spec](https://stripe.com/docs/api/go#cancel_subscription) specifies that canceling a subscription returns a "canceled subscription object". Looks like [all language examples except Go](https://stripe.com/docs/api/python#cancel_subscription) return the canceled subscription object. Especially since we can not fetch inactive subscriptions via the API, it would be awesome of at least the now-inactive subscription was returned to the Cancel call.

remistr:
When you cancel a subscription or delete a customer it now returns the
corresponding resource.

Fixes #150.

### DIFF
--- a/customer/client.go
+++ b/customer/client.go
@@ -138,12 +138,16 @@ func (c Client) Update(id string, params *stripe.CustomerParams) (*stripe.Custom
 
 // Del removes a customer.
 // For more details see https://stripe.com/docs/api#delete_customer.
-func Del(id string) error {
+func Del(id string) (*stripe.Customer, error) {
 	return getC().Del(id)
 }
 
-func (c Client) Del(id string) error {
-	return c.B.Call("DELETE", "/customers/"+id, c.Key, nil, nil, nil)
+func (c Client) Del(id string) (*stripe.Customer, error) {
+
+	cust := &stripe.Customer{}
+	err := c.B.Call("DELETE", "/customers/"+id, c.Key, nil, nil, cust)
+
+	return cust, err
 }
 
 // List returns a list of customers.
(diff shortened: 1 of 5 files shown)


brandur:
@remistr Hey! Wow, so first of all, thanks a lot for putting this together! We really appreciate the contribution.

The changes the the subscription cancellation piece look great to me! I think we should get this merged in to immediately address the concerns in #150.

Regarding the changes to `Del` on `Customer` though: I think I missed something, but is there any particular reason that we're making those in this PR? As far as I can tell, the change doesn't seem to have been requested in #150. I only ask because these changes are a little more tricky because it seems like convention so far is to just have `Del` return an error only:

```
$ ag 'func Del'
account/client.go
176:func Del(id string) error {

bankaccount/client.go
104:func Del(id string, params *stripe.BankAccountParams) error {

card/client.go
119:func Del(id string, params *stripe.CardParams) error {

coupon/client.go
97:func Del(id string) error {

customer/client.go
150:func Del(id string) error {

discount/client.go
18:func Del(customerID string) error {
28:func DelSub(customerID, subscriptionID string) error {

invoiceitem/client.go
113:func Del(id string) error {

paymentsource/client.go
94:func Del(id string, params *stripe.CustomerSourceParams) error {

plan/client.go
113:func Del(id string) error {

recipient/client.go
144:func Del(id string) error {
```

IMO it _should_ actually just return the deleted resource as you have it, but we should be a little careful here because we'll be introducing a second standard for how a resource `Del` function should be written.

I'd suggest that if possible we move the cancellation work into a separate pull and try to get that merged post haste. I we still want to make the changes to `Del` though, we may want to open a separate pull and solicit a little more feedback before making a decision. What do you think?

remistr:
Closing this one as invalid. Creating a new PR for the subscription cancelation and then a separate one for any call to `Del`

PR #152, opened 2015-11-13
Return the resource when canceling a subscription

### LINKED ISSUE #150: Subscription.Cancel does not return the subscription object (per spec)

remistr:
When you cancel a subscription it now returns the subscription object
instead of just an error.

r? @brandur 

Fixes #150. Supersedes #151.

### DIFF
--- a/sub/client.go
+++ b/sub/client.go
@@ -151,11 +151,11 @@ func (c Client) Update(id string, params *stripe.SubParams) (*stripe.Sub, error)
 
 // Cancel removes a subscription.
 // For more details see https://stripe.com/docs/api#cancel_subscription.
-func Cancel(id string, params *stripe.SubParams) error {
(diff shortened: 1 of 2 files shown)


brandur:
Great job @remistr! Thank-you.

Let us know if you want to get that other pull request opened as well.

brandur:
Bumped major version to 9.0.0 and added change notes in 58113b6.

Issue #149, opened 2015-11-10
Add Relay support


steadicat:
What’s the ETA for adding support for Orders and Products? The [docs for Go](https://stripe.com/docs/api#orders) already show code samples using features that don’t seem implemented yet.

brandur:
Hi @steadicat! Thanks for logging this and our apologies for the delay in getting these features into the Go bindings.

Unfortunately I don't have an immediate update for you today, but please expect to see these integrated relatively soon. We'll bump this issue when they become available.

brandur:
Closed by #154. Let us know if you find any problems. Thanks!

PR #154, opened 2015-11-17
Add orders, products, and SKUs to the API.

rasmusrygaard:
r? @siddarth @spolu 
cc @brandur


brandur:
You guys rock! Let me know if you need an additional review or anything ...

Fixes #149. /cc @steadicat

rasmusrygaard:
@brandur If you have the bandwidth to look over this quickly that would be great. My main concern is that this has lived long enough internally that we've come to accept any quirks that may exist.

brandur:
@rasmusrygaard Cool! Will take a look in a bit.

> My main concern is that this has lived long enough internally that we've come to accept any quirks that may exist.

Yeah :/ Unfortunately, as with some of the other modules that have been added here, we may not be able to catch all the quirks that come up on the initial pass, but we'll try to handle them appropriately as they're discovered.

.gitignore
```diff
@@ -0,0 +1 @@
+*.test
```
spolu: Surprised to see this part of that PR?

spolu:
(otherwise LGTM as this has been pretty well tested in our own deployments)

.gitignore
rasmusrygaard: Happy to open something separate for this, but it doesn't seem like there's any .gitignore in place at the moment

spolu: I'm completely fine with it :)

brandur:
Looks good as far as I can tell! Will wait for a pass by @siddarth, but +1 for merging. Thanks again.

rasmusrygaard:
Going to :ship: here as @siddarth already did a pass while this was used internally

Issue #155, opened 2015-11-18
Shipping/Address type conversion hell


steadicat:
I’m having trouble understanding the logic of some of the types used in this API.

There are currently four nearly identical structs that contain shipping information (name, address, phone):

```Go
type Shipping struct {
    Name    string  `json:"name"`
    Address Address `json:"address"`
    Phone   string  `json:"phone"`
}

type CustomerShippingDetails struct {
    Name    string  `json:"name"`
    Address Address `json:"address"`
    Phone   string  `json:"phone"`
}

type ShippingParams struct {
    Name    string
    Address *AddressParams
    Phone   string
}

type ShippingDetails struct {
    Name     string  `json:"name"`
    Address  Address `json:"address"`
    Phone    string  `json:"phone"`
    Tracking string  `json:"tracking_number"`
    Carrier  string  `json:"carrier"`
}
```

I find myself having to convert between these *a lot*. I realize that in some cases Go lets you convert similar structures by casting, but that works only half the time because of subtle incompatibilities. For example `ShippingParams(Shipping)` fails because one takes `Address` and the other takes `AddressParams`. To make things worse, `Address` and `AddressParams` are also subtly incompatible:

```Go
type Address struct {
    Line1   string `json:"line1"`
    Line2   string `json:"line2"`
    City    string `json:"city"`
    State   string `json:"state"`
    Zip     string `json:"postal_code"`
    Country string `json:"country"`
}

type AddressParams struct {
    Line1      string
    Line2      string
    City       string
    State      string
    PostalCode string
    Country    string
}
```

Can you spot it? Yeah, `PostalCode` vs `Zip`. This causes me to have to copy every field manually in both `*Address` and the `*Shipping` that contains it, if e.g. I want to create a customer and an order (or a charge) using the same shipping info.

Is there any way to make these structs compatible? Or any other way to make working with these easier?

brandur:
@steadicat Thanks for getting this issue going!

+1 that the state of today is pretty bad. I think that in general though, we'll actually want to keep these structs different so that if some minor incompatibilities in addresses are introduced between endpoints (which already exist) we can deal with them in a sane way.

Want we need to fix though is to make these actually inconsistent in how they're named and referenced. My initial thought is that every type of resource should just get its own shipping type so that there's never any guesswork involved in what gets used. This might look like:

```
Shipping                -> OrderShipping
CustomerShippingDetails -> CustomerShipping
ShippingParams          -> OrderShipping
[...]

PR #158, opened 2015-11-20
Deprecate `Zip` in favor of `PostalCode`

brandur:
Deprecates an address' the `Zip` property in favor of `PostalCode`, but
for now maintains `Zip` for backward compatibility.

The impetus for this is that the bindings are already inconsistent with
the API in that this property is serialized under a different name, but
as described in #155 are also self-inconsistent in that the property is
just called `PostalCode` elsewhere.

Any thoughts on this one appreciated! Thanks!

### DIFF
--- a/account.go
+++ b/account.go
@@ -129,12 +129,15 @@ type LegalEntity struct {
 
 // Address is the structure for an account address.
 type Address struct {
-	Line1   string `json:"line1"`
-	Line2   string `json:"line2"`
-	City    string `json:"city"`
-	State   string `json:"state"`
-	Zip     string `json:"postal_code"`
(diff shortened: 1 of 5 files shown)


remistr:
@brandur The issue is that the API still clearly references `address_zip` in a lot of places, for `shipping` in Charges or Customers or on the Card object directly. I'd me more in favor of deprecating `PostalCode` instead if given the choice.

brandur:
@remistr Hm, looking at this again, it does seem like we are a little more standardized in the direction of "zip" than "postal code" (for better or for worse). I'm going to close this out for now then and possibly get an internal discussion going about fully moving in that direction. Thanks!

PR #159, opened 2015-11-21
Return the resource when calling Del

remistr:
* Deleting a resource now returns it as deleted.
* Added Deleted property to Account, BankAccount, Coupon, Card, Discount,
InvoiceItem, PaymentSource, Plan and Recipient resources.
* Added Deleted property to Card resource
* Added Customer property to BankAccount.
* Added Customer property to BankAccountParams.
* Added first test for BankAccount.
* You can now delete a bank account from a Customer.

r? @brandur

### DIFF
--- a/customer/client_test.go
+++ b/customer/client_test.go
@@ -159,13 +159,13 @@ func TestCustomerGet(t *testing.T) {
 func TestCustomerDel(t *testing.T) {
 	res, _ := New(nil)
 
-	c, err := Del(res.ID)
+	customerDel, err := Del(res.ID)
 
 	if err != nil {
 		t.Error(err)
 	}
 
-	if !c.Deleted {
+	if !customerDel.Deleted {
 		t.Errorf("Customer id %q expected to be marked as deleted on the returned resource\n", c.ID)
 	}
 
--- a/customer/client_test.go
+++ b/customer/client_test.go
@@ -159,10 +159,14 @@ func TestCustomerGet(t *testing.T) {
 func TestCustomerDel(t *testing.T) {
 	res, _ := New(nil)
 
-	err := Del(res.ID)
+	c, err := Del(res.ID)
 
 	if err != nil {
 		t.Error(err)
+	}
+
+	if !c.Deleted {
+		t.Errorf("Customer id %q expected to be marked as deleted on the returned resource\n", c.ID)
 	}
 
 	target, err := Get(res.ID, nil)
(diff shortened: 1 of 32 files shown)


brandur:
@remistr My apologies for the delay on this one! I'll definitely be able to take a pass on it tomorrow.

brandur:
OMG this is so boss! Thanks a lot @remistr!

@cosn / @kyleconroy  Could I solicit your opinions on this one? As you can probably tell, the overarching theme here is to add the `Deleted` property to any structs where it was missing, and to change all `Del` functions in the project to return a deleted object along with an `err`.

The idea was sparked by #150, which Remi addressed with a targeted patch in #152. The complication though is that really `Cancel` _is_ the `Del` of a subscription and it was obvious that originally no object had been returned because the function signature of `Del` had been re-used for it. This patch brings of the rest of the `Del` functions in-line with that update. A nice side effect is that we'll also be making the Go bindings more consistent with other languages like Ruby where a deleted object is made available.

I think we can all agree that in general breaking API changes are bad, but here are a few points to support this one:

* In general we've been quite liberal with breaking API changes in this project it seems. Anyone upgrading may already have to tweak a few calls.
* Despite the API change, the fix is trivial (`err := Del(...)` to `_, err := Del(...)`). The compiler will help developers find and fix these in seconds.

So ... thoughts? Thank-you!

customer/client_test.go
```diff
@@ -159,18 +159,22 @@ func TestCustomerGet(t *testing.T) {
 func TestCustomerDel(t *testing.T) {
 	res, _ := New(nil)
 
-	err := Del(res.ID)
+	c, err := Del(res.ID)
```
brandur: From what I can tell, it doesn't seem too idiomatic in these to use single letter variable names in places like this (also, `c` is very often used in this project to refer to a client). Do you think we should rename this to something like `resDel`?

example_test.go
```diff
@@ -70,11 +70,15 @@ func ExampleInvoice_update() {
 func ExampleCustomer_delete() {
 	stripe.Key = "sk_key"
 
-	err := customer.Del("acct_example_id")
+	c, err := customer.Del("acct_example_id")
```
brandur: And same single letter comment here.

customer/client_test.go
remistr: Totally fair. I did the customer ones first and later used the xxxDeleted convention. updating now

remistr:
@brandur Fixed the one letter variables and also found a discrepancy with the example format of customer's id.

cosn:
:+1: for this, we should just bump the major version to indicate a major change

kyleconroy:
We've been aggressive with breaking changes to this library, so I'm :+1: to making this change. I agree with Cos that this will need to be a major version bump.

brandur:
@cosn @kyleconroy Thanks for the input guys and of course thank-you @remistr for the awesome patch!

As suggested, I've bumped the major version to 10.0.0 in 9e24fce.

PR #160, opened 2015-11-22
Fix failing tests

remistr:
* Properly pass the Currency when creating an ExternalAccount.
* Only attempt to retrieve the plans the test just created
* Create a new product and SKU for each Order test. This might be an issue
over time as you can't delete Product and SKU.

r? @brandur

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -117,10 +117,10 @@ func TestAccountUpdateWithBankAccount(t *testing.T) {
 
 	params = &stripe.AccountParams{
 		ExternalAccount: &stripe.AccountExternalAccountParams{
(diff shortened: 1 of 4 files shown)


brandur:
@remistr Thanks a lot for taking a look at this! These tests could definitely use some love.

+1 from me. The one thing that I may have forgotten to mention to you is that it's pretty conventional to run `gofmt` before commit in order to normalize the style of the code. Can you run `gofmt -w .` before we pull this in? Thanks!

brandur:
Looks like gofmt was run so I'm pulling this in.

Thanks again for the great work on this one Remi!

PR #161, opened 2015-11-26
Include country on account requests

telecoda:
Creating an account with a different country to the platform account causes an error

type: "invalid_request_error"
message: "Address for business must match account country"
param: "legal_entity[address][country]"

This is because the country parameter is not being populated in the API request.


brandur:
Sorry about the trouble here, and thank-you so much for the patch @telecoda! This looks great.

PR #167, opened 2015-12-04
closes #166

### LINKED ISSUE #166: Add card to Account not consistent with docs or not Implemented
From the Docs:

```
stripe.Key = "sk_test_12345"

c, err := card.New(&stripe.CardParams{
  Account: "acct_1649BKGdIrMVkPmQ",
  Token: "",
}
```

This code throws an error on compile time: "unknown stripe.CardParams field 'Account' in struct literal"

There is an Account field on stripe.CardParams.Params so this compiles:
```
	cardArgs := &stripe.CardParams{
		Token: stripeToken,
	}
	cardArgs.Params.SetAccount(id)

       return card.New(cardArgs)
```

However it throws this error: "Invalid card params: either customer or recipient need to be set"

This is thrown because [card.New](https://github.com/stripe/stripe-go/blob/master/card/client.go#L43-L60) is looking explicitly for the presence of Customer or Recipient.

Looking deeper, stripe.CardParams.Params is not where we want Account as all these fields are [wrapped in metadata[...]](https://github.com/stripe/stripe-go/blob/master/params.go#L103-L117) for the request.

collinglass:


card.go
```diff
@@ -24,8 +24,8 @@ type CardFunding string
 // For more details see https://stripe.com/docs/api#create_card and https://stripe.com/docs/api#update_card.
 type CardParams struct {
 	Params
-	Token                                         string
-	Customer, Recipient                           string
+	External_Account, Token                       string
```
brandur: I don't think we need an `External_Account` here.

bankaccount/client.go
```diff
@@ -108,15 +107,7 @@ func Del(id string, params *stripe.BankAccountParams) (*stripe.BankAccount, erro
...
-	var err error
-
-	if len(params.Customer) > 0 {
-		err = c.B.Call("DELETE", fmt.Sprintf("/customers/%v/bank_accounts/%v", params.Customer, id), c.Key, nil, nil, ba)
-	} else if len(params.AccountID) > 0 {
-		err = c.B.Call("DELETE", fmt.Sprintf("/accounts/%v/bank_accounts/%v", params.AccountID, id), c.Key, nil, nil, ba)
-	} else {
-		err = errors.New("Invalid bank account params: either customer or AccountID need to be set")
-	}
+	err := c.B.Call("DELETE", fmt.Sprintf("/accounts/%v/bank_accounts/%v", params.AccountID, id), c.Key, nil, nil, ba)
```
brandur: Hm. Is there any particular reason that you collapsed these lines?

brandur:
@collinglass Hey! I left a couple comments, but this generally looks good! Thanks for taking a stab at it.

It looks like you figured it out already, but the `Account` property under the parent `Params` struct is for authentication rather than referencing an account during an operation.

bankaccount/client.go
collinglass: It's the only function in the client that handled Customers, I made an educated assumption. It should have been in another commit though

card.go
collinglass: How do you add a debit card to an account?

brandur: Ah, I see. I think you you may have misreferenced the new field in `AppendDetails` below though. As far as I can tell, this is new used. Also, I don't think it's a string unfortunately ([create a card](https://stripe.com/docs/api/curl#create_card-source%20|%20external_account)).

bankaccount/client.go
brandur: Cool! Let's maybe leave this one out for now. As far as I can tell it's actually the `/customers` version that's non-deprecated.

card.go
collinglass: How would you handle this? From that link it tells me it's either a token or a dictionary (similar to the current implementation for customer). Second, why is ```AppendDetails``` using card instead of ```source```?

collinglass: When I remove the External_Account Param I get the follow error response:

{"type":"invalid_request_error","message":"Missing required param: external_account.","param":"external_account","request_id":"req_foo_mustache","status":400}

When I use that param, I get a successful request.

collinglass: I see my confusion :) Thanks for your patience.

collinglass:
@brandur I made the suggested changes. Is there anything else needed on my part to get this merged?

card.go
brandur: > How would you handle this? From that link it tells me it's either a token or a dictionary (similar to the current implementation for customer).

Good question. I notice that the bindings seem to have just not bothered for support on the dictionary version for customer, so it seems fine to leave it out of here for now too.

brandur:
> @brandur I made the suggested changes. Is there anything else needed on my part to get this merged?

Sorry about the delay.

I still have some doubts around the naming of `Account` in that it seems that `ExternalAccount` might also be a good choice, but it seems that not even in the API is this anywhere near consistent and because the field on the card object is called `account`, the proposed naming here seems fine.

Thanks for the patch!

PR #168, opened 2015-12-05
Add a method for updating a coupon

### LINKED ISSUE #165: Cannot update existing coupon
There is no `Update()` method for coupons. I understand the oversight, since I think the only thing that is modifiable is `Meta`, but regardless it would be useful.

brandur:
Fixes #165.

/cc @remistr Mind giving this a quick review pass? Thanks!

### DIFF
--- a/coupon/client.go
+++ b/coupon/client.go
@@ -92,6 +92,23 @@ func (c Client) Get(id string, params *stripe.CouponParams) (*stripe.Coupon, err
 	return coupon, err
 }
 
+// Update updates a coupon's properties.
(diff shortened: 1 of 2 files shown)


remistr:
@brandur: I'd test that the metadata is present on the updated coupon to make sure it worked by adding
```
if target.Meta["key"] != updateParams.Meta["key"] {
	t.Errorf("Meta %v does not match expected Meta %v\n", target.Meta, updateParams.Meta)
}
```

Other than that :ship:

brandur:
@remistr Cool, +1. Done in ba774cc. Thanks!

PR #169, opened 2015-12-08
Handling External accounts in account response

collinglass:
Implementation for [external_accounts](https://stripe.com/docs/api#account_object-external_accounts) property on account response.

What is the preferred approach to handling the ambiguity of the external accounts array?

_Edit by @brandur:_ Here's [a link to the diff without whitespace changes](https://github.com/stripe/stripe-go/pull/169/files?w=0) which is easier to read.

### DIFF
--- a/account.go
+++ b/account.go
@@ -73,23 +73,24 @@ type Account struct {
 	ChargesEnabled bool   `json:"charges_enabled"`
 	Country        string `json:"country"`
 	// Currencies is the list of supported currencies.
-	Currencies           []string `json:"currencies_supported"`
-	DefaultCurrency      string   `json:"default_currency"`
-	DetailsSubmitted     bool     `json:"details_submitted"`
-	TransfersEnabled     bool     `json:"transfers_enabled"`
-	Name                 string   `json:"display_name"`
-	Email                string   `json:"email"`
-	Statement            string   `json:"statement_descriptor"`
-	Timezone             string   `json:"timezone"`
-	BusinessName         string   `json:"business_name"`
-	BusinessPrimaryColor string   `json:"business_primary_color"`
-	BusinessUrl          string   `json:"business_url"`
-	SupportPhone         string   `json:"support_phone"`
-	SupportEmail         string   `json:"support_email"`
-	SupportUrl           string   `json:"support_url"`
-	ProductDesc          string   `json:"product_description"`
-	Managed              bool     `json:"managed"`
-	DebitNegativeBal     bool     `json:"debit_negative_balances"`
+	Currencies           []string         `json:"currencies_supported"`
+	DefaultCurrency      string           `json:"default_currency"`
+	DetailsSubmitted     bool             `json:"details_submitted"`
+	TransfersEnabled     bool             `json:"transfers_enabled"`
+	Name                 string           `json:"display_name"`
+	Email                string           `json:"email"`
+	ExternalAccounts     ExternalAccounts `json:"external_accounts"`
+	Statement            string           `json:"statement_descriptor"`
+	Timezone             string           `json:"timezone"`
+	BusinessName         string           `json:"business_name"`
+	BusinessPrimaryColor string           `json:"business_primary_color"`
+	BusinessUrl          string           `json:"business_url"`
+	SupportPhone         string           `json:"support_phone"`
+	SupportEmail         string           `json:"support_email"`
+	SupportUrl           string           `json:"support_url"`
+	ProductDesc          string           `json:"product_description"`
+	Managed              bool             `json:"managed"`
+	DebitNegativeBal     bool             `json:"debit_negative_balances"`
 	Keys                 *struct {
 		Secret  string `json:"secret"`
 		Publish string `json:"publishable"`
@@ -109,6 +110,33 @@ type Account struct {
 	} `json:"tos_acceptance"`
 	SupportAddress *Address `json:"support_address"`
 	Deleted        bool     `json:"deleted"`
+}
+
+type ExternalAccounts struct {
+	Object     string            `json:"object"`
+	Data       []ExternalAccount `json:"data"`
+	HasMore    bool              `json:"has_more"`
+	TotalCount uint              `json:"total_count"`
+	Url        string            `json:"url"`
+}
+
+type ExternalAccount struct {
+	BankAccount *BankAccount
+	Card        *Card
+}
+
+func (ea *ExternalAccount) UnmarshalJSON(b []byte) (err error) {
+	bank, card := BankAccount{}, Card{}
+	if err = json.Unmarshal(b, &bank); err == nil {
+		ea.BankAccount = &bank
+		return
+	}
+	if err = json.Unmarshal(b, &card); err == nil {
+		ea.Card = &card
+		return
+	}
+
+	return
 }
 
 // LegalEntity is the structure for properties related to an account's legal state.


account.go
```diff
@@ -111,6 +112,33 @@ type Account struct {
 	Deleted        bool     `json:"deleted"`
 }
 
+type ExternalAccounts struct {
```
brandur: I think in this particular case it might be a good idea to re-use the fields in `ListMeta` and to rename `Data` to `Values` for consistency. See [this example in cards](https://github.com/stripe/stripe-go/blob/5d35501411915df467df98dcfebabfde5bba9813/card.go#L68,L71).

account.go
```diff
@@ -111,6 +112,33 @@ type Account struct {
...
+}
+
+type ExternalAccount struct {
+	BankAccount *BankAccount
+	Card        *Card
+}
+
+func (ea *ExternalAccount) UnmarshalJSON(b []byte) (err error) {
+	bank, card := BankAccount{}, Card{}
+	if err = json.Unmarshal(b, &bank); err == nil {
```
brandur: Would this actually return an error? It might just fill the field's that it's able to and return a card as a bank account. We'll certainly need to add some tests to verify that one way or another :$

Unfortunately, we may need to do an initial deserialization pass to a `map[string]interface{}` and then check its object property before deserializing it again appropriately.

brandur:
> What is the preferred approach to handling the ambiguity of the external accounts array?

Good question. Just to provide a little more information, the property in question here is [`external_accounts` on the account object](https://stripe.com/docs/api/curl#account_object-external_accounts) which has been made polymorphic so that members can be bank accounts or cards. Documentation:

> The list contains all external accounts that have been attached to the Stripe account. These may be bank accounts or cards.

As far as I know, we haven't really had to do this anywhere else. For better or for worse, the right way to go here may be an interface that represents an external account and which is implemented by both `BankAccount` and `Card`. Possibly the only common properties will be `ID`, `Name`, and `Deleted`. The major downside here of course is that this is yet more edge casing that will be difficult to remove in a world of generated bindings.

@cosn @kyleconroy This is a large-ish change. Any thoughts?

brandur:
@collinglass And of course thank-you for putting this patch together!

account.go
cosn: +1

cosn:
I think I personally like this better because as the user, I can check which is non-nil and access the properties, but with the interface approach, I have to now also cast since it's likely that I'm interested in some properties that are not shared.

kyleconroy:
Agreed with Cos. An interface really wouldn't help here, as most people are interested in getting the properties on the struct instead of calling methods.

brandur:
@cosn Cool, thanks!

I'm amenable to a number of approaches here so if we decide to go with the "struct wrapper", cool.

Here's one more option that just came to me: try to avoid a special case here by generalizing a little bit. Define a `StripeObject` interface with a `Type()` function which could then be used to easily differentiate types from within a `switch`. Here's a working sample:

``` go
package main

import "fmt"

type StripeType int

const (
	TypeBankAccount StripeType = iota
	TypeCard
)

type StripeObject interface {
	Type() StripeType
}

type BankAccount struct {
}

func (ba *BankAccount) Type() StripeType {
	return TypeBankAccount
}

type Card struct {
}

func (c *Card) Type() StripeType {
	return TypeCard
}

func main() {
	var obj StripeObject
	obj = &BankAccount{}

	switch {
	case obj.Type() == TypeBankAccount:
		fmt.Printf("Got bank account\n")
	case obj.Type() == TypeCard:
		fmt.Printf("Got card\n")
	}
}
```

cosn:
Sure that works, but not sure what problem it's solving though. We'd still need to have the wrapper struct with multiple properties of different types, so is this just changing the `nil` check for a type check?

brandur:
> Sure that works, but not sure what problem it's solving though. We'd still need to have the wrapper struct with multiple properties of different types, so is this just changing the nil check for a type check?

The idea is to kill this wrapper type:

``` go
type ExternalAccount struct {
	BankAccount *BankAccount
	Card        *Card
}
```

By changing this:

``` go
type ExternalAccounts struct {
	Data []ExternalAccount `json:"data"`
}
```

To this:

``` go
type ExternalAccounts struct {
	Data []StripeObject `json:"data"`
}
```

You still have a net gain of a type (`StripeObject`), but you avoid the gnarly wrapper and hopefully the `StripeObject` abstraction is general enough that it could be reapplied for other purposes down the line.

kyleconroy:
Would it be possible to just embed a Card and a BankAccount in the ExternalAccount struct?

collinglass:
Out of these two options I prefer the embeded because the data is important to me. I will always have to cast to use the object.

brandur:
> Would it be possible to just embed a Card and a BankAccount in the ExternalAccount struct?

@kyleconroy Yeah, certainly possible. I think that's what's proposed in the original patch here.

> Out of these two options I prefer the embeded because the data is important to me. I will always have to cast to use the object.

@collinglass Again, not ruling anything out, but is casting really all that different? Here are what the usages of each approach might look like:

``` go
	// cast
	switch {
	case obj.Type() == TypeBankAccount:
		obj.(*BankAccount)
	case obj.Type() == TypeCard:
		obj.(*Card)
	}

	// nil check
	switch {
	case obj.BankAccount != nil:
		obj.BankAccount
	case obj.Card != nil
		obj.Card
	}
```

I suppose some argument could be made that casts are inherently risky, but practically speaking a forgotten check on a possible `nil` is just as threatening to your program as a miscast.

I don't want to push any particular solution too hard, but one nice thing about the casting method is that it has some nice characteristics for maintenance that the wrapper type doesn't:

* If a new type gets added to a polymorphic list, nothing changes for a slice of `StripeObject` instances, but the wrapper must be extended with a new property to support it.
* If a new polymorphic list is introduced somewhere in the API, this may just look like a new list of `StripeObject` instances, but with the wrapper approach a new type must be added.

kyleconroy:
Switching on a `Type` method doesn't feel like usual Go. It would probably be better to define a shared interface and then do a type swtich.

```go
switch obj := obj.(type) {
default:
    fmt.Printf("unexpected type %T\n", t)     // %T prints whatever type t has
case stripe.BankAccount:
    fmt.Printf("bank %+v\n", t)             // t has type BankAccount
case stripe.Card:
    fmt.Printf("card %+v\n", t)             // t has type Card
}
```

brandur:
@kyleconroy Cool! That's definitely a more refined implementation. Thanks.

One counterargument for `StripeObject` that just struck me: a user may have to read docs to actually know what kind of objects may be part of the list whereas with a wrapper struct it's somewhat discoverable by looking at types (even if the implementation may be outdated).

kyleconroy:
Another object is to get rid of the external account wrapper struct by changing the external account list struct

```go
type ExternalAccounts struct {
	Object     string            `json:"object"`
	BankAccount       []BankAccount
        Cards       []Card
	HasMore    bool              `json:"has_more"`
	TotalCount uint              `json:"total_count"`
	Url        string            `json:"url"`
}
```

brandur:
@kyleconroy True! I think the biggest problem with that approach is that you need to a different `for` loop for every different type in the list (with the additional risk of forgetting a type), but it does introduce some extra safety. Hm ...

brandur:
Hey @collinglass, I've cherry picked your commit into another branch and am going to supersede this pull with the one in #206. Thanks for laying the groundwork here!

Regarding the discussion above, here's the interface that I went with in the end:

``` go
// ExternalAccountList is a list of external accounts that may be either bank
// accounts or cards.
type ExternalAccountList struct {
	ListMeta

	// BankAccountValues contains any bank accounts that are part of the list.
	BankAccountValues []*BankAccount

	// CardValues contains any cards that are part of the list.
	CardValues []*Card
}
```

Issue #164, opened 2015-12-04
Standard syntax for referencing AccountID, Customer, Recipient


collinglass:
There should be a standard syntax for referencing the structs, either:

Account, Customer, Recipient

OR

AccountID, CustomerID, RecipientID

brandur:
Yes! Agreed that this has unfortunately become a huge mess.

CC @stripe/api-libraries Mini-proposal: make parameters _without_ an "ID" suffix the one true way forward because that's how they look in the actual API and because many of the parameters are already in that format. I don't think renaming existing *ID parameters is worse it for now, but we should consider if we ever need to make a large breaking change at some point in the future. Thoughts?

kyleconroy:
== 

Since most of these resources can be expanded, we should always use a struct instead of an ID.

brandur:
@kyleconroy Cool, thanks!

@collinglass I'm going to close this out for now, but we'll keep an eye open for it during the review of future integrations. Thanks for bringing the problem to our attention.

collinglass:
No problem, the [BankAccountParams](https://github.com/stripe/stripe-go/blob/master/bankaccount.go#L14-L19) has Account and AccountID, which is particularly confusing.

brandur:
> No problem, the BankAccountParams has Account and AccountID, which is particularly confusing.

Yeah, that's pretty bad. I've added a small amount of documentation in #170 to try and improve this situation a bit.

PR #170, opened 2015-12-08
Provide some nominal documentation for `BankAccountParams`

brandur:
As mentioned in #164, it's pretty confusing that these parameters
contain both an `Account` and an `AccountID` with neither being
documented. To make matters worse, `Account` was improperly named
and should actually be called `AccountNumber`. This patch adds a small
amount of documentation to help point out the distinction.

/cc @remistr Mind giving this one a quick review?

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -18,7 +18,8 @@ type BankAccountParams struct {
 	// nested.
 	AccountID string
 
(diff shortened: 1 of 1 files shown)


remistr:
@brandur :+1:

brandur:
@remistr Thanks!

PR #173, opened 2015-12-21
add default to bank account

collinglass:
[Reference](https://stripe.com/docs/api#bank_account_object-default_for_currency)


brandur:
Nice! Thanks a lot Collin, and sorry about the delay in getting this in.

PR #174, opened 2015-12-22
Fix README.md

chansuke:
`cd stripe` to `cd stripe-go`


brandur:
Oops! Thank-you for that!

PR #175, opened 2015-12-22
Fix Order.New commonParams issue

dominicphillips:
Had an issue creating orders for accounts, the SetAccount would not work. Figured it had something to do with the params not being passed to the request. This PR fixes this issue.


brandur:
Great catch. Thanks for sending this back upstream Dominic!

PR #177, opened 2016-01-06
Add OrderParams to order Get

dominicphillips:
This PR adds OrderParams to the order Get call, this way its possible to retrieve orders that have been created on another account. Without the params there is no way to set the account and the order will 404.

### DIFF
--- a/order/client.go
+++ b/order/client.go
@@ -213,14 +213,22 @@ func (c Client) Pay(id string, params *stripe.OrderPayParams) (*stripe.Order, er
 
 // Get returns the details of an order
 // For more details see https://stripe.com/docs/api#retrieve_order.
-func Get(id string) (*stripe.Order, error) {
-	return getC().Get(id)
+func Get(id string, params *stripe.OrderParams) (*stripe.Order, error) {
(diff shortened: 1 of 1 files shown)


brandur:
Code looks good to me. Judging by the Go code example in the docs, this method may have been intended to have a second parameter anyway:

``` go
stripe.Key = "sk_test_..."
o, err := order.Get("or_17Q7aL2eZvKYlo2CTnNMY3Nw", nil)
```

@siddarth Would you mind taking a quick review pass on this one as well? Thanks!

brandur:
Oh and @dominicphillips, thanks for the patch! If it's not too much trouble, would you mind rebasing against master so we can get rid of that merge commit? Thanks!

dominicphillips:
@brandur - Yes, totally! I mean, the use case, creating orders for accounts and retrieving is pretty standard use case for relay, I reckon. At least, thats what we do!

dominicphillips:
@brandur rebased

brandur:
@dominicphillips Thank-you! Pulling this in.

PR #178, opened 2016-01-06
Support special `now` value for trial end on subscription update

### LINKED ISSUE #176: Sub.Update does not support the TrialEndNow bool
It looks like you can't use the special `trial_end=now` value when calling `Sub.Update()` (only appears to be supported by `Sub.New()`.

brandur:
This was previously accepted on subscription creations, but not updates.
The [docs][docs] on update indicate that it can be used in both places:

> Unix timestamp representing the end of the trial period the customer
> will get before being charged for the first time. If set, trial_end
> will override the default trial period of the plan the customer is
> being subscribed to. The special value now can be provided to end the
> customer's trial immediately.

Fixes #176.

[docs]: https://stripe.com/docs/api#update_subscription

### DIFF
--- a/sub/client.go
+++ b/sub/client.go
@@ -130,7 +130,9 @@ func (c Client) Update(id string, params *stripe.SubParams) (*stripe.Sub, error)
 		body.Add("coupon", params.Coupon)
 	}
 
-	if params.TrialEnd > 0 {
+	if params.TrialEndNow {
+		body.Add("trial_end", "now")
+	} else if params.TrialEnd > 0 {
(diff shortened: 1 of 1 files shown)


brandur:
Regarding tests: unfortunately the update test is already testing this parameter, but lacks (along with the rest of the test suite), a sane way of checking the response without building out a special test (that hits production) for every possible parameter. Going to leave it for now with the hope that we can fix the more general problem at some point.

cosn:
:+1:

PR #179, opened 2016-01-11
Allow update of the country field in customer shipping details.

JonathanHarford:
Attempts to update the country of a customer's shipping address were previously setting the country to null.


brandur:
@JonathanHarford Thank-you! It also looks like it was missing from the shipping on a charge as well. I'll send another pull to get that updated.

PR #180, opened 2016-01-11
Adds "country" to charge shipping details

brandur:
This follows up #179.


brandur:
Released this and #179 as 11.0.1 in ba7a1f4.

PR #183, opened 2016-01-25
Support deleting SKUs and products.

rasmusrygaard:
r? @brandur 
cc @jimdanz 

API endpoints: [SKU deletion](https://stripe.com/docs/api#delete_sku), [product deletion](https://stripe.com/docs/api#delete_product)

### DIFF
--- a/product/client_test.go
+++ b/product/client_test.go
@@ -134,7 +134,7 @@ func TestProductUpdate(t *testing.T) {
 
 func TestProductList(t *testing.T) {
 	const runs = 3
-	randID := fmt.Sprintf("example.com/", RandSeq(16))
+	randID := fmt.Sprintf("example.com/%s", RandSeq(16))
 	for i := 0; i < runs; i++ {
 		shippable := i == 0
 		active := i == 1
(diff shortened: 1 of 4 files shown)


product/client_test.go
```diff
@@ -134,7 +134,7 @@ func TestProductUpdate(t *testing.T) {
 
 func TestProductList(t *testing.T) {
 	const runs = 3
-	randID := fmt.Sprintf("example.com/", RandSeq(16))
+	randID := fmt.Sprintf("example.com/%s", RandSeq(16))
```
brandur: Oops, thanks for the fix!

brandur:
Looks good to me @rasmusrygaard!

Just as a general sanity check: can you confirm that the new tests looked okay when running locally?

rasmusrygaard:
Yup, local tests are looking good!

brandur:
Released as 11.1.0 in https://github.com/stripe/stripe-go/commit/8c57a8d3acee8cbc025e897bcaf2d9c5b50060ac.

PR #185, opened 2016-01-28
Allow a recipient's bank account to be updated by token

### LINKED ISSUE #184: Cannot update recipient's bank account using token
Based on https://stripe.com/docs/api#update_recipient it supposed to be possible to update recipient's bank account using token. It is possible via ruby version of stripe API.

In package recipient **New** method has

```
	if params.Bank != nil {
		if len(params.Bank.Token) > 0 {
			body.Add("bank_account", params.Bank.Token)
		} else {
			params.Bank.AppendDetails(body)
		}
	}
```

But **Update** method has only

```
		if params.Bank != nil {
			params.Bank.AppendDetails(body)
		}
```

Please fix **Update** method to accept Token

brandur:
As described in #184, we allow a bank account to be referenced by a
token when creating a new recipient, but we currently don't allow the
same when updating a recipient. This patch addresses that problem.

Fixes #184.

/cc @remistr Would you mind taking a review pass on this one? Thanks!

### DIFF
--- a/recipient/client.go
+++ b/recipient/client.go
@@ -105,7 +105,11 @@ func (c Client) Update(id string, params *stripe.RecipientParams) (*stripe.Recip
 		}
 
 		if params.Bank != nil {
-			params.Bank.AppendDetails(body)
+			if len(params.Bank.Token) > 0 {
+				body.Add("bank_account", params.Bank.Token)
+			} else {
+				params.Bank.AppendDetails(body)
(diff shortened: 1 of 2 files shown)


remistr:
Modified the test to update with and without a bank account token

r? @brandur

brandur:
Excellent. Thanks @remistr!

PR #189, opened 2016-02-06
Added support for CountrySpec

remistr:
r? @brandur 
Wasn't sure if sub-types would have been better for each `hash` or not. Let me know!

### DIFF
--- a/countryspec.go
+++ b/countryspec.go
@@ -1,0 +1,16 @@
+package stripe
+
+// CountrySpec is the resource representing the rules required for a Stripe account.
+// For more details see https://stripe.com/docs/api/#country_specs.
+type CountrySpec struct {
+	ID                             string                         `json:"id"`
+	SupportedBankAccountCurrencies map[string][]string            `json:"supported_bank_account_currencies"`
+	SupportedPaymentCurrencies     []string                       `json:"supported_payment_currencies"`
+	SupportedPaymentMethods        []string                       `json:"supported_payment_methods"`
+	VerificationFields             map[string]map[string][]string `json:"verification_fields"`
+}
+
+// CountrySpecListParams are the parameters allowed during CountrySpec listing.
+type CountrySpecListParams struct {
+	ListParams
+}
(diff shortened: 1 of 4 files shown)


countryspec/client.go
```diff
@@ -0,0 +1,78 @@
...
+	return countrySpec, err
+}
+
+// List lists available CountrySpecs.
+func List(params *stripe.CountrySpecListParams) *Iter {
+	return getC().List(params)
+}
+
+func (c Client) List(params *stripe.CountrySpecListParams) *Iter {
+	type CountrySpecList struct {
```
brandur: Fairly small, but at least a few of the other methods start with a lower case letter (i.e. `countrySpecList`). I don't think it matters here, but it normally conveys a "non-exported" name (i.e. internal to the package).

countryspec.go
```diff
@@ -0,0 +1,16 @@
+package stripe
+
+// CountrySpec is the resource representing the rules required for a Stripe account.
+// For more details see https://stripe.com/docs/api/#country_specs.
+type CountrySpec struct {
+	ID                             string                         `json:"id"`
+	SupportedBankAccountCurrencies map[string][]string            `json:"supported_bank_account_currencies"`
+	SupportedPaymentCurrencies     []string                       `json:"supported_payment_currencies"`
+	SupportedPaymentMethods        []string                       `json:"supported_payment_methods"`
+	VerificationFields             map[string]map[string][]string `json:"verification_fields"`
```
brandur: Not a big one, but looking at some other models like `Charge` ([see here](https://github.com/stripe/stripe-go/blob/e4ca8d9583a2be58c844e890d2dbf85457aaace4/charge.go#L58,L84)), it seems like there's a slight preference here for defining new types instead of using long untyped constructs (e.g. `map[string]map[string][]string`). This is a little more convenient for those who have auto-complete systems set up their editors because the structure becomes self-describing.

So for example we might consider even:

``` go
map[Currency][]string
```

Or:

``` go
map[AccountType]VerificationFields
```

With something like:

```
type AccountType string

type VerificationFields struct {
    AdditionalFields []string
    MinimumFields []string
}
```

Thoughts?

brandur:
@remistr Left some minor comments on this one, but awesome work! Thanks.

PR #192, opened 2016-02-12
Added support for source_transaction and source_type for Transfers

remistr:
r? @brandur

### DIFF
--- a/transfer/client.go
+++ b/transfer/client.go
@@ -16,8 +16,14 @@ const (
 	Failed   stripe.TransferStatus = "failed"
 	Canceled stripe.TransferStatus = "canceled"
 
-	Card stripe.TransferType = "card"
-	Bank stripe.TransferType = "bank_account"
+	Card          stripe.TransferType = "card"
+	Bank          stripe.TransferType = "bank_account"
+	StripeAccount stripe.TransferType = "stripe_account"
+
+	SourceAlipay  stripe.TransferSourceType = "alipay_account"
+	SourceBank    stripe.TransferSourceType = "bank_account"
+	SourceBitcoin stripe.TransferSourceType = "bitcoin_receiver"
+	SourceCard    stripe.TransferSourceType = "card"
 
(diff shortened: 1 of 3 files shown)


transfer/client.go
```diff
@@ -16,8 +16,14 @@ const (
...
-	Card stripe.TransferType = "card"
-	Bank stripe.TransferType = "bank_account"
+	Card          stripe.TransferType = "card"
+	Bank          stripe.TransferType = "bank_account"
+	StripeAccount stripe.TransferType = "stripe_account"
+
+	SourceAlipay  stripe.TransferSourceType = "alipay_account"
+	SourceBank    stripe.TransferSourceType = "bank_account"
+	SourceBitcoin stripe.TransferSourceType = "bitcoin_receiver"
+	SourceCard    stripe.TransferSourceType = "card"
```
brandur: I'm still overall a little uncertain about the ideal naming/location of constants like these ones. Obviously the transfer type constants are already in here, but we can see similar constants from other resources, [PaymentSourceType](payment_source.go) for example, located right at the top level of the package.

What do you think about giving these a full `TransferSource` type prefix for now just as a hedge against future re-organization? (The current `Source` fits the naming in `Transfer`, but could be read as a little generic.)

brandur:
Awesome! Left a minor comment above, but looks great. Thanks Remi!

brandur:
Hey @remistr! Mind taking another look at this one to see whether it's still worthwhile to get in? Let me know if you're pretty constrained for time and I can take a look.

brandur:
I'm taking over for Remi on this one: `source_transaction` came in on another branch, so I've rebased and removed it from this PR.

Regarding my comment on naming above: I've taken another look at this now, and given that these names are qualified by the `transfer` package, I think that what Remi did originally was the right thing. My mistake!

Pulling this in.

brandur:
Released as 13.6.0.

PR #194, opened 2016-02-12
Fix a few broken tests

remistr:
r? @brandur


brandur:
@remistr +1 on the sentiment! It seems that this pull may have left CI a bit broken though (I tried a rebuild, but it didn't fix it). Mind taking a look?

brandur:
And another rebuild fixed it ...

Alright, we should get that fixed, but this looks good. Thanks @remistr!

PR #197, opened 2016-02-22
Allow `Account` to be specified for `ListParams`

### LINKED ISSUE #196: How do we fetch balance transactions for a managed account?
I am trying to fetch a list of all transactions based on the [List All Balance History](https://stripe.com/docs/api/go#balance_history) API. However, I want do it for a particular managed account. 

I see that Params struct has a `Account` field that is used to set the Account header.
```
// Params is the structure that contains the common properties
// of any *Params structure.
type Params struct {
	Exp                     []string
	Meta                    map[string]string
	Extra                   url.Values
	IdempotencyKey, Account string
}
```
```
if account := strings.TrimSpace(params.Account); account != "" {
    req.Header.Add("Stripe-Account", account)
}
```

However, I could not find an `Account` field for the ListParams. Is there another way to set the `Stripe-Account` header for all the `$resource$.List(stripe.$Resource$ListParams)` calls?

brandur:
As discussed in #196, we cannot currently send an account header when
using list endpoints. I believe that this was a known limitation of the
original implementation that was never addressed because of subtle
differences in the way `Params` and `ListParams` are used (the former is
serialized around an `http.Request` object while the latter is just used
to build a set of parameter values).

This patch adds an `Account` field to `ListParams` and suggests that
it's injected in list callbacks using a new `ToParams` function on
`ListParams`. This implementation is not complete, and just demonstrates
its use on a single endpoint. The new field will have to be added to all
list callbacks to be fully workable.

@remistr Thoughts on this implementation? Obviously not currently complete and
missing tests, but what do you think about the overall approach?

Fixes #196.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,19 @@ stripe := &client.API{}
 stripe.Init("access_token", nil)
 ```
 
+In order to make a request on behalf of a managed account, use the
+`StripeAccount` field on a `ListParams` or `Params` class. For example:
+
+```go
+// For a list request
+listParams := &stripe.ChargeListParams{StripeAccount: merchantID}
+```
+
+```go
+// For any other kind of request
+params := &stripe.CustomerParams{StripeAccount: merchantID}
+```
+
 ### Google AppEngine
 
 If you're running the client in a Google AppEngine environment, you'll need to
--- a/README.md
+++ b/README.md
@@ -97,10 +97,27 @@ for i.Next() {
 Alternatively, you can use the `even.Data.Raw` property to unmarshal to the
 appropriate struct.
 
-### Connect Flows
+### Authentication with Connect
 
-If you're using an access token you will need to use a client. Simply pass the
-access token value as the `tok` when initializing the client.
+There are two ways of authenticating requests when performing actions on behalf
+of a connected account, one that uses the `Stripe-Account` header containing an
+account's ID, and one that uses the account's keys. Usually the former is the
+recommended approach. [See the documentation for more information][connect].
+
+To use the `Stripe-Account` approach, pass the `StripeAccount` field to a
+`ListParams` or `Params` class. For example:
+
+```go
+// For a list request
+listParams := &stripe.ChargeListParams{StripeAccount: merchantID}
+```
+
+```go
+// For any other kind of request
+params := &stripe.CustomerParams{StripeAccount: merchantID}
+```
+
+To use a key, pass it to `API`'s `Init` function:
 
 ```go
 
@@ -111,19 +128,6 @@ import (
 
 stripe := &client.API{}
 stripe.Init("access_token", nil)
-```
-
-In order to make a request on behalf of a managed account, use the
-`StripeAccount` field on a `ListParams` or `Params` class. For example:
-
-```go
-// For a list request
-listParams := &stripe.ChargeListParams{StripeAccount: merchantID}
-```
-
-```go
-// For any other kind of request
-params := &stripe.CustomerParams{StripeAccount: merchantID}
 ```
 
 ### Google AppEngine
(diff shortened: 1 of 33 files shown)


brandur:
@remistr's been pretty busy this week, so this one is now up for grabs for anyone who feels like reviewing some Go! Thanks :)

/cc @stripe/api-libraries

mmcgrana:
Generally this seems reasonable to me.

There are a couple account-shaped fields already in some `*ListParams` types:

* `AccountID` in `BankAccountListParams`
* `Account` in `CardListParams`

Which could cause confusion since these types would get a new `Account` field via the embedded `ListParams` struct. We should be sure we understand how these existing fields relate to the new one. I _think_ these are referring to a different account but I'm not positive.

AFICT we don't really document how to use the Go client with connected accounts, at least on a per-request basis with an `Account` param. (The closest thing I see is https://github.com/stripe/stripe-go#connect-flows.) Would it be worth adding that while we're here? This might look for example like a bit in the readme, inline docs for `Account` in `Params` (it's documented well in `ListParams` in this PR), and perhaps some clarifying comments re: the other `Account`/`AccountID` fields above. 

cc @cosn in case he has some other state.

cosn:
I recall adding some `Account`-related things to the API for Connect ~a year ago, mainly as a header to be able to work on different APIs.

brandur:
Thanks for the detailed review here Mark!

> There are a couple account-shaped fields already in some *ListParams types:

Wow, great catch. Yet another couple examples of some truly unfortunate naming. The `Account` parameter in particular is very awkward because I don't think we can reclaim it even if we're willing to give up backward compatibility – a change would still compile because of the new addition at the parent's level, and just now do the wrong thing.

That leaves one option: pick something more elaborate for the new parameter name to avoid any other account-related collisions. `StripeAccount` or `StripeAccountID` may be possibilities there. We could also add the same parameter name to the normal `Params` root struct to keep things somewhat consistent.

> AFICT we don't really document how to use the Go client with connected accounts, at least on a per-request basis with an Account param. (The closest thing I see is https://github.com/stripe/stripe-go#connect-flows.) Would it be worth adding that while we're here? This might look for example like a bit in the readme, inline docs for Account in Params (it's documented well in ListParams in this PR), and perhaps some clarifying comments re: the other Account/AccountID fields above. 

+1. I'm happy to add some explanatory text in the README.

mmcgrana:
`StripeAccount` as the field in Go seems very reasonable to me since the corresponding HTTP header is `Stripe-Account` (https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header). Would we change `Params` as well?

brandur:
> StripeAccount as the field in Go seems very reasonable to me since the corresponding HTTP header is Stripe-Account (https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header). Would we change Params as well?

Yes and yes! We'd probably keep the current `Account` in `Params` around for at least a while so as to avoid breaking too many people, but we'd add `StripeAccount` and prefer that. I've updated the pull to reflect this idea; would you mind having another look?

I've also added a small section in the `README` and I've added a `testing` package in case anyone has any feelings on either of those.

README.md
```diff
@@ -113,6 +113,19 @@ stripe := &client.API{}
 stripe.Init("access_token", nil)
 ```
 
+In order to make a request on behalf of a managed account, use the
```
mmcgrana: For this doc section, should we refactor to match the outline [here](https://stripe.com/docs/connect/authentication) of saying there are two ways to connect - with Stripe account IDs (suggested) and with access tokens, then provide details for each?

I'm not sure that it's still the case that a client is required, as mentioned in the older docs above?

I think the terminology we want here is "on behalf of a connected account" or "for a connected account", and not "of  a managed account", which means something a bit [more specific](https://stripe.com/docs/connect/managed-accounts).

mmcgrana:
Code lg! Left a few comments on the docs inline.

mmcgrana:
Are the build failures expected?

brandur:
> For this doc section, should we refactor to match the outline here of saying there are two ways to connect - with Stripe account IDs (suggested) and with access tokens, then provide details for each?

Sure! Reworded and reorganized the doc in 6f9220c.

> Are the build failures expected?

Unfortunately yes for a bit longer. It turns out that data deletion hadn't been making any progress for the last week. I deployed a fix yesterday, and we're now 850k objects into clean-up, so I'm going to leave it going for a bit longer since we've already waited this long.

brandur:
Alright, I've fleshed out the remaining clients in 44c7199. The test suites seem roughly as happy as they ever are.

brandur:
Alright, going for a merge here to get this one wrapped up.

Issue #198, opened 2016-02-23
Invoice preview, no subscription_plan


dstpierre:
On the documentation, there's a `subscription_plan` field we can pass to `invoice.GetNext`'s `&stripe.InvoiceParams`.

I'm not finding anything on the `invoice.go`'s `Invoice` struct. I guess `subscription` on the doc is `Sub`, but no way to pass plan.

Any help would be appreciated, getting a preview of a plan change is required at the moment to show to users the effect of prorating etc.

Thanks

brandur:
Added a fix in #200.

dstpierre:
Thanks, and yes, my PR was wrongly targeting the `Invoice` ;)

brandur:
@dstpierre Fix released as 11.4.0.

PR #199, opened 2016-02-23
added subscription_plan to Invoice

dstpierre:
this is related to #198


brandur:
@dstpierre Thanks for the patch! Unfortunately, I don't think this will do what you want it to :/ In #198 I think you talked about wanting to get a plan onto `InvoiceParams` so that you can use it during retrieval, but this struct represents the deserialized invoice resource, [which doesn't have a `subscription_plan` field](https://stripe.com/docs/api/go#invoice_object).

If you hold tight a bit longer, I think I can get you an alternative.

brandur:
Superseded by #200. Closing this out for now. Thanks again!

PR #200, opened 2016-02-23
Add missing invoice parameters for use with `GetNext`

### LINKED ISSUE #198: Invoice preview, no subscription_plan

brandur:
Adds a large set of missing parameters to `InvoiceParams` that are
designed to be used to [retrieve an upcoming invoice][docs]. Fixes #198.

A few design justifications here:

* I've continued using the prefix of `Sub` for "subscription" because
  this seems consistent with elsewhere in the library.
* The new parameter serializations should really be tested as a unit
  somewhere, but because we don't have a pattern for this, I've re-used
  the same approach as elsewhere in the suite by just hitting the API.
* I've made the single huge test in the invoices suite even more huge
  because as stated in its comments, there's a huge amount of
  bootstrapping that's needed in order to properly test any of this.
* I've only made minimal effort to test the invoice which the API
  responds with. This is because testing other fields according to other
  input parameters is non-trivial. It's not great, but it's better than
  what we have today.

And lastly: mandatory comment about how we need to fix this test suite.

/cc @remistr Would you mind reviewing this one? Thanks!

[docs]: https://stripe.com/docs/api/go#upcoming_invoice

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -14,6 +14,12 @@ type InvoiceParams struct {
 	Desc, Statement, Sub string
 	Fee                  uint64
 	Closed, Forgive      bool
+	SubPlan              string
+	SubProrate           bool
+	SubProrationDate     int64
+	SubProrationDateNow  bool
+	SubQuantity          uint64
+	SubTrialEnd          int64
 	TaxPercent           float64
 }
 
--- a/invoice.go
+++ b/invoice.go
@@ -15,7 +15,7 @@ type InvoiceParams struct {
 	Fee                  uint64
 	Closed, Forgive      bool
 	SubPlan              string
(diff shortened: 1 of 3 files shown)


remistr:
@brandur :ship:

brandur:
@remistr Thanks for the review!

PR #202, opened 2016-02-25
bindings for PII tokenization

gukjoon:
r? @brandur 

`POST /v1/tokens` needs to be able to accept `pii[personal_id_number]` to create PII tokens

Made the simplest change possible. Open to suggestions.

### DIFF
--- a/token/client.go
+++ b/token/client.go
@@ -11,6 +11,7 @@ import (
 const (
 	Card stripe.TokenType = "card"
 	Bank stripe.TokenType = "bank_account"
+	PII	 stripe.TokenType = "pii"
 )
 
 // Client is used to invoke /tokens APIs.
@@ -37,8 +38,10 @@ func (c Client) New(params *stripe.TokenParams) (*stripe.Token, error) {
 		params.Card.AppendDetails(body, true)
 	} else if params.Bank != nil {
 		params.Bank.AppendDetails(body)
+	} else if params.PII != nil {
+		params.PII.AppendDetails(body)
 	} else if len(params.Customer) == 0 {
-		err := errors.New("Invalid Token params: either Card or Bank need to be set")
+		err := errors.New("Invalid Token params: either Card, Bank, or PII need to be set")
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


token/client.go
```diff
@@ -11,6 +11,7 @@ import (
 const (
 	Card stripe.TokenType = "card"
 	Bank stripe.TokenType = "bank_account"
+	PII	 stripe.TokenType = "pii"
```
brandur: @gukjoon Can you run any Go files you touched through a `gofmt -w <file>`? It'll stripe little whitespace discrepancies like this out one out. Thanks!

pii.go
```diff
@@ -0,0 +1,15 @@
+package stripe
+
+import (
+  "net/url"
```
brandur: Also, the convention in Go is to only use tabs for leading whitespace on lines. `gofmt` again will fix this for you automatically.

token/client_test.go
```diff
@@ -73,3 +73,25 @@ func TestTokenGet(t *testing.T) {
...
+
+	if err != nil {
+		t.Error(err)
+	}
+
+	if target.Created == 0 {
+		t.Errorf("Created date is not set\n")
+	}
+
+	if target.Type != PII {
```
brandur: Great! Thanks for adding the test!

pii.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+type PIIParams struct {
+  Params
+  PersonalIdNumber string
```
brandur: If you look elsewhere in the bindings, it seems that convention is to upcase "ID" (e.g. `ID` under the `Charge` type). Can you change that?

brandur:
@gukjoon Looks generally good so modulo minor comments above. Thanks!

Two other questions:

1. Is "PII" ever going to be more of a general purpose concept that applies to something beyond just tokens? There's some possibility that if _not_, we may just want to keep it right in the `token.go` file.
2. As touched upon yesterday, if this personal ID is ever available in a response, we should probably add it to those structs as well. Does it ever come back with an API resource or is it purely a value that's injected only?

gukjoon:
Thanks for the quick review @brandur 

1. :+1: 
2. It does not. Infact we won't even return last 4 of personal_id_number.

Made some fixes. re-r? @brandur 
Thanks again!

brandur:
@gukjoon Cool, sounds good to me. Thanks!

brandur:
Released as 11.5.0.

PR #203, opened 2016-02-27
add support for managed account event fetch

aravind88:


brandur:
Thanks @aravind88! Can you just give me general confirmation that the tests in the event suite succeed locally with your change?

aravind88:
@brandur The event suite is failing for me with and without my change. https://gist.github.com/aravind88/9e053f46e1fe37db24fb

Am I missing some configuration?

aravind88:
oh. never mind. Figured out the issue. 
The first event with a type "charge.*" was a dispute event in my scenario, which did not have a `source` field. 
So I cleared all my test data, and created a dummy charge. Now there was a panic in the test because was there was no `card` field. I had to make the following change (358106e1be21f10801863385a6f1e995f2554dc5) in order to get the tests to pass. Let me know if it looks ok.

brandur:
@aravind88 Awesome! Yes, that test change sounds reasonable to me; thanks for taking a look at it.

Would you mind squashing your commits (i.e. `git rebase -i master`) and re-pushing? After that, this is good to come in.

aravind88:
@brandur done.

brandur:
@aravind88 Thanks again!

brandur:
Released as 12.0.0.

PR #206, opened 2016-03-14
Upgrade to API version 2016-03-07

### LINKED ISSUE #205: Modernize the coupled API version
The bundled version is pretty out-of-date. Let's follow through the changelog and pull this onto a newer version.

brandur:
Follows the [changelog][changelog] through and makes necessary changes to support the latest Stripe API version.

Important changes:

* `BankAccounts` on `Account` replaced by `ExternalAccounts`.
* `Currencies` on `Account` removed in favor of new `CountrySpec` resource.

There were quite a few other changes in the API too, but they were either already integrated or didn't require any modifications to the bindings.

Fixes #205. Supersedes #169.

/cc @mmcgrana Sorry to be a pain, but would you mind taking a look at this one? Thanks! (And BTW the tests are still broken due to a _still_ in progress test data deletion :/)

[changelog]: https://stripe.com/docs/upgrades

### DIFF
--- a/account.go
+++ b/account.go
@@ -73,23 +73,24 @@ type Account struct {
 	ChargesEnabled bool   `json:"charges_enabled"`
 	Country        string `json:"country"`
 	// Currencies is the list of supported currencies.
-	Currencies           []string `json:"currencies_supported"`
-	DefaultCurrency      string   `json:"default_currency"`
-	DetailsSubmitted     bool     `json:"details_submitted"`
-	TransfersEnabled     bool     `json:"transfers_enabled"`
-	Name                 string   `json:"display_name"`
-	Email                string   `json:"email"`
-	Statement            string   `json:"statement_descriptor"`
-	Timezone             string   `json:"timezone"`
-	BusinessName         string   `json:"business_name"`
-	BusinessPrimaryColor string   `json:"business_primary_color"`
-	BusinessUrl          string   `json:"business_url"`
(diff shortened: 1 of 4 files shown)


mmcgrana:
Thanks Brandur.

This mostly looks good.

I have a question around the design of `ExternalAccountList`. The approach there of breaking out `BankAccountValues` and `CardValues` at the top level seems different than the perhaps related case of [`SourceList`](https://godoc.org/github.com/stripe/stripe-go#SourceList), where `Values` is of `*PaymentSource`s but then those have fields for the different types.

Was choosing a different approach here explicit and if so what was the thinking?

I don't have a strong opinion one way or another, just would like to understand the context here.

brandur:
> Was choosing a different approach here explicit and if so what was the thinking?

Hey @mmcgrana: so wow, we'd started discussing this in #169 and brainstorming ideas for implementation around external accounts, but somehow throughout that discussion failed to go and look up prior art (namely with payment sources). So when I started continuing that work I decided to go and reinvent the wheel thinking that we were breaking new ground. It's a really good thing that you looked this up so as to prevent that from happening.

I've now re-implemented external accounts to look like payment sources. Mind taking one more look here? Thanks!

mmcgrana:
LG thanks @brandur!

brandur:
Thanks Mark!

brandur:
Released as version 13.0.0.

PR #207, opened 2016-03-15
Add account.reject method.

gregsabo:
To match the new [Account.reject API](https://stripe.com/docs/api/go#reject_account).

r? @stripe/api-libraries

### DIFF
--- a/account.go
+++ b/account.go
@@ -187,7 +187,7 @@ type TOSAcceptanceParams struct {
 }
 
 // RejectionParams is the structure for the Reject function.
-type AccountRejectionParams struct {
+type AccountRejectParams struct {
 	Reason string `json:"reason"`
 }
 
--- a/account.go
+++ b/account.go
@@ -186,6 +186,11 @@ type TOSAcceptanceParams struct {
 	UserAgent string `json:"user_agent"`
 }
 
+// RejectionParams is the structure for the Reject function.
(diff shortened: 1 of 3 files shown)


account.go
```diff
@@ -186,6 +186,11 @@ type TOSAcceptanceParams struct {
 	UserAgent string `json:"user_agent"`
 }
 
+// RejectionParams is the structure for the Reject function.
+type AccountRejectionParams struct {
```
brandur: Another nit, but what do you think about renaming this to `AccountRejectParams`? I think it may be more consistent with the way other params are named (e.g. we have `AccountListParams` instead of `AccountListingParams`)

account.go
```diff
@@ -186,6 +186,11 @@ type TOSAcceptanceParams struct {
 	UserAgent string `json:"user_agent"`
 }
 
+// RejectionParams is the structure for the Reject function.
```
brandur: I think the name of the struct in this comment may have fallen out of date.

brandur:
Nice! Thanks a lot for getting that test in there. I left a couple minor comments, but this looks good to me.

gregsabo:
@brandur OK, mind taking another look?

brandur:
Great. Thanks!

brandur:
Released as 13.1.0.

PR #208, opened 2016-03-18
Add SourceTx and SourceTransfer

tylerb:


transfer.go
```diff
@@ -38,24 +38,25 @@ type TransferListParams struct {
...
+	Status            TransferStatus    `json:"status"`
+	Type              TransferType      `json:"type"`
+	Tx                *Transaction      `json:"balance_transaction"`
+	Meta              map[string]string `json:"metadata"`
+	Bank              *BankAccount      `json:"bank_account"`
+	Card              *Card             `json:"card"`
+	Recipient         *Recipient        `json:"recipient"`
+	Statement         string            `json:"statement_descriptor"`
+	Reversals         *ReversalList     `json:"reversals"`
+	SourceTransaction *Transaction      `json:"source_transaction"`
```
brandur: I'm not crazy about the existing naming, but what do you about renaming this to `SourceTx` for consistency with the `Tx` field?

brandur:
Thanks! Left a minor comment, but looks good to me.

transfer.go
tylerb: That sounds fine to me. I'll make that change now.

tylerb:
Changed to `SourceTx` - let me know if anything else needs to be changed.

brandur:
Thanks!

brandur:
Released as version 13.2.0.

PR #216, opened 2016-03-29
Add metadata field to Card resource

### LINKED ISSUE #214: card structure is missing metadata
the [api](https://stripe.com/docs/api#card_object) says card object has metadata field(the card endpoint does return the metadata) while the golang lib's [card struct](https://github.com/stripe/stripe-go/blob/master/card.go) does not have.

brandur:
Adds a missing metadata field (Meta) to the Card resource.

Fixes #214.


brandur:
Released as v13.4.0.

PR #220, opened 2016-04-06
added additional fields for card info

xiaolekuang:
Sync the card info. structure in sdk with API document.
The returned example from API is like following:

> {
> id: xxxxxxxxxx
> object: "card"
> address_city: null
> address_country: null
> address_line1: null
> address_line1_check: null
> address_line2: null
> address_state: null
> address_zip: null
> address_zip_check: null
> brand: "xxxx"
> country: "xx"
> customer: xxxxxxxxx
> cvc_check: "unchecked"
> **description: "Debit MasterCard"**
> dynamic_last4: null
> exp_month: 3
> exp_year: 2021
> fingerprint: "xxxxxxxxxxxxxxxxxxx"
> funding: "debit"
> **iin: "111111"**
> **issuer: "CIMB BANK BERHAD"**
> last4: "2222"
> metadata:
> name: null
> tokenization_method: null
> use: "debit"
> }

PR #221, opened 2016-04-06
Add internal fields for card

brandur:
Adds a set of fields for specialized use that may be available depending on
gating, but which usually are not.

Replaces #220.


brandur:
Released as 13.7.0.

PR #222, opened 2016-04-12
	charge: add outcome fields

mduvall:
## What's in here

This PR adds `charge.Outcome` which is an embedded struct that contains the outcome pertaining to a particular charge. The motivation is to freshen the Go bindings in preparation for the imminent `outcome` field in the charge responses from the API.

r? @brandur

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -73,7 +73,7 @@ type Charge struct {
 	Invoice        *Invoice          `json:"invoice"`
 	Live           bool              `json:"livemode"`
 	Meta           map[string]string `json:"metadata"`
-	Outcome        *Outcome          `json:"outcome"`
+	Outcome        *ChargeOutcome    `json:"outcome"`
 	Paid           bool              `json:"paid"`
 	Refunded       bool              `json:"refunded"`
 	Refunds        *RefundList       `json:"refunds"`
@@ -94,7 +94,7 @@ type FraudDetails struct {
 
 // Outcome is the charge's outcome that details whether a payment
 // was accepted and why.
-type Outcome struct {
+type ChargeOutcome struct {
 	NetworkStatus string `json:"network_status"`
 	Reason        string `json:"reason"`
 	SellerMessage string `json:"seller_message"`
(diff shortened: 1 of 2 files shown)


mduvall:
Also happy to bump the minor on `clientversion`!

charge.go
```diff
@@ -91,6 +92,15 @@ type FraudDetails struct {
 	StripeReport FraudReport `json:"stripe_report"`
 }
 
+// Outcome is the charge's outcome that details whether a payment
+// was accepted and why.
+type Outcome struct {
```
brandur: What do you think about naming this `ChargeOutcome` instead? That feels a little less ambiguous to me.

brandur:
@mduvall Thanks! Except for the very minor comment above, this looks good to me.

> Also happy to bump the minor on clientversion!

Nah, don't worry, we'll do it! No point introducing unnecessary possibility of conflict.

Can you also just confirm that the new tests pass locally? (Unfortunately our CI process doesn't really run most of the test suite.)

charge.go
brandur: (And to be clear, I'm suggesting the change because `Outcome` lives in the parent namespace which increases the likelihood of some eventually collision.)

mduvall: Ah yeah agreed, will change! I'll keep the named field on the `Charge` struct as `Outcome` since that's (I think) not as ambiguous.

brandur: > I'll keep the named field on the Charge struct as Outcome since that's (I think) not as ambiguous.

Yep, that works for me! (And is identical to what I did in Java.)

mduvall:
@brandur Awesome, mind taking another look? Here is a test run, the results are same for before/after my changes (I'm not sure what's up with some of the failures), the pertinent tests pass in `charge`: https://gist.github.com/mduvall/9b6cc1ee134d345da4eb5733c35e6047.

brandur:
Looks good to me! Thanks for making those changes.

> Here is a test run, the results are same for before/after my changes (I'm not sure what's up with some of the failures), the pertinent tests pass in charge: https://gist.github.com/mduvall/9b6cc1ee134d345da4eb5733c35e6047.

Yeah :/ Tests that are not run in CI are destined by entropy to not be useful as tests for very long, and that's unfortunately what's happened here. The important part is that the new additions are running correctly.

brandur:
Released as 13.8.0.

mduvall:
@brandur Woot thank you! :100:

PR #224, opened 2016-04-14
Add support for go sub functions not passing in customer

jxu577:
This PR adds go support for v1/subs - methods on subscriptions that no longer require a customerID

Changes:
DelSub now only takes in a subscriptionID and not a customerID anymore
All URLs are now directed towards v1/subs rather than v1/cus
All functions have been changed to support if params passed in are `nil` - previously params had to include customerID but now it is possible they are `nil` values
List now takes in optional customer and plan parameters 
Added test to ensure both old and new methods still work

### DIFF
--- a/discount/client.go
+++ b/discount/client.go
@@ -28,13 +28,13 @@ func (c Client) Del(customerID string) (*stripe.Discount, error) {
 
 // DelSub removes a discount from a customer's subscription.
 // For more details see https://stripe.com/docs/api#delete_subscription_discount.
-func DelSub(customerID, subscriptionID string) (*stripe.Discount, error) {
-	return getC().DelSub(customerID, subscriptionID)
+func DelSub(subscriptionID string) (*stripe.Discount, error) {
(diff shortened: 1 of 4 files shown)


mmcgrana:
It looks like this may still need review? (I understand it's also waiting for the API changes to land, but we should be able to get it reviewed in parallel).

@jxu577 since this is a non-trivial (indeed, breaking) change, could you please add a description to the PR for the reviewers' benefit? It may also be helpful to provide some context on why the methods on `Client` needed to change so much (though perhaps that's obvious to anyone more familiar with the codebase).

cc @stripe/api-libraries for visibility.

kyleconroy:
@jxu577 Can you rebase this onto master? Looks like it's gotten out of sync.

kyleconroy:
We'll want to make sure that we update the change log to include migration instructions for people updating the library. A simple one-liner is all we'll need.

One last thing, the subscription struct is missing the `Created` field.

PR #226, opened 2016-04-18
Introduce `TaxPercentZero` field for invoice and subscription parameters

### LINKED ISSUE #225: Can't set tax back to zero after setting it to non-zero value.
Seeing as how TaxPercent isn't set in the request if it's zero, there's no way to actually set it to zero. This is a problem if your tax rate has been set previously and you want to change it.

brandur:
Introduces a new field for `TaxPercentZero` for invoices in
subscriptions so that an already non-zero value can be zeroed out (the
current check prevents a zero from ever making it to the server because
it's not added to the serialized form of parameters).

This might be a little contentious, but I didn't add any tests for the
new fields. Sub's `QuantityZero` field does have a test, but it's sixty
lines long and tests a single conditional branch, which doesn't seem
sustainable to me from the standpoints of either code quality or the
amount of time that it takes to run the suite (given that all of these
are still doing live API calls). Definitely open for discussion if
anyone has strong opinions on the subject.

Fixes #225.

/cc @tetrakai Would you mind taking a review pass on this one? Thanks!

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -21,6 +21,7 @@ type InvoiceParams struct {
 	SubQuantity          uint64
 	SubTrialEnd          int64
 	TaxPercent           float64
+	TaxPercentZero       bool
 }
 
 // InvoiceListParams is the set of parameters that can be used when listing invoices.
--- a/invoice/client.go
+++ b/invoice/client.go
@@ -52,6 +52,8 @@ func (c Client) New(params *stripe.InvoiceParams) (*stripe.Invoice, error) {
 
 	if params.TaxPercent > 0 {
 		body.Add("tax_percent", strconv.FormatFloat(params.TaxPercent, 'f', 2, 64))
+	} else if params.TaxPercentZero {
+		body.Add("tax_percent", "0")
 	}
 
 	invoice := &stripe.Invoice{}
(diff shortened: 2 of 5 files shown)


tetrakai:
Ugh, I dislike the reasons for needing this field, but agree that it's the best solution for now.

It doesn't look like we have tests for the `Invoice` endpoint at all, so I'm fine with leaving this particular field untested there.

Could we add a test for `TaxPercentZero` to the existing `TestSubscriptionZeroQuantity` method though? It would be an easy field to leave out in a refactor (especially believing that that endpoint was already being tested) so I'd prefer to hit that code path at least once, and should only take a few lines.

Otherwise, LGTM.

brandur:
> Could we add a test for TaxPercentZero to the existing TestSubscriptionZeroQuantity method though? It would be an easy field to leave out in a refactor (especially believing that that endpoint was already being tested) so I'd prefer to hit that code path at least once, and should only take a few lines.

Yes! Done in 0fb2686.

Thanks for taking a look!

ssoroka:
Looks like this got the New() methods but missed the Update methods. TaxPercentZero should be there as well.

brandur:
> Looks like this got the New() methods but missed the Update methods. TaxPercentZero should be there as well.

Ouch, that's my mistake. Added a fix in #229.

PR #228, opened 2016-04-19
Expose `stripe.PlanList`

### LINKED ISSUE #227: Expose planList (and possibly other similar types)
I recently tried to do a mock of the `GET /plans` endpoint. However, when I tried to inject my own fixture data, I ran into trouble because the `planList` type that is being used to handle that response is private.

Is there any reason that [planList](https://github.com/stripe/stripe-go/blob/9492eb6d10c700ce855dcf94dfeee6c80798bd3c/plan/client.go#L131-L134) is declared this way and is also private? Can this just be made into a public `PlanList` type? (similar to `SubList` and others)

As it stands now, I need to rely on an external fixture file that I read and pass to `json.Unmarshal`, but I'd prefer to use a raw value directly in my tests to keep them more self-contained.

brandur:
Moves the struct used to represent a list of plans into an exported type
to ease mocking of API endpoints.

Fixes #227.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -38,3 +38,9 @@ type Plan struct {
 	Statement     string            `json:"statement_descriptor"`
 	Deleted       bool              `json:"deleted"`
 }
+
+// PlanList is a list of plans as returned from a list endpoint.
+type PlanList struct {
+	ListMeta
+	Values []*Plan `json:"data"`
+}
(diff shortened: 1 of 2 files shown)


brandur:
I'm going to leave this open for an hour or two just in case there's any random commentary, but the change is so benign that I'm not going to seek out more official review.

PR #229, opened 2016-04-20
Add TaxPercentZero logic to update methods

brandur:
Follows up #226 by adding `TaxPercentZero` handling to the update
methods for invoices and subscriptions.


brandur:
Released as 13.10.1.

PR #231, opened 2016-04-22
New external accounts can be marked as default from a token

remistr:
Fixed the ability to create a new external account on a managed account from a token and mark it as default at the same time.

r? @brandur

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	stripe "github.com/stripe/stripe-go"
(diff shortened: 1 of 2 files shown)


brandur:
Thanks Remi! :)

It looks like you may need to run gofmt to get the build happy, but +1 after that's done.

remistr:
@brandur nice catch, sorry for missing it! I just pushed an updated version

brandur:
Released as 13.11.1.

PR #237, opened 2016-04-28
Add the Currency field to Card objects

remistr:
r? @brandur


brandur:
Thanks Remi!

FWIW, I don't necessarily think that we _have_ to add new tests for every new field we tack onto any of these objects. It's somewhat useful for a little extra certainty, but it might have a downside in that it'll make these tests pretty hard to untangle if we ever want to do a more complete refactor. Not a big deal though :)

brandur:
Released as 13.12.0.

Issue #238, opened 2016-04-29
order.New should return error if params.Shipping is nil


ghophp:
We are creating a service that uses the stripe-go API to communicate with the backend of stripe.. we expect that the method `order.New()` from the `order` package return an error if something is wrong with the `stripe.OrderParams` object... It would be great if we could introduce a small check if the  `params.Shipping` is nil before accessing it.

https://github.com/stripe/stripe-go/blob/master/order/client.go#L64

I did a PR that prevent this exception. This change look valid since you don't need to force people to initialise an empty `Shipping` object when initializing the `OrderParams` object. In our case, we receive the order as a JSON and just Unmarshal with the `OrderParams` object.

Thank you very much in advance for providing a nice library in golang, and for the great job with the whole technology at Stripe.

PR #239, opened 2016-04-29
prevent nil exception if the params.Shipping is nil

ghophp:
here is the changes to fix the issue #238


kyleconroy:
Thanks @ghophp!

PR #242, opened 2016-05-06
Added json tags for camel cased order params

makarski:
We are using your library from inside of our backend proxy which holds a private `apiKey`. The clients are calling our service with the json serialized payloads equal to the ones expected by the original stripe api. We have though an issue while creating and deserializing shipping params. See example: 

Lets assume we have a call like this:
```
curl -k -X POST --dump-header - 'proxy-service-host/v1/orders/' -H 'Content-Type: application/json' --data '{"currency":"usd","metadata":{"user_device_type":"iphone"},"items":[{"parent":"H83027_380","type":"sku"}],"email":"hujnn@gghh.de","shipping":{"name":"zhnn","address":{"line1":"jjbmkgbb","city":"hvbmnh","line2":"","country":"US","postal_code":"52690","state":"AZ"}}}'
```
`postal_code` is snake cased.

Our go script does the following:
```go
orderParams := &stripe.OrderParams{
        Address: &stripe.AddressParams{},
    },
}

if err := json.Unmarshal(orderPayload, orderParams); err != nil {
    return nil, err
}

order, err := order.New(orderParams)
if err != nil {
    return nil, err
}
// etc
```

Provided that PostalCode property does not have a json tag https://github.com/stripe/stripe-go/blob/master/order.go#L38, we can't deserialize it properly.

Please, let us know, if you think this PR makes sense. Thank you for the sdk!

### DIFF
--- a/order.go
+++ b/order.go
@@ -35,14 +35,14 @@ type AddressParams struct {
 	Line2      string
 	City       string
 	State      string
-	PostalCode string
+	PostalCode string `json:"postal_code"`
 	Country    string
 }
 
 type OrderUpdateParams struct {
 	Params
 	Coupon                 string
-	SelectedShippingMethod string
+	SelectedShippingMethod string `json:"selected_shipping_method"`
 	Status                 OrderStatus
 }
 
@@ -100,7 +100,7 @@ type OrderPayParams struct {
 	Params
 	Source         *SourceParams
 	Customer       string
-	ApplicationFee int64
+	ApplicationFee int64 `json:"application_fee"`
 	Email          string
 }
 


brandur:
Hi @makarski, thanks for the PR!

I think it's a little too unusual to assign a JSON attribute to these three select fields when no others have them, and they'll never get used as part of any normal usage of the SDK. As an alternative, would you mind just copy and pasting these three parameter structs into your own source code and adding the JSON tags there? Thanks!

makarski:
hi @brandur. Thanks for the feedback. I think your sdk is generally used from the proxy backends, which implies those backends need to be able to accept the requests a stripe api would accept, add auth credentials and repass the requests further on. If you think it makes sense to add json tags to the other properties, I would add them.

brandur:
> hi @brandur. Thanks for the feedback. I think your sdk is generally used from the proxy backends, which implies those backends need to be able to accept the requests a stripe api would accept, add auth credentials and repass the requests further on. If you think it makes sense to add json tags to the other properties, I would add them.

Hi @makarski,

I definitely see where you're coming from, but I think that instead of calling those backends proxies, I might just call them APIs. Although there's little doubt that some merchants are probably doing little more than proxying requests upstream to us, I suspect that many have modeled custom endpoints that add a rich layer of other logic relating to their individual domains. As such, I think it might be better if we stayed more general and recommended that SDK users continue to define their own parameter structs which are then converted into those in the Stripe SDK.

Keep in mind also that by re-using these parameter structs in your endpoints, you're introducing an implicit dependence on how we've model these data types. It's possible that if we change something down the line (i.e. remove a field) and you upgrade that you could run into results that you didn't necessarily expect. Most of these endpoints are pretty stable, so you're probably fine, but it could probably be considered good practice to explicitly model the inputs to all your own API endpoints.

Let me know if anyone else disagrees or has anymore thoughts on this one though!

brandur:
I'm going to close this one for now, but let me know if you have any other thoughts/comments. Thanks!

PR #246, opened 2016-05-11
Add missing `Reversed` and `AmountReversed` fields to `Transfer`

### LINKED ISSUE #244: Extend Transfer type with Reversed and AmountReversed
Hello,

Is it possible to extend **Transfer** type with **Reversed** and **AmountReversed**  (those are listed in https://stripe.com/docs/api#retrieve_transfer)?

type Transfer struct {
        ....
	Reversed       bool              `json:"reversed"`
	AmountReversed int64             `json:"amount_reversed"`
}

see https://github.com/stripe/stripe-go/blob/master/transfer.go

Also is there a way to test full/partial reversals in test Stripe mode? Looks like transfers to bank accounts are immediately "paid" and I'm getting "Transfers to non-Stripe accounts can currently only be reversed while they are pending" on attempt to make reversal.

brandur:
Fixes #244.

### DIFF
--- a/transfer.go
+++ b/transfer.go
@@ -43,26 +43,28 @@ type TransferListParams struct {
 // Transfer is the resource representing a Stripe transfer.
 // For more details see https://stripe.com/docs/api#transfers.
 type Transfer struct {
-	ID         string             `json:"id"`
-	Live       bool               `json:"livemode"`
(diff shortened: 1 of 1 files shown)


brandur:
Released as 14.2.0.

PR #247, opened 2016-05-12
Makes routing_number only be sent in POST if set

zachwick:
When tokenizing a bank account, the routing number is only required
for US bank accounts. If using an IBAN for `account_number`, then
this field is optional. Prior to this commit, `routing_number` was
always being sent, even as a an empty string which was causing the
following error to be sent back from the Stripe API:

{
  "type":"invalid_request_error",
  "message":"Invalid bic",
  "param":"bank_account",
  "request_id":"req_XXXYYYZZZ",
  "status":400
}

Relevant docs URL: https://stripe.com/docs/api/go#create_bank_account_token-bank_account-routing_number
Signed-off-by: Zach Wick <zwick@stripe.com>

### DIFF
--- a/bankaccount/client.go
+++ b/bankaccount/client.go
@@ -43,8 +43,11 @@ func (c Client) New(params *stripe.BankAccountParams) (*stripe.BankAccount, erro
 		body.Add("external_account[object]", "bank_account")
 		body.Add("external_account[country]", params.Country)
 		body.Add("external_account[account_number]", params.Account)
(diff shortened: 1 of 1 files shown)


brandur:
Nice work! Thanks for taking the time to get this fixed via PR.

Also, sorry about the delay here. I've been on a bit of vacation, and am only now getting caught up on email :$

brandur:
Released as 14.2.2.

PR #248, opened 2016-05-12
add Skus to the client API struct

sjsnider:
perhaps this was done on purpose for a good reason, but Sku isn't included on the client API struct and I wanted to be able to use it and would like to think I'm not the first one.  This just adds it real quick

I was not getting all the tests to pass even though I'm quite certain I was setting my test key correctly (and also noticed at least in the client directory that the key is hardcoded anyway).  Hoping this PR is still viable.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -27,6 +27,7 @@ import (
 	"github.com/stripe/stripe-go/recipient"
 	"github.com/stripe/stripe-go/refund"
 	"github.com/stripe/stripe-go/reversal"
+	"github.com/stripe/stripe-go/sku"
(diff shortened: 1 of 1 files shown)


brandur:
@sjsnider Thanks for the PR! I think this was just an oversight :/

> I was not getting all the tests to pass even though I'm quite certain I was setting my test key correctly (and also noticed at least in the client directory that the key is hardcoded anyway). Hoping this PR is still viable.

Unfortunately, a large part of the test suite doesn't currently run in CI and is therefore very prone to failure as it falls out of date for one reason or another. This looks good as is!

brandur:
Released as version 14.2.1.

PR #249, opened 2016-05-19
Fix to Pull #247

alessandropietrobelli:
Related to pull #247 , also when need to create token for bank account, routing_number is optional according to documentation [https://stripe.com/docs/api#create_bank_account_token](url)


brandur:
Thanks!

PR #250, opened 2016-05-19
Add order returns, delivery estimates

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz 

Bring our Go libraries up to speed with some of our recent API improvements.

### DIFF
--- a/order.go
+++ b/order.go
@@ -60,10 +60,27 @@ type Shipping struct {
 }
 
 type ShippingMethod struct {
-	ID          string   `json:"id"`
-	Amount      int64    `json:"amount"`
-	Currency    Currency `json:"currency"`
-	Description string   `json:"description"`
+	ID               string            `json:"id"`
+	Amount           int64             `json:"amount"`
+	Currency         Currency          `json:"currency"`
+	Description      string            `json:"description"`
+	DeliveryEstimate *DeliveryEstimate `json:"delivery_estimate"`
+}
+
+type EstimateType string
+
+const (
+	Exact EstimateType = "exact"
+	Range EstimateType = "range"
+)
+
+type DeliveryEstimate struct {
+	Type EstimateType `json:"type"`
+	// If Type == Range
+	Earliest string `json:"earliest"`
+	Latest   string `json:"latest"`
+	// If Type == Exact
+	Date string `json:"date"`
 }
 
 type Order struct {
--- a/order.go
+++ b/order.go
@@ -46,6 +46,13 @@ type OrderUpdateParams struct {
 	Status                 OrderStatus
 }
 
+// OrderReturnParams is the set of parameters that can be used when returning
+// orders. For more details, see: https://stripe.com/docs/api#return_order.
+type OrderReturnParams struct {
+	Params
+	Items []*OrderItemParams
+}
+
 type Shipping struct {
 	Name    string  `json:"name"`
 	Address Address `json:"address"`
--- a/order/client.go
+++ b/order/client.go
@@ -289,7 +289,7 @@ func (i *Iter) Order() *stripe.Order {
 // Update updates an order's properties.
 // For more details see https://stripe.com/docs/api#update_order.
 func Return(id string, params *stripe.OrderReturnParams) (*stripe.OrderReturn, error) {
-	return getC().Update(id, params)
+	return getC().Return(id, params)
 }
 
 // Update updates an order's properties.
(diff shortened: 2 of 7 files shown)


order/client.go
```diff
@@ -285,6 +286,51 @@ func (i *Iter) Order() *stripe.Order {
 	return i.Current().(*stripe.Order)
 }
 
+// Update updates an order's properties.
```
brandur: Can you fix the Godoc comment here? Thanks!

order/client.go
```diff
@@ -285,6 +286,51 @@ func (i *Iter) Order() *stripe.Order {
...
+// For more details see https://stripe.com/docs/api#update_order.
+func Return(id string, params *stripe.OrderReturnParams) (*stripe.OrderReturn, error) {
+	return getC().Return(id, params)
+}
+
+// Update updates an order's properties.
+// For more details see https://stripe.com/docs/api#update_order.
+func (c Client) Return(id string, params *stripe.OrderReturnParams) (*stripe.OrderReturn, error) {
```
brandur: And same here.

order.go
```diff
@@ -46,17 +46,41 @@ type OrderUpdateParams struct {
 	Status                 OrderStatus
 }
 
+// OrderReturnParams is the set of parameters that can be used when returning
+// orders. For more details, see: https://stripe.com/docs/api#return_order.
+type OrderReturnParams struct {
+	Params
+	Items []*OrderItemParams
```
brandur: You'd know better than I, but is there a fundamental link between an order return item and an order item that's guaranteed to stay true in the future? You'd know better than I, but there's some possibility that you may just want to define a new `OrderReturnItemParams` struct if there's some chance that they'll deviate down the line.

brandur:
Left a couple minor comments, but looks generally good to me. Thanks!

order/client.go
rasmusrygaard: Duh, good catch! Thanks, updated!

order.go
rasmusrygaard: They're referring to the same resource in the API so I think it's extremely unlikely that they'll ever diverge. Basically returns specify a subset of the items that the order were created with. The only difference is that you can get away with specifying a subset of the `OrderItemParams` on returns because we can infer the rest, but they fundamentally refer to the same data. Internally we also use the same definition for these so I think it's peaceful leaving as is (:

order/client.go
```diff
@@ -285,6 +286,51 @@ func (i *Iter) Order() *stripe.Order {
...
+// For more details see https://stripe.com/docs/api#return_order.
+func (c Client) Return(id string, params *stripe.OrderReturnParams) (*stripe.OrderReturn, error) {
+	var body *url.Values
+	var commonParams *stripe.Params
+
+	if params != nil {
+		body = &url.Values{}
+
+		if len(params.Items) > 0 {
+			for _, item := range params.Items {
```
brandur: I also just wanted to point out that it seems like any multi-item form arrays may be generally broken in the Go bindings right now (including here). See #251 for more details. I'll try to get this one addressed soon.

order.go
brandur: Cool! Thanks for the clarification.

brandur:
Released in version 14.3.0.

order/client.go
rasmusrygaard: Wow, that's quite bad. Happy to pair on getting this working again, seems like we would need `body` to be some `url.Values`-like abstraction that respects ordering

brandur: > Wow, that's quite bad. Happy to pair on getting this working again, seems like we would need body to be some url.Values-like abstraction that respects ordering

+1. I'm hoping to take a stab at implementation in a couple days. I suspect that it'll largely to be a rote find and replace after designing a new abstraction, but will definitely take your pairing offer in mind ;) Thanks!

tdestro:
Programming in Go: Creating Applications for the 21st Century 
"4.3.4 Key-Ordered Map Iteration"

https://books.google.com/books?id=HW-5SZ1HKusC&pg=PA170&lpg=PA170&dq=%224.3.4+Key-Ordered+Map+Iteration%22&source=bl&ots=Fq903bY-oo&sig=9mfzSgTtYNlPfRoEvNnTuKQDs3o&hl=en&sa=X&ved=0ahUKEwiN_ZKdnenMAhXJWT4KHaZvDv4Q6AEIHTAA#v=onepage&q=%224.3.4%20Key-Ordered%20Map%20Iteration%22&f=false

order/client.go
tdestro: Programming in Go: Creating Applications for the 21st Century 
"4.3.4 Key-Ordered Map Iteration"

https://books.google.com/books?id=HW-5SZ1HKusC&pg=PA170&lpg=PA170&dq=%224.3.4+Key-Ordered+Map+Iteration%22&source=bl&ots=Fq903bY-oo&sig=9mfzSgTtYNlPfRoEvNnTuKQDs3o&hl=en&sa=X&ved=0ahUKEwiN_ZKdnenMAhXJWT4KHaZvDv4Q6AEIHTAA#v=onepage&q=%224.3.4%20Key-Ordered%20Map%20Iteration%22&f=false

brandur: > You could use something like this: 

@tdestro Yep, looks promising! Thanks!

tdestro: It wasn't. I looked into it further, it's key ordered (ie: alphabetical) , not insertion ordered, which is what you need. check out https://github.com/cevaris/ordered_map

PR #252, opened 2016-05-20
Add a StableForm type that can support ordered duplicates

### LINKED ISSUE #251: Multiple items in order params doesn't seem to work.
I think I may have found a bug. This seems to only work for me if I have one item. Also I noticed the tests in stripe-go/order/client_test.go in  func TestOrder(t *testing.T) don't test for orders with multiple items. I get this in the stripe panel:
screen shot 2016-05-20 at 10 32 55 am

```

	order, err := sc.Orders.New(&stripe.OrderParams{
		Currency: "usd",
		Items: []*stripe.OrderItemParams{
			{
				Type: "sku",
				Parent: "sku_8KuzQIEccDprsR",
			},
			{
				Type: "sku",
				Parent: "sku_8KbgQAaeg7C2S9",
			},
		},
			Name: "Jenny Rosen",
			Address: &stripe.AddressParams{
				Line1:      "1234 Main Street",
				City:       "Anytown",
				Country:    "US",
				PostalCode: "123456",
			},
			Phone: "6504244242",
		},
		Email: "jenny@ros.en",
		Params: stripe.Params{
			Meta: map[string]string{
				"foo": "bar",
			},
		},
	})

```

brandur:
As described in #251, using `url.Values` to contain form values is
problematic for the Stripe API because it depends on "Rack style" form
encoding that requires duplicated parameters to occur in a certain order
for arrays to be correctly encoded.

This patch adds a new type called `StableForm` that encodes parameters
in the order that they were added so that arrays can be properly
supported. It also adds a `Form` interface that's shared between
`StableForm` and `url.Values` so that we can demonstrate this concept
for now without updating every use of `url.Values` everywhere. For
consistency, before the final merge we should probably move to
`StableForm` everywhere and remove the interface (which will produce a
messy diff).

From #251, here's an incorrectly encoded form containing a multi-item array:

```
items[][amount]=0&items[][amount]=0&items[][description]=&items[][description]=&items[][parent]=sku_8KuzQIEccDprsR&items[][parent]=sku_8KbgQAaeg7C2S9&items[][type]=sku&items[][type]=sku
```

And here's the same form encoded with this patch:

```
items[][description]=&items[][type]=sku&items[][amount]=0&items[][parent]=sku_8KuzQIEccDprsR&items[][description]=&items[][type]=sku&items[][amount]=0&items[][parent]=sku_8KbgQAaeg7C2S9
```

/cc @rasmusrygaard Would you mind letting me know what you think about this
approach? Thanks!

/cc @stripe/api-libraries 

Fixes #251.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -9,7 +9,6 @@ import (
 	"io/ioutil"
 	"log"
 	"net/http"
-	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -38,7 +37,7 @@ const TotalBackends = 2
 // Backend is an interface for making calls against a Stripe service.
 // This interface exists to enable mocking for during testing if needed.
 type Backend interface {
-	Call(method, path, key string, body *url.Values, params *Params, v interface{}) error
+	Call(method, path, key string, body Form, params *Params, v interface{}) error
 	CallMultipart(method, path, key, boundary string, body io.Reader, params *Params, v interface{}) error
 }
 
@@ -144,9 +143,9 @@ func SetBackend(backend SupportedBackend, b Backend) {
 }
 
 // Call is the Backend.Call implementation for invoking Stripe APIs.
-func (s BackendConfiguration) Call(method, path, key string, form *url.Values, params *Params, v interface{}) error {
+func (s BackendConfiguration) Call(method, path, key string, form Form, params *Params, v interface{}) error {
 	var body io.Reader
-	if form != nil && len(*form) > 0 {
+	if form != nil {
 		data := form.Encode()
 		if strings.ToUpper(method) == "GET" {
 			path += "?" + data
--- a/order/client.go
+++ b/order/client.go
@@ -24,11 +24,11 @@ func New(params *stripe.OrderParams) (*stripe.Order, error) {
 // New POSTs a new order.
 // For more details see https://stripe.com/docs/api#create_order.
 func (c Client) New(params *stripe.OrderParams) (*stripe.Order, error) {
-	var body *url.Values
+	var body *stripe.StableForm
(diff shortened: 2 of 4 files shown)


stripe.go
```diff
@@ -144,9 +143,9 @@ func SetBackend(backend SupportedBackend, b Backend) {
 }
 
 // Call is the Backend.Call implementation for invoking Stripe APIs.
-func (s BackendConfiguration) Call(method, path, key string, form *url.Values, params *Params, v interface{}) error {
+func (s BackendConfiguration) Call(method, path, key string, form Form, params *Params, v interface{}) error {
 	var body io.Reader
-	if form != nil && len(*form) > 0 {
```
brandur: I removed the `len` check here because it's problematic for our interface above, and both `Encode` implementations will encode safely to an empty string which is not harmful in an HTTP request regardless of whether it's in a query string or body.

rasmusrygaard:
Looks great, thanks for the quick fix!

tdestro:
Just a note about a discussion I had tonight with @slovichon. Maps (ie url.values) are reference types, so they should not have pointers unless there is some reason to do so.

brandur:
> I think your simple solution is the way to go. Just a note about a discussion I had tonight with @slovichon. Maps (ie url.Values) are reference types (already), so they should not have pointers unless there is some reason to do so.

Yep. Great point!

I've opened #253 with the complete changes necessary to refactor the entire parameter system. It's the same as this pull except with `StableForm` renamed to the more general `RequestValues` and the interface removed because it's now applied across the entire codebase. I opened a new pull because the diff is huge and this pull will act as a simple way to examine the new design without having to wade through that.

PR #253, opened 2016-05-23
Added DeclineCode property to stripe.Error struct

hexdigest:
This field is present in java,ruby,python client libraries but not in golang client.

### DIFF
--- a/error.go
+++ b/error.go
@@ -37,6 +37,7 @@ type Error struct {
 	Type           ErrorType `json:"type"`
 	Msg            string    `json:"message"`
 	Code           ErrorCode `json:"code,omitempty"`
+	DeclineCode    string    `json:"decline_code,omitempty"`
 	Param          string    `json:"param,omitempty"`
 	RequestID      string    `json:"request_id,omitempty"`
 	HTTPStatusCode int       `json:"status,omitempty"`
--- a/stripe.go
+++ b/stripe.go
@@ -299,6 +299,10 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
 			err.ChargeID = charge.(string)
 		}
 
+		if declineCode, ok := root["decline_code"]; ok {
+			err.DeclineCode = declineCode.(string)
+		}
+
 		if LogLevel > 0 {
 			Logger.Printf("Error encountered from Stripe: %v\n", err)
 		}


brandur:
Thanks for the pull!

The one thing to consider here is that `decline_code` is not a general error field, and instead is specific to charge errors. If we add it into the root error type like this, you'll never be able to access it reliably, and instead have to implement a logical ladder so that if it's empty you fall back to other fields (i.e. in case it's _not_ a charge error that comes back). IMO, the right thing to do here is implement a similar approach to other language SDKs and introduce an error hierarchy with `decline_code` added on something like `ChargeError` struct.

Let me take a look at trying to implement this a little later today.

hexdigest:
Hi, I thought about it but this will break backward compatibility and therefore you'll need to keep several versions of Golang documentation here: https://stripe.com/docs/api

However I'm fine with both solutions. I guess we need to have special error types for every Error.Type:
"invalid_request_error", "api_error", "card_error"

hexdigest:
I think I realized how to keep backward compatibility. Please wait for the next PR.

brandur:
@hexdigest Cool! Will do so.

I [also wrote up a branch here that does something along the lines that I suggested above](https://github.com/stripe/stripe-go/compare/master...brandur-errors-overhaul). Usage would be something like this:

``` go
	if cardErr, ok := err.(*stripe.CardError); ok {
		fmt.Printf("Card declined with code: %v\n", cardErr.DeclineCode)
	} else if stripeErr, ok := err.(*stripe.StripeError); ok {
		fmt.Printf("Generic Stripe error: %v\n", stripeErr.Msg)
	} else {
		fmt.Printf("Unknown error: " + err.Error())
	}
```

The one problem that I've encountered so far is that you sort of need to know in advance what error types to expect because Go isn't able to cast a type to a parent type (i.e. a `CardError` cannot be casted to `StripeError` even if it includes `StripeError`). I'm not crazy about that, so unless we can find a way around it, we may have to leave this as is :/

brandur:
Actually, thinking about this a bit more, it might be a reasonable idea to copy the model of `net.OpError` in that a single error type is always returned, but it includes an accessor that allows a more specific error to be extracted.

So getting a decline code would look something like `cardErr, ok := err.(*stripe.Error).Err.(*stripe.CardError).DeclineCode`.

This would also be an improvement in that we could maintain backward compatibility. Thoughts?

hexdigest:
In your example only (*stripe.CardError) type assertion is safe so in reality we can get a panic if Err is not *stripe.CardError but stripe.APIError for example. So this approach needs two nested if conditions in order to make it safe:

```go
if stripeErr, ok := err.(*stripe.Error); ok {
  if cardErr, ok := stripeErr.Err.(*stripe.CardError);ok {
     ...
  }
}
```
Which looks a bit awkward to me. 

If we don't care about backward compatibility so much than it's better to follow go way and replace stripe.Error with interface and also create stripe.CardError, stripe.APIError and stripe.InvalidRequestError interfaces. This will make type switches like this possible:
```
switch e := err.(type) {
case nil:
  fmt.Println("We're fine")
case stripe.CardError:
  fmt.Println("CardError:", e.Code())
case stripe.InvalidRequestError:
  fmt.Println("Don't try to repeat this request please")
case stripe.Error:
  fmt.Printf("Some unknown stripe error type: %v", e)
case error:
  fmt.Printf("Technical problem: %v", e)
}
```

hexdigest:
I gave up to make clear backward compatible solution. I think this is the only way to do it.
Consider creating new stripe-go-v2 repo. I'll help to rewrite error handling in idiomatic way and fix other issues I noticed.

brandur:
> In your example only (*stripe.CardError) type assertion is safe so in reality we can get a panic if Err is not *stripe.CardError but stripe.APIError for example. So this approach needs two nested if conditions in order to make it safe:

Right. I guess I'd point out a few things here though:

1. If you're not overly concerned with a particular error type, you're safe to stay only one level deep and still get something useful.
2. Getting back something other than a `stripe.Error` is possible, but unlikely, and probably won't occur as part of day-to-day operation. It may be relatively safe (i.e. for the 99.99% case) to just allow the panic in this situation.
2. The first cast to `stripe.Error` is already in place today so not much is changing beyond adding an additional optional detail granularity.

> If we don't care about backward compatibility so much than it's better to follow go way and replace stripe.Error with interface and also create stripe.CardError, stripe.APIError and stripe.InvalidRequestError interfaces. This will make type switches like this possible:

Oh right! The only thing that I don't like about this approach though is that you have to handle _every_ possible custom error type :/ Due to the way that the Go type system works (i.e. in that you can't cast a "subtype" like `stripe.CardError` to a parent like `stripe.Error`), there is no way to say something like "do something special if it's a `stripe.CardError`, but otherwise handle it as a generic `stripe.Error`". This makes any kind of error handling beyond the basic of `fmt.Println(err.Error())` unduly burdensome.

> I gave up to make clear backward compatible solution. I think this is the only way to do it.
Consider creating new stripe-go-v2 repo. I'll help to rewrite error handling in idiomatic way and fix other issues I noticed.

Although not technically kosher in the Go world, we're okay with introducing the occasional breaking change into the mainline repository with the assumption that almost everyone building a real app is using some kind of Go package manager these days.

That said, I'm leaning toward the backwards compatible change I outlined above here, and am currently drawing up a patch for it.

We would definitely still appreciate contributions and/or reports of any other problems that you've noticed along the way though!

brandur:
I've implemented the alternative proposed above in #254.

hexdigest:
> i.e. in that you can't cast a "subtype" like stripe.CardError to a parent like stripe.Error

if these are Structs - yes. But stripe.Error and stripe.CardError could be different Interfaces and stripe.CardError can also implement stripe.Error. So if you need just general stripe.Error and you don't care about subtypes you can just cast to stripe.CardError interface and this is actually go way to do it.
Consider this:

```go
  package stripe
  ...

  type Error Interface {
     error
     HTTPStatusCode() int
  }

  type CardError {
     Error
     Code() ErrorCode
     DeclineCode() string
  }

  type CardErrorStruct struct {
     Code ErrorCode
     DeclineCode string
     ChargeID string
  }

  //implementation on CardErrorStruct
```

Using this approach you can cast an instance of stripe.CardError to stripe.Error so there's no need to have to handle every possible custom type. In my previous comment you could do something like this easily:

```go
switch e := err.(type) {
case nil:
  fmt.Println("We're fine")
case stripe.Error:
  //this will work for CardError, InvalidRequestError etc
  fmt.Printf("Some stripe error: %v", e)
default:
  fmt.Printf("Technical problem: %v", e)
}
```

brandur:
> if these are Structs - yes. But stripe.Error and stripe.CardError could be different Interfaces and stripe.CardError can also implement stripe.Error. So if you need just general stripe.Error and you don't care about subtypes you can just cast to stripe.CardError interface and this is actually go way to do it.

Cool! Aside from the minor concern that it seems a little unusual to use interfaces for error handling like this (I can't think of any places in the standard library where this is done off hand), this approach certainly makes sense.

I think that both these routes are pretty good options, however, because one is a breaking change and one isn't, I'd tend to prefer the latter here.

hexdigest:
> Cool! Aside from the minor concern that it seems a little unusual to use interfaces for error handling like this 

The fact that CardErrorStruct implements stripe.Error and stripe.CardError interfaces doesn't stop you from casting stripe.Error or regular error interface to concrete underlying type that implements it:
```go
switch e := err.(type) {
case nil:
  fmt.Println("We're fine")
case *stripe.CardErrorStruct:
  fmt.Printf("Some stripe error: %v", e.Code)
case stripe.Error:
  //this will work for other types of stripe errors like InvalidRequestError or APIError
  fmt.Printf("Some stripe error: %v", e)
default:
  fmt.Printf("Technical problem: %v", e)
}
```

However I'm fine with your current solution, thanks!

brandur:
> The fact that CardErrorStruct implements stripe.Error and stripe.CardError interfaces doesn't stop you from casting stripe.Error or regular error interface to concrete underlying type that implements it:

Cool! Yes, totally right.

Anyway, thanks a lot for the discussion here! For now I'm going to close this down in favor of #254.

PR #254, opened 2016-05-23
Overhaul the errors subsystem

brandur:
Adds specific types for all specific errors that can be returned by the Stripe API. These types are then grafted onto the common `stripe.Error` under the new field `Err` so that a user can access type-specific information about the error.

Usage looks a little like:

``` go
_, err := // Go library call

if err != nil {
    stripeErr := err.(*stripe.Error)

	if cardErr, ok := stripeErr.Err.(*stripe.CardError); ok {
        fmt.Printf("Card was declined with code: %v\n", cardErr.DeclineCode)
    } else {
        fmt.Printf("Other Stripe error occurred: %v\n", stripeErr.Error())
    }

    return nil, err
}
```

The change is designed to overhaul the error system in a way that enables access to important features, conforms to standard library convention (see [`net.OpError`](https://golang.org/pkg/net/#OpError)), and all the while still maintains backwards compatibility.

Presented as an alternative to #253.

### DIFF
--- a/error.go
+++ b/error.go
@@ -28,11 +28,19 @@ const (
 	Missing       ErrorCode = "missing"
 	ProcessingErr ErrorCode = "processing_error"
 	RateLimit     ErrorCode = "rate_limit"
+
+	// These additional types are written purely for backward compatibility
+	// (the originals were given quite unsuitable names) and should be
+	// considered deprecated. Remove them on the next major version revision.
+
+	APIErr         ErrorType = ErrorTypeAPI
+	CardErr        ErrorType = ErrorTypeCard
+	InvalidRequest ErrorType = ErrorTypeInvalidRequest
 )
 
 // StripeError is the response returned when a call is unsuccessful.
 // For more details see  https://stripe.com/docs/api#errors.
-type StripeError struct {
+type Error struct {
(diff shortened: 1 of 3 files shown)


error.go
```diff
@@ -41,10 +48,82 @@ type Error struct {
...
-// Error serializes the Error object and prints the JSON string.
+// Error serializes the error object to JSON and returns it as a string.
 func (e *Error) Error() string {
 	ret, _ := json.Marshal(e)
 	return string(ret)
 }
+
+// APIConnectionError is a failure to connect to the Stripe API.
+type APIConnectionError struct {
+	StripeErr *Error
```
kyleconroy: Why not embed StripeErr here instead of a field?

brandur: @kyleconroy Thanks for taking a look!

So what I was trying to do was just make this a reference back up to the "parent" error so that we could use its context to produce an error message from `Error()` rather than explicitly embed a struct and make its fields available here.

Now that I look at this again, I'm going to try to further reduce the interface surface area by making this field unexported as well.

brandur: Changed in 73e44ee.

brandur:
Pulling this in!

PR #255, opened 2016-05-24
Add a RequestValues type that can support ordered duplicates

### LINKED ISSUE #251: Multiple items in order params doesn't seem to work.

brandur:
As described in #251, using url.Values to contain form values is
problematic for the Stripe API because it depends on "Rack style" form
encoding that requires duplicated parameters to occur in a certain order
for arrays to be correctly encoded.

This pull continues the solution introduced in #252 by applying the new type
across every file. As a result, the commit is a bit of a mess so I'm going to
rely on the design #252 having been reviewed already.

Replaces #252. Fixes #251.

### DIFF
--- a/account.go
+++ b/account.go
@@ -3,7 +3,6 @@ package stripe
 import (
 	"encoding/json"
 	"fmt"
-	"net/url"
 	"strconv"
 )
 
@@ -274,7 +273,7 @@ type AccountRejectParams struct {
 }
 
 // AppendDetails adds the legal entity to the query string.
-func (l *LegalEntity) AppendDetails(values *url.Values) {
+func (l *LegalEntity) AppendDetails(values *RequestValues) {
 	values.Add("legal_entity[type]", string(l.Type))
 
 	if len(l.BusinessName) > 0 {
@@ -397,7 +396,7 @@ func (l *LegalEntity) AppendDetails(values *url.Values) {
 }
 
 // AppendDetails adds the transfer schedule to the query string.
-func (t *TransferScheduleParams) AppendDetails(values *url.Values) {
+func (t *TransferScheduleParams) AppendDetails(values *RequestValues) {
 	if t.Delay > 0 {
 		values.Add("transfer_schedule[delay_days]", strconv.FormatUint(t.Delay, 10))
 	} else if t.MinimumDelay {
@@ -413,7 +412,7 @@ func (t *TransferScheduleParams) AppendDetails(values *url.Values) {
 }
 
 // AppendDetails adds the terms of service acceptance to the query string.
-func (t *TOSAcceptanceParams) AppendDetails(values *url.Values) {
+func (t *TOSAcceptanceParams) AppendDetails(values *RequestValues) {
 	if t.Date > 0 {
 		values.Add("tos_acceptance[date]", strconv.FormatInt(t.Date, 10))
 	}
--- a/account/client.go
+++ b/account/client.go
@@ -2,7 +2,6 @@
 package account
 
 import (
-	"net/url"
 	"strconv"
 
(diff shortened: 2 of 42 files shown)


brandur:
I also forgot to mention it, but I'm going to do a major version bump after pulling this in. Although the _recommended_ public interface should stay identical, there are some exported functions like `GetIter` and `AppendDetails` that have had their signature change. Hopefully no one is using these, but you never know.

dominicbarnes:
Unfortunately, this change has broken my mock tests. This is because the order is now important, and that requires me to know the internals of each API call I mock to get the params added in the right order.

The previous `url.Values` implementation always sorted it's output alphabetically by key, so that made it easy for me to be ignorant of the internals. I'm not exactly sure what to propose in place of this, as I understand the change itself to be useful/important.

brandur:
@dominicbarnes Hmm, sorry to hear about the trouble :/

Could you add a helper that converts the new `RequestValues` back to `url.Values`? I wouldn't 100% recommend this, because as described in #251, `url.Values` is not actually capable of accurately representing a form-encoded request when it comes to more complex Rack data structures, but it sounds like it would work in your environment.

dominicbarnes:
I don't _think_ I can since the internals of `RequestValues` are all private. Maybe you can make `RequestValues.values` and `formValues` public instead of private? Or add a public method that converts it to `url.Values`?

brandur:
@dominicbarnes It looks like you saw it already, but I added a helper to get back to `url.Values` in #260.

However, thinking about it a little more, I think it might be slightly cleaner if you set up a mock HTTP server and injected its address to Stripe clients as a `BackendConfiguration`, then, when receiving a request, just extract a `url.Values` via the incoming `http.Request`. Does that sound like a huge pain?

And on a similar note, we should probably be documenting some testing guidance somewhere in this project ...

dominicbarnes:
@brandur setting up a whole mock server does sound a bit of a pain atm. My setup now is actually quite nice, since I just use a custom Backend which is based on the [testify mock library](https://godoc.org/github.com/stretchr/testify/mock). I just inspect the incoming method, url, body, etc to determine which mock I want to have returned. Each and every test is completely self-contained, and there's no real setup/teardown logic to deal with. (unlike what I would expect from a mock server atm)

brandur:
@dominicbarnes Cool! Alright, I'm fine with merging this for now.

Your testing strategy sounds good; it might even be better if we could modify the backend interfaces slightly so that parameters structs (e.g. `AccountParams`) could be plumbed all the way through instead of the lower level `RequestValues`. That way you could compare against named fields directly instead of trying to unpack the raw request. Unfortunately though, that would entail more interface changes, so maybe a project for another day.

PR #257, opened 2016-05-25
Add support for Customer business_vat_id

guillaumehenriot:
Added support for the business_vat_id property in customers: https://stripe.com/docs/api#create_customer-business_vat_id

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -17,6 +17,7 @@ type CustomerParams struct {
 	TrialEnd      int64
 	DefaultSource string
 	Shipping      *CustomerShippingDetails
+	BusinessVatID string
 }
 
 // SetSource adds valid sources to a CustomerParams object,
@@ -52,6 +53,7 @@ type Customer struct {
 	Subs          *SubList                 `json:"subscriptions"`
 	Deleted       bool                     `json:"deleted"`
 	Shipping      *CustomerShippingDetails `json:"shipping"`
+	BusinessVatID string                   `json:"business_vat_id"`
 }
 
 // CustomerList is a list of customers as retrieved from a list endpoint.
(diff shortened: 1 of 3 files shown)


guillaumehenriot:
r? @brandur 
forgot that line...

customer.go
```diff
@@ -17,6 +17,7 @@ type CustomerParams struct {
 	TrialEnd      int64
 	DefaultSource string
 	Shipping      *CustomerShippingDetails
+	BusinessVatId string
```
brandur: Would you mind renaming this to `BusinessVatID` just to be more in-line with other identifiers in the project? [For example](https://github.com/stripe/stripe-go/blob/adffbfe3d540c9cb8fd8d7962df88b6c86714583/discount/client.go#L31).

It maybe should be `BusinessVATID`, but that looks like we're already using `BusinessVatID` elsewhere in the project.

brandur:
Left a minor comment above, but looks good. Thanks!

guillaumehenriot:
Good catch! I changed it.

brandur:
@guillaumehenriot Thanks! But actually, would you mind going `BusinessVatID`? Even if it's less strictly correct, I noticed [that this capitalization is already in use elsewhere in the project](https://github.com/stripe/stripe-go/blob/adffbfe3d540c9cb8fd8d7962df88b6c86714583/account.go#L212), so we should stick to it.

guillaumehenriot:
@brandur I'm really sorry, I read your comment too quickly and went for the wrong version. Changed it to the good one now!

brandur:
Awesome!

LGTM.

brandur:
Released as 15.1.0.

PR #258, opened 2016-05-26
Normalize indentation in README

brandur:
Normalizes all code intentation in the README to use tabs which is Go's
convention. Gets rids of two-space indentation and tab indentation
that's been expanded to four spaces.


mmcgrana:
Good call, thanks!

PR #260, opened 2016-06-03
Add `ToValues` to convert `RequestValues` to `url.Values`

brandur:
Adds a helper on `RequestValues` to convert it to a `url.Values` map.
This allows objects to be compared in an order-invariant way, which
might be useful in testing.

See #255.


dominicbarnes:
Awesome! Thanks :)

brandur:
Released as 15.2.0.

PR #262, opened 2016-06-15
Introduce `BalanceZero` field for customer parameters

### LINKED ISSUE #261: Can't set account balance to zero.
Because there's no ZeroBalance parameter, and because zero checks are assumed to be unset, the following code makes it so that you cannot set a zero balance.
```
		if params.Balance != 0 {
			body.Add("account_balance", strconv.FormatInt(params.Balance, 10))
		}
```
I think we're going to keep seeing this issue come up again and again. Using the default initialization values to imply null makes it so that you can't set parameters to their default values. 

Possible solutions:
- define things like ZeroBalance all over the place.
- expose behavior like the underlying `body.Add("account_balance", ...)`
- adopt nullable values or use pointers to values so you can tell the difference between null and 0.

guillaumehenriot:
r? @brandur 

Introduces a new field `BalanceZero` for customers so that an already non-zero value can be zeroed out. It's modeled on the `TaxPercentZero` behavior.

I added a test to the existing `TestCustomerUpdate` subpackage. Let me know if you prefer it to be separated.

Fixes #261

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -9,6 +9,7 @@ import (
 type CustomerParams struct {
 	Params
 	Balance       int64
+	BalanceZero   bool
 	Token, Coupon string
 	Source        *SourceParams
 	Desc, Email   string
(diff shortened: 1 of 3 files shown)


brandur:
Thanks for taking a look at this @guillaumehenriot!

LGTM.

brandur:
Released as 15.3.0.

PR #263, opened 2016-06-20
Add Tokenization Method to Card object

domwong:
It's shown in the API docs but not in the struct. https://stripe.com/docs/api#card_object-tokenization_method


brandur:
Thanks @domwong!

Looks good to me!

brandur:
Released as 15.4.0.

PR #265, opened 2016-06-23
Add `tax_percent` paramter to customer create

### LINKED ISSUE #264: CustomerParams.TaxPercent does not exist even though it is available in the API
Would like to be able to set TaxPercent when creating the initial customer, but this does not seem to be possible via the go lib.

https://stripe.com/docs/api#create_customer lists tax_percent as a param that can be passed to create_customer

brandur:
It seems that this field is missing for the customer create method. This
change reflects the pattern that we already have for `tax_percent` in
`sub.go`.

Fixes #264.

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -8,17 +8,19 @@ import (
 // For more details see https://stripe.com/docs/api#create_customer and https://stripe.com/docs/api#update_customer.
 type CustomerParams struct {
 	Params
-	Balance       int64
-	BalanceZero   bool
-	Token, Coupon string
-	Source        *SourceParams
-	Desc, Email   string
-	Plan          string
-	Quantity      uint64
-	TrialEnd      int64
-	DefaultSource string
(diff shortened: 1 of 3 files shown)


guillaumehenriot:
Would it be worth to add a check to see if the subscription exists and the TaxPercent has been set? @brandur 

Otherwise LGTM.

brandur:
> Would it be worth to add a check to see if the subscription exists and the TaxPercent has been set? @brandur 

Good call! I'm already a little concerned at how many API calls this test is making though, so I think I'm going to leave that additional check out for now.

Unfortunately the whole approach to testing in this project is pretty off. Instead of testing that we're adding new parameters to the HTTP request as expected, we're actually testing that the API's behavior matches how it was when the tests were written, and often at great expense. Hopefully we can get all of this cleaned up the next time we do a refresh on this code.

Thanks!

PR #266, opened 2016-06-24
Support modifying SKU/product attributes.

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz 

Support updating SKU and product attributes after creation.

### DIFF
--- a/sku/client.go
+++ b/sku/client.go
@@ -125,6 +125,10 @@ func (c Client) Update(id string, params *stripe.SKUParams) (*stripe.SKU, error)
 			body.Add("image", params.Image)
 		}
 
+		for k, v := range params.Attrs {
+			body.Add(fmt.Sprintf("attributes[%v]", k), v)
+		}
+
 		inventory := params.Inventory
 
 		if len(inventory.Type) > 0 {
--- a/product/client.go
+++ b/product/client.go
@@ -125,6 +125,17 @@ func (c Client) Update(id string, params *stripe.ProductParams) (*stripe.Product
 			body.Add("url", params.URL)
 		}
 
+		// Passing empty attributes should unset the attributes.
+		if params.Attrs != nil {
+			if len(params.Attrs) > 0 {
+				for _, v := range params.Attrs {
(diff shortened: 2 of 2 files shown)


sku/client.go
```diff
@@ -125,6 +125,10 @@ func (c Client) Update(id string, params *stripe.SKUParams) (*stripe.SKU, error)
 			body.Add("image", params.Image)
 		}
 
+		for k, v := range params.Attrs {
+			body.Add(fmt.Sprintf("attributes[%v]", k), v)
```
brandur: I worry quite a bit about this sort of interface:

1. If you don't understand that the update will act as a replacement, then you may end up overwriting your old attributes because you didn't explicitly copy them over.
2. If you do understand that the update will act as a replacement, you won't succeed in overwriting your old attributes if your new set is smaller than your old set (i.e. because attributes at the upper indexes will still exist). Is it even possible to blank this field out like you can with a product?

rasmusrygaard: I'm not sure I 100% follow here, but maybe I'm not giving you enough context around how `attributes` works.

- The keyspace for the `attributes` hash is fixed and determined by the associated product
- Attributes can be partially updated. If I have a SKU with `{attr1: a, attr2: b}`, I can send `{attr1: c}` and the resulting product attributes would be `{attr1: c, attr2: b}`. 
- Attributes cannot be unset. It's not valid to send `{attr1: ""}`. This makes more sense when you think about what attributes actually mean. A shirt can't just lose its size or color, and it's not really well-defined what you'd do with a product that has attributes `size, color` and a SKU with only attribute `color`.

Hope that helps clarify the semantics (: My gut feeling is that this is ~sane, but we should definitely strive to have something simple and obvious here

brandur: Ah! Sorry this was my mistake. I was assuming that `Attrs` here was a slice for some reason, but now that I go and look at it, this is obviously not the case. Thanks for the clarification!

brandur:
LGTM.

brandur:
Released as 15.6.0.

PR #268, opened 2016-06-30
Fix SKU attributes, send product.

rasmusrygaard:
r? @stripe/api-libraries 
cc @jimdanz 

A couple of minor fixes for SKUs:
- We don't have a `Name` attribute on SKUs
- We support updating the product of a SKU, and `sku.Update` should support that as well


brandur:
LGTM.

Thanks Rasmus!

brandur:
Released as 16.0.0.

PR #269, opened 2016-07-01
3ds charges [don't merge yet]

jonashuckestein:


brandur:
Cool. Thanks for taking a crack at this! Let me know when you want me to take a look for review.

danielchatfield:
@brandur I've updated this PR. I'm not 100% happy with this as the 3DS stuff doesn't neatly fit in like other resources do.

There are two quite separate 3DS objects - the one you get when you create a 3DS object from a card and the one that is nested inside a card. I could separate this into two structs but I think this is even less desirable than the approach I've taken.

threedsecure/client.go
```diff
@@ -0,0 +1,39 @@
...
+
+func New(params *stripe.ThreeDSecureParams) (*stripe.ThreeDSecure, error) {
+	return getC().New(params)
+}
+
+func (c Client) New(params *stripe.ThreeDSecureParams) (*stripe.ThreeDSecure, error) {
+	body := &stripe.RequestValues{}
+	body.Add("amount", strconv.FormatUint(params.Amount, 10))
+	body.Add("currency", string(params.Currency))
+	body.Add("card", params.Card)
```
brandur: Mind just tweaking this slightly to come above `currency`?

threedsecure/client.go
```diff
@@ -0,0 +1,39 @@
...
+	body.Add("return_url", params.ReturnURL)
+
+	params.AppendTo(body)
```
brandur: And mind just putting this below `customer`?

brandur:
@danielchatfield Thanks for picking this up! I left a few minor nitpicks, but overall this looks great to me.

> There are two quite separate 3DS objects - the one you get when you create a 3DS object from a card and the one that is nested inside a card. I could separate this into two structs but I think this is even less desirable than the approach I've taken.

Yeah, +1 that this could be slightly problematic in that it's misleading if you're using autocomplete or something, but I think that what you've done is probably for the best given that it's strong convention throughout the rest of the resources. I've mentioned the inconsistency to the team that implemented the feature in case they want to consider changing it.

danielchatfield:
@brandur I've addressed those comments :)

brandur:
@danielchatfield Awesome! Thanks again for taking a look at this.

LGTM.

brandur:
Released as 16.7.0.

PR #271, opened 2016-07-07
Bump API version.

rasmusrygaard:
r? @stripe/api-libraries 

We're a few versions behind, this PR bumps us to the most recent API version.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -20,7 +20,7 @@ const (
 )
 
 // apiversion is the currently supported API version
-const apiversion = "2016-03-07"
+const apiversion = "2016-07-06"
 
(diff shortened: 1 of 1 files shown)


brandur:
Re-assigning this one to me :$ I think we need to get that GitHub list updated, haha.

Cool! Looks good. [Taking a look at recent API versions](https://stripe.com/docs/upgrades), it looks like this should be safe as far as the Go bindings are concerned. We should make sure to implement filtering for cancelled subscriptions though.

Looks good to me (as long as we can get the tests passing here).

brandur:
LGTM.

brandur:
Released as 16.2.0.

Issue #36, opened 2014-10-29
No way to set zero (0) Quantity subscription from Golang client


donjaime:
This page https://stripe.com/docs/subscriptions claims that: 
"If quantity is set to zero, the subscription will act like a free plan, creating invoices but not charging customers."

This in fact works when you call the API with something like CURL.

However, when omitted, the stripe API doc says "quantity optional, default is 1".

In Go zero (0) is the default value for an uninitialized number. As such stripe.SubParams.Quantity being set to zero means "un-initialized". Which gets elided from the request that is made. This means you cannot set a zero quantity subscription from the golang client. Trying to do so makes a subscription of quantity 1 since it falls back to the default.

Which is unfortunate because we have need for such a thing :).

PR #272, opened 2016-07-11
Using SubParams.QuantityZero in subscription update action

mattevans:
Using the Golang client there isn't a way to update a subscription `Quantity` to `0`. This should hopefully resolve that!

It seems awhile back this functionality was added for creating subscriptions, but never implemented for updating them. Reference: #36


brandur:
@mattevans Great! Thanks for taking the trouble of putting this in!

LGTM.

brandur:
Released as 16.2.1.

PR #275, opened 2016-07-19
adds NoClosed parameter to open an invoice

### LINKED ISSUE #273: Can't re-open an invoice
`&stripe.InvoiceParams{Closed: false}` is [ignored](https://github.com/stripe/stripe-go/blob/master/invoice/client.go#L134), confused with the zero-value.

There's no good way to allow it given the current API -- if it recognized `false` then the lack of any explicitly set `Closed` value would also trigger re-opening.

Maybe something like this:

```go
// package stripe
type InvoiceClosedState int

const (
    Unset InvoiceClosedState = iota
    Closed
    Open
)

// package stripe
type InvoiceParams struct {
    // ...
    Closed InvoiceClosedState
    // ...
}
```

That would implicitly treat zero value of the `Closed` type as `Unset`.

guillaumehenriot:
r? @brandur 
fixes #273 

This commit adds a `NoClosed` parameter to the `InvoiceParams` struct to fix the fact that passing a `false` value to `Closed` would be ignored. You just need to pass `true` to this parameter to reopen the invoice.

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -14,6 +14,7 @@ type InvoiceParams struct {
 	Desc, Statement, Sub string
 	Fee                  uint64
 	Closed, Forgive      bool
+	NoClosed             bool
 	SubPlan              string
 	SubNoProrate         bool
 	SubProrationDate     int64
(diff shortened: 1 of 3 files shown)


brandur:
Thanks @guillaumehenriot!

LGTM.

brandur:
Released as 16.3.0.

PR #282, opened 2016-07-27
Strip duplicate country conditional

brandur:
We actually have two of these in `writeAccountParams`. Take one out.

cc @guillaumehenriot


guillaumehenriot:
LGTM

brandur:
@guillaumehenriot Thank-you!

PR #286, opened 2016-08-03
allow update of debit_negative_balance and managed flags

aravind88:
There doesn't seem to be a way to update debit_negative_balance for Managed accounts. 

Since we only want to update it if its set, we need to convert the fields to a pointer. 

r @brandur

### DIFF
--- a/account.go
+++ b/account.go
@@ -59,7 +59,7 @@ type AccountParams struct {
 	ExternalAccount           *AccountExternalAccountParams
 	LegalEntity               *LegalEntity
 	TransferSchedule          *TransferScheduleParams
-	Managed, DebitNegativeBal bool
+	Managed, DebitNegativeBal *bool
 	TOSAcceptance             *TOSAcceptanceParams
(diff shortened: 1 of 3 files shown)


aravind88:
cc @brandur

account.go
```diff
@@ -59,7 +59,7 @@ type AccountParams struct {
 	ExternalAccount           *AccountExternalAccountParams
 	LegalEntity               *LegalEntity
 	TransferSchedule          *TransferScheduleParams
-	Managed, DebitNegativeBal bool
+	Managed, DebitNegativeBal *bool
```
remistr: Changing the type of the variable is a breaking change and not something I'd recommend. You'd need to follow the solutions we've used in other PRs such as https://github.com/stripe/stripe-go/pull/275 and use another parameter for this. It's not pretty but ensures that it doesn't break someone's integration

aravind88: ah. got it.

Issue #287, opened 2016-08-05
Meta data on CardParams is not included when creating charges


telecoda:
We have noticed when we create a charge with metadata in the Source, CardParams, it is not sent to the Stripe API.

	charge, err := charge.New(&stripe.ChargeParams{
		Amount:   uint64(1234),
		Currency: stripe.Currency("GBP"),
		Source: &stripe.SourceParams{
			Card: &stripe.CardParams{
				Name:   "testing metadata",
				Number: "4242424242424242",
				Month:  "12",
				Year:   "35",
				Params: stripe.Params{
					Meta: map[string]string{
						"paymentMethodId": "test-method-id",
					},
				},
			},
		},
		Desc: "test desc",
		Params: stripe.Params{
			Meta: map[string]string{
				"customerId": "test-cust-id",
				"orderId":    "test-order-id",
			},
		},
	})

The paymentMethodId will not appear in the meta part of the request.

telecoda:
I have done some further testing with the following results:

Request:

    {
      "amount": "1234",
      "currency": "GBP",
      "card": {
        "object": "card",
        "number": "4242424242424242",
        "exp_month": "12",
        "exp_year": "35",
        "name": "testing metadata",
        "metadata": {
          "paymentMethodId": "test-method-id"
        }
      },
      "description": "test desc",
      "capture": "true",
      "metadata": {
        "orderId": "test-order-id",
        "customerId": "test-cust-id"
      }
    }

Response:

    {
      "id": "ch_xxx",
      "object": "charge",
      "amount": 1234,
      "amount_refunded": 0,
      "application_fee": null,
      "balance_transaction": "txn_xxx",
      "captured": true,
      "created": 1470833205,
      "currency": "gbp",
      "customer": null,
      "description": "test desc",
      "destination": null,
      "dispute": null,
      "failure_code": null,
      "failure_message": null,
      "fraud_details": {},
      "invoice": null,
      "livemode": false,
      "metadata": {
        "orderId": "test-order-id",
        "customerId": "test-cust-id"
      },
      "order": null,
      "paid": true,
      "receipt_email": null,
      "receipt_number": null,
      "refunded": false,
      "refunds": {
        "object": "list",
        "data": [],
        "has_more": false,
        "total_count": 0,
        "url": "/v1/charges/ch_xxx/refunds"
      },
      "shipping": null,
      "source": {
        "id": "card_xxx",
        "object": "card",
        "address_city": null,
        "address_country": null,
[...]

PR #288, opened 2016-08-05
Include CardParams metadata

telecoda:
This PR, fixes https://github.com/stripe/stripe-go/issues/287

To include CardParam meta data in the API request.

### DIFF
--- a/card.go
+++ b/card.go
@@ -190,6 +190,9 @@ func (c *CardParams) AppendDetails(values *RequestValues, creating bool) {
 			values.Add("address_country", c.Country)
 		}
 	}
+
+	c.AppendTo(values)
+
 }
 
 // Display human readable representation of a Card.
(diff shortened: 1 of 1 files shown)


brandur:
I have a slight preference here for using the the "prefix style" Rack encoding just because that's how we do it elsewhere. i.e. `card[metadata][key]=value`

Regardless though, given that [all metadata encoding currently happens in one place](https://github.com/stripe/stripe-go/blob/224faec412d280d27b46853512a30a228e8df2ac/params.go#L216,L218), we'd probably want to see if we could find a more generalizable solution here.

telecoda:
@brandur I've been testing by locally changing the stripe-go library to pass the Source metadata on a Charge request and it is passed to the Stripe API.  However when I get a response back from Stripe the source metadata on charge creation is blank.  Charge level metadata works ok.

So my first question is, does the Stripe API support metadata being passed on Charge request Source objects or not?  I've noticed the metadata works ok on create card requests.

If it doesn't I'll close this PR.

Issue #193, opened 2016-02-12
The bindings should handle the `destination` parameter on the Transfer object


remistr:
The `destination` [parameter](https://stripe.com/docs/api#transfer_object-destination) isn't handled currently. It can be an Account, a BankAccount or a Card right now so it will need some parent class.

Looks similar to the discussion in https://github.com/stripe/stripe-go/pull/169

brandur:
Doh, thanks @remistr! I lost track of this one somehow, but let's take a look at it soon.

MDrollette:
Same story for `source` on `Transaction`'s. There should be a way to get the object if it's expanded.

MDrollette:
This can be closed as it's supported as of 17.0.0 by #294

brandur:
@MDrollette +1. Thanks again!

PR #290, opened 2016-08-05
include destination on transfer

MDrollette:
A temporary solution for #193 to at least have access to the destination as a string ID.


MDrollette:
I'm going to re-submit this with support for the full hydrated objects.

PR #291, opened 2016-08-09
add missing Invoice.ReceiptNumber

tj:
Just adds `ReceiptNumber`. Apparently I'm using an older version of the API, feel free to ignore if this field is no longer relevant! I imagine it's not really "secure" to expose any stripe id's though, and this one looks a bit nicer haha.

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -44,35 +44,36 @@ type InvoiceLineListParams struct {
 // Invoice is the resource representing a Stripe invoice.
 // For more details see https://stripe.com/docs/api#invoice_object.
 type Invoice struct {
-	ID           string            `json:"id"`
-	Live         bool              `json:"livemode"`
-	Amount       int64             `json:"amount_due"`
-	Attempts     uint64            `json:"attempt_count"`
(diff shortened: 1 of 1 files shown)


brandur:
Thanks @tj!

LGTM.

BTW, due to the static nature of the language, the Go bindings behave a little differently than most other ones in that they lock requests into [a predefined version of Go](https://github.com/stripe/stripe-go/blob/224faec412d280d27b46853512a30a228e8df2ac/stripe.go#L23), so what your account is set to isn't as important. `ReceiptNumber` is still current though.

brandur:
Released as 16.6.0.

tj:
awesome thanks! Strangely it seems like ReceiptNumber is not always populated, sometimes it is sometimes it isnt haha

remistr:
@tj the value is populated only when an email receipt has been sent for the associated charge. By default, in Test mode we don't send an email receipt but if you do send one in the dashboard for example you'll see that value is then set.

tj:
Ahh I see, it would be cool to have a similar value to use for custom receipt emails

PR #293, opened 2016-08-10
support expanding transaction sources

MDrollette:
This updates `Transaction.Src` to follow the same pattern as [PaymentSource](https://github.com/stripe/stripe-go/blob/master/payment_source.go#L97) to allow for expanding the polymorphic source property.

Note that there may be other object types that are valid for "source", but because they aren't listed anywhere I only added the ones I was aware of. If an unknown source shows up, it will still fall back gracefully to only having the `ID` field populated.

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -11,6 +11,43 @@ type TransactionStatus string
 // "application_fee_refund", "transfer", "transfer_cancel", "transfer_failure".
 type TransactionType string
 
+// TransactionSourceType consts represent valid balance transaction sources.
+type TransactionSourceType string
+
+const (
+	// TransactionSourceCharge is a constant representing a transaction source of charge
+	TransactionSourceCharge TransactionSourceType = "charge"
+
+	// TransactionSourceAdjustment is a constant representing a transaction source of dispute
+	TransactionSourceDispute TransactionSourceType = "dispute"
+
+	// TransactionSourceFee is a constant representing a transaction source of application_fee
+	TransactionSourceFee TransactionSourceType = "application_fee"
+
+	// TransactionSourceRefund is a constant representing a transaction source of refund
+	TransactionSourceRefund TransactionSourceType = "refund"
+
+	// TransactionSourceReversal is a constant representing a transaction source of reversal
+	TransactionSourceReversal TransactionSourceType = "reversal"
+
+	// TransactionSourceTransfer is a constant representing a transaction source of transfer
+	TransactionSourceTransfer TransactionSourceType = "transfer"
+)
+
+// TransactionSource describes the source of a balance Transaction.
+// The Type should indicate which object is fleshed out.
+// For more details see https://stripe.com/docs/api#retrieve_balance_transaction
+type TransactionSource struct {
+	Type     TransactionSourceType `json:"object"`
+	ID       string                `json:"id"`
+	Charge   *Charge               `json:"-"`
+	Dispute  *Dispute              `json:"-"`
+	Fee      *Fee                  `json:"-"`
+	Refund   *Refund               `json:"-"`
+	Reversal *Reversal             `json:"-"`
+	Transfer *Transfer             `json:"-"`
+}
+
 // BalanceParams is the set of parameters that can be used when retrieving a balance.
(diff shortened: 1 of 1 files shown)


balance.go
```diff
@@ -11,6 +11,43 @@ type TransactionStatus string
...
+	TransactionSourceCharge TransactionSourceType = "charge"
+
+	// TransactionSourceRefund is a constant representing a transaction source of refund
+	TransactionSourceRefund TransactionSourceType = "refund"
+
+	// TransactionSourceTransfer is a constant representing a transaction source of transfer
+	TransactionSourceTransfer TransactionSourceType = "transfer"
+
+	// TransactionSourceReversal is a constant representing a transaction source of reversal
+	TransactionSourceReversal TransactionSourceType = "reversal"
```
brandur: As with the other PR, mind alphabetizing this list?

balance.go
```diff
@@ -11,6 +11,43 @@ type TransactionStatus string
...
+)
+
+// TransactionSource describes the source of a balance Transaction.
+// The Type should indicate which object is fleshed out (eg. Fee or Charge)
```
brandur: Same sort of copy paste problem. Since we have so many potential object types, maybe just remove the part in parenthesis.

balance.go
```diff
@@ -11,6 +11,43 @@ type TransactionStatus string
...
+// For more details see https://stripe.com/docs/api#retrieve_balance_transaction
+type TransactionSource struct {
+	Type     TransactionSourceType `json:"object"`
+	ID       string                `json:"id"`
+	Dispute  *Dispute              `json:"-"`
+	Fee      *Fee                  `json:"-"`
+	Charge   *Charge               `json:"-"`
+	Refund   *Refund               `json:"-"`
+	Transfer *Transfer             `json:"-"`
+	Reversal *Reversal             `json:"-"`
```
brandur: And please alphabetize here as well.

balance.go
brandur: Oh and this is awkward, but I think that `reversal` here will come out as `transfer_refund`.

balance.go
```diff
@@ -11,6 +11,43 @@ type TransactionStatus string
 // "application_fee_refund", "transfer", "transfer_cancel", "transfer_failure".
 type TransactionType string
 
+// TransactionSourceType consts represent valid balance transaction sources.
+type TransactionSourceType string
+
+const (
+	// TransactionSourceAdjustment is a constant representing a transaction source of dispute
+	TransactionSourceDispute TransactionSourceType = "dispute"
```
brandur: I don't think that `dispute` will ever be at type returned here.

See the `type` field here: https://stripe.com/docs/api#balance_transaction_object

MDrollette:
I never even thought to alphabetize ordering... that makes sense. Should be all set now.

brandur:
Thanks @MDrollette! I left a couple other minor comment since your new push, but this looks great.

One other thing to note: this is a breaking change on account of the changing return type of `src`. This isn't that big of a deal with the Go bindings, but it's worth pointing out.

MDrollette:
@brandur The "Type" here actually corresponds to the [object type](https://github.com/MDrollette/stripe-go/blob/7361da60dfe95ef26bf310fa850ed50bb555b1c7/balance.go#L41) on the expanded "Source", not to the Type on the [Transaction](https://github.com/MDrollette/stripe-go/blob/7361da60dfe95ef26bf310fa850ed50bb555b1c7/balance.go#L93)

So for a balance transaction related to a Dispute, the transaction type will be "adjustment" but the expanded source object will be a "dispute".

balance.go
MDrollette: @brandur The "Type" here actually corresponds to the object type on the expanded "Source", not to the Type on the Transaction

So for a balance transaction related to a Dispute, the transaction type will be "adjustment" but the expanded source object will be a "dispute".

brandur: Ah, I see! Thanks.

brandur:
LGTM.

brandur:
Released as 17.0.0.

PR #295, opened 2016-08-11
include originating transaction on application fee object

MDrollette:
The diff looks bigger than it is... only added line 30 for the `OriginatingTransaction` property.


brandur:
+1. Thanks @MDrollette!

LGTM.

brandur:
Released as 16.8.0.

PR #296, opened 2016-08-11
allow nil params when getting a refund

MDrollette:
Updated to match the Charge client. Prevents nil pointer exception when passing a nil params argument.


brandur:
LGTM!

@MDrollette Sorry about the crazy delay on this stuff. I've been a little weighed down responding to support tickets about TLS problems, and am now just starting to get some of my backlog cleaned out.

brandur:
Released as 16.7.1.

PR #299, opened 2016-08-31
Add support for Source objects

spolu:
Add support for Sources to Go.

### DIFF
--- a/address.go
+++ b/address.go
@@ -1,0 +1,11 @@
+package stripe
+
+// Standard address parameters.
+type AddressParams struct {
+	Line1      string
+	Line2      string
+	City       string
+	State      string
+	PostalCode string
+	Country    string
+}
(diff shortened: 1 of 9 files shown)


tejasmanohar:
+100

spolu:
r? @brandur 
cc @andrewpthorp

brandur:
Wow, crazy PR here! Thanks :)

LGTM.

brandur:
Released as 17.3.0.

spolu:
Thanks @brandur!!

address.go
```diff
@@ -0,0 +1,11 @@
+package stripe
+
+// Standard address parameters.
+type AddressParams struct {
+	Line1      string
+	Line2      string
+	City       string
+	State      string
+	PostalCode string
```
remistr: Why use `PostalCode` and not `Zip` like we do on Card?

spolu: ZIP is a US concept

PR #303, opened 2016-09-08
Add various missing fields from legal_entity

### LINKED ISSUE #301: Account's LegalEntity: missing address_kana and address_kanji
Hi Everybody,

This is not something very urgent. 

Update account has address_kana and address_kanji (https://stripe.com/docs/api#update_account) but LegalEntity (https://github.com/stripe/stripe-go/blob/master/account.go) does not have those.

Also: business_name_kana, business_name_kanji, first_name_kana, first_name_kanji, last_name_kana, last_name_kanji, personal_address_kana, personal_address_kanji, phone_number, gender

UPDATE: maiden_name

brandur:
Adds a lot of missing fields to account's legal_entity; most focusing on
alternate fields for Kana and Kanji. This also refactors address
encoding so that we can re-use it in a whole bunch of other places.

Fixes #301.

r? @guillaumehenriot Mind taking a quick look at this one when you get a
chance? Thanks!

### DIFF
--- a/account.go
+++ b/account.go
@@ -288,8 +288,6 @@ type Owner struct {
 	Last         string               `json:"last_name"`
 	DOB          DOB                  `json:"dob"`
 	Address      Address              `json:"address"`
-	AddressKana  Address              `json:"address_kana"`
-	AddressKanji Address              `json:"address_kanji"`
 	Verification IdentityVerification `json:"verification"`
 }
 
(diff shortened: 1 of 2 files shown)


guillaumehenriot:
@brandur I don't think we handle the the Kana and Kanji version of the address for the Additional Owners in the API yet, no? Otherwise the rest looks good.

brandur:
> @brandur I don't think we handle the the Kana and Kanji version of the address for the Additional Owners in the API yet, no? Otherwise the rest looks good.

@guillaumehenriot Ah, shoot. I misread the docs. I just pushed a fix. Thanks!

brandur:
Released as 17.4.0.

PR #305, opened 2016-09-13
Add receipt_number and status to refund object

domwong:


brandur:
Great! Thanks @domwong!

brandur:
Released as 17.4.0.

PR #309, opened 2016-09-14
<error>.Error() method should have a literal receiver

tejasmanohar:
Tried doing
```go
func someFn(err error) {
  stripeError, isStripeError := err.(stripe.Error)
  // ...
}
```
but got
```
impossible type assertion:
	".../github.com/stripe/stripe-go".Error does not implement error (Error method has pointer receiver)
```
This should fix that.

### DIFF
--- a/error.go
+++ b/error.go
@@ -57,7 +57,7 @@ type Error struct {
 }
 
 // Error serializes the error object to JSON and returns it as a string.
-func (e *Error) Error() string {
+func (e Error) Error() string {
 	ret, _ := json.Marshal(e)
 	return string(ret)
 }
@@ -68,7 +68,7 @@ type APIConnectionError struct {
 }
 
 // Error serializes the error object to JSON and returns it as a string.
-func (e *APIConnectionError) Error() string {
+func (e APIConnectionError) Error() string {
 	return e.stripeErr.Error()
 }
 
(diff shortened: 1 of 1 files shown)


brandur:
Hi @tejasmanohar,

So currently, an assertion based on a pointer does work. e.g.

``` go
package stripe_test

import (
	"fmt"
	"testing"
)

func TestCast(t *testing.T) {
	err := fmt.Errorf("hello")
	stripeError, isStripeError := err.(*stripe.Error)
	fmt.Printf("%v %v\n", stripeError, isStripeError)

	err = &stripe.Error{}
	stripeError, isStripeError = err.(*stripe.Error)
	fmt.Printf("%v %v\n", stripeError, isStripeError)
}
```

Prints:

```
null false
{"type":"","message":""} true
```

And based on my reading of [this article on error handling in Go](https://blog.golang.org/error-handling-and-go), I don't think this is too unidiomatic. You'll see this approach in the `json` package as well for example:

``` go
if err := dec.Decode(&val); err != nil {
    if serr, ok := err.(*json.SyntaxError); ok {
        line, col := findLine(f, serr.Offset)
        return fmt.Errorf("%s:%d:%d: %v", f.Name(), line, col, err)
    }
    return err
}
```

I don't mind making the change if there's some other advantage that I haven't considered. If there is, could you describe it in more detail? Thanks!

tejasmanohar:
Nope, that make sense. My bad- thanks!

PR #310, opened 2016-09-15
Removes the need to pass DOB and Type to update Legal Entity

### LINKED ISSUE #308: Can't update legal entity without DOB
It default to 0 and Stripe API returns a validation error.
It also sends empty string for the type, which is unexpected.

```go
params := &stripe.AccountParams{
  LegalEntity: &stripe.LegalEntity{
    PersonalID: "000-00-0000",
  },
}

account.Update(req.AccountID, params)
```

```
&{values:[
  {Key:legal_entity[type] Value:} 
  {Key:legal_entity[dob][day] Value:0} 
  {Key:legal_entity[dob][month] Value:0} 
  {Key:legal_entity[dob][year] Value:0} 
  {Key:legal_entity[personal_id_number] Value:000-00-0000} 
]}
```

```json
{"field_name":"legal_entity[dob][day]", "code":"", "messages":["Not a valid day"]}
```

guillaumehenriot:
It fixes #308.

Prior to that we needed to pass Type and the full DOB to update Legal Entity. These parameters are now checked to see if empty like everything else.

r? @brandur

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -273,7 +273,7 @@ func TestAccountUpdateLegalEntity(t *testing.T) {
 	}
 
 	if acct.LegalEntity.Address.Line1 != params.LegalEntity.Address.Line1 {
-		t.Error("The account address line1 %s does not match the params address line1: %s", acct.LegalEntity.Address.Line1, params.LegalEntity.Address.Line1)
+		t.Errorf("The account address line1 %v does not match the params address line1: %v", acct.LegalEntity.Address.Line1, params.LegalEntity.Address.Line1)
 	}
 }
 
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -238,6 +238,45 @@ func TestAccountUpdate(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


account/client_test.go
```diff
@@ -238,6 +238,45 @@ func TestAccountUpdate(t *testing.T) {
...
+		},
+	}
+
+	acct, err = Update(acct.ID, params)
+	if err != nil {
+		t.Error(err)
+	}
+
+	if acct.LegalEntity.Address.Line1 != params.LegalEntity.Address.Line1 {
+		t.Error("The account address line1 %s does not match the params address line1: %s", acct.LegalEntity.Address.Line1, params.LegalEntity.Address.Line1)
```
brandur: I think you're looking for `Errorf` here.

guillaumehenriot: Oops, you're right. I'll change that.

brandur:
LGTM!

brandur:
Released as 17.4.1.

PR #311, opened 2016-09-16
Set commonParams on order.Update

pingmulan:
Small fix to allow updating orders on connected stripe accounts.


brandur:
Thanks a lot for taking the time to submit this!

brandur:
Released as 17.4.2.

PR #313, opened 2016-09-22
add retries to iterator interface

### LINKED ISSUE #300: Retry iterator request?
If an iterator request fails, how would we handle retries? I guess we need to grab the last ID from `Current` and manually fill in the `starting_after` param? Seems like there could be a nice retry or simpler, request() function on the iterator.

tejasmanohar:
Resolves #300. Let me know if you think these settings are reasonable. I only chose them because they're what we use internally for most all requests at https://segment.com/sources.

// cc @brandur


brandur:
Thanks for the shot at this @tejasmanohar! I definitely appreciate your persistence while tracking this down :)

A few broad comments:

1. I think if at all possible we want to implement retries closer to the network layer so that all API calls could take advantage of them.
2. It'll probably be a little awkward in Go, but we'll want to put some testing in place to make sure that the behavior actually works as we expected. The compiler is nice, but not enough.
3. We're almost certainly not going to want to pull in a new dependency to implement retries.

I realize these are some fairly large-ish design asks though. If you think you could rework to consider them, that would be awesome, but otherwise we can also take a shot at implementing this instead. Thanks again!

tejasmanohar:
@brandur Alright. I agree w/ 1 & 2. 3 is a bit iffy, but if you'd really like... I'll copy Backo#Duration here :P

opening another PR so I can rework branches on my fork (pointing stuff to there for now)

PR #318, opened 2016-09-23
Add subscription items

kyleconroy:


subitem/client.go
```diff
@@ -0,0 +1,192 @@
...
+
+		params.AppendTo(body)
+
+		lp = &params.ListParams
+		p = params.ToParams()
+	}
+
+	return &Iter{stripe.GetIter(lp, body, func(b *stripe.RequestValues) ([]interface{}, stripe.ListMeta, error) {
+		list := &stripe.SubList{}
+		err := c.B.Call("GET", "/subscriptions", c.Key, b, p, list)
```
brandur: @kyleconroy I'm not sure whether this is finished, but this doesn't seems quite right. I believe the new endpoint is at `/subscription_items`, and if so, it means that not only is this wrong, but your test suite is either broken or not exhaustive enough to be able to test for the failure.

subitem/client.go
```diff
@@ -0,0 +1,192 @@
...
+	var commonParams *stripe.Params
+
+	if params != nil {
+		body = &stripe.RequestValues{}
+		params.AppendTo(body)
+		commonParams = &params.Params
+	}
+
+	item := &stripe.Sub{}
+	err := c.B.Call("DELETE", fmt.Sprintf("/subscriptions_item/%v", id), c.Key, body, commonParams, item)
```
brandur: And same with here (and the other CRUD endpoints). The patch merged to Ruby was for `subscription_items` (i.e. non-plural subscription), so one of these must be wrong.

subitem/client.go
kyleconroy: Yeah, this PR isn't ready yet. I'll ping you for review once it's ready.

kyleconroy:
@brandur this PR is now ready for review

subitem/client_test.go
```diff
@@ -0,0 +1,155 @@
...
+		if err != nil {
+			t.Errorf("del err: %s", err)
+		}
+		if !item.Deleted {
+			t.Errorf("item should be deleted %#v", item)
+		}
+	}
+}
+
+func TestCreateAndDelItem(t *testing.T) {
```
brandur: Was leaving this test empty intentional?

kyleconroy: Whoops, nope. I'll go ahead and fill it out.

kyleconroy:
@brandur added that missing test. Mind taking another look?

brandur:
LGTM.

PR #321, opened 2016-09-30
Split sub item param structs into two.

kyleconroy:
We were sharing a single struct between two API calls. This made things
confusing, as certain parameters were not allowed. Also remove the
per-item proration parameters, as those are not supported.


brandur:
LGTM.

Issue #322, opened 2016-10-03
Transfer.SourceTx has wrong type


comstud:
Transfer.SourceTx has a type of *Transaction but it should be *TransactionSource

PR incoming.

brandur:
Closed by #323.

PR #323, opened 2016-10-03
Fix issue #322: Transfer.SourceTx has wrong type

comstud:
This fixes the type of Transfer.SourceTx to be *TransactionSource


brandur:
@comstud Oops, yes! Thank you. And wow, thanks for including tests too.

LGTM.

Issue #324, opened 2016-10-03
refund.charge is expandable but has type string


comstud:
When listing or fetching refunds, 'charge' is expandable, however, Refund.Charge has a type of string.

This really should be a *Charge type, although it would be a breaking change.

brandur-stripe:
Closing this out as fixed.

PR #325, opened 2016-10-03
Fix issue #324: Refund.Charge should be a *Charge

comstud:
This flips the type of Refund.Charge to be a *Charge instead of a
string. 'charge' is expandable when querying for refunds, so the type
needs to be the object.

Unfortunately this is a breaking change for people already using
Refund.Charge.


brandur:
Cool! Thanks, and great tests!

This sort of breaking change is a little unfortunate, but this seems useful, and the compiler error should be quite easy to understand and compensate for on upgrade. I'll ship a major version.

LGTM.

PR #328, opened 2016-10-08
Add a couple missing fields

tejasmanohar:
I spotted that a couple fields were missing so added them. They're all in https://stripe.com/docs/api

### DIFF
--- a/transfer.go
+++ b/transfer.go
@@ -71,30 +71,30 @@ type TransferListParams struct {
 // Transfer is the resource representing a Stripe transfer.
 // For more details see https://stripe.com/docs/api#transfers.
 type Transfer struct {
-	ID                 string              `json:"id"`
-	Live               bool                `json:"livemode"`
-	Amount             int64               `json:"amount"`
-	AmountReversed     int64               `json:"amount_reversed"`
-	Currency           Currency            `json:"currency"`
-	Created            int64               `json:"created"`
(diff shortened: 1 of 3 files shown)


transfer.go
```diff
@@ -71,29 +71,30 @@ type TransferListParams struct {
...
+	Meta               map[string]string   `json:"metadata"`
+	Bank               *BankAccount        `json:"bank_account"`
+	Card               *Card               `json:"card"`
+	Recipient          *Recipient          `json:"recipient"`
+	Statement          string              `json:"statement_descriptor"`
+	Reversals          *ReversalList       `json:"reversals"`
+	Reversed           bool                `json:"reversed"`
+	SourceTx           *TransactionSource  `json:"source_transaction"`
+	SourceType         TransferSourceType  `json:"source_type"`
+	DestinationPayment string              `json:"destination_payment"`
```
brandur: Thanks! Can you rename this one to `DestinationPayment`? The naming choices in this library kill me, but I think it's a good idea to stay consistent with other struct fields.

tejasmanohar: @brandur isn't it called `DestinationPayment`?

tejasmanohar: ahh, fixed.

brandur:
LGTM.

Thanks!

brandur:
Released as 18.2.0.

PR #330, opened 2016-10-10
Add `NoDiscountable` to `InvoiceItemsParams`

### LINKED ISSUE #329: Invoice Item discountable defaults to true
According to the Stripe documentation, the `discountable` argument for invoice item 
> Defaults to false for prorations or negative invoice items, and true for all other invoice items.

For my project, I actually want a Stripe coupon to only discount the plan and not the invoice items. However, setting `discountable` to false is ignored here: https://github.com/stripe/stripe-go/blob/master/invoiceitem/client.go#L41 . I will replicate here for convenience:

```
	if params.Discountable {
		body.Add("discountable", strconv.FormatBool(true))
	}
```

The `discountable=false` argument is never added to the request to Stripe and it will be implied as true, which is not I wanted.

I propose we change the logic to at least:
```
	if (params.Amount > 0 && !params.Discountable) || (params.Amount < 0 && params.Discountable) {
		body.Add("discountable", strconv.FormatBool(params.Discountable))
	}
```
 or

```
	if (params.Amount > 0) != params.Discountable {
		body.Add("discountable", strconv.FormatBool(params.Discountable))
	}
```
This change will add the discountable param to the request if it is different from the default behavior.

brandur:
Currently, `Discountable` cannot be explicitly set to `false`, which is
a problem because for certain types of invoice items it will be
implicitly set to true on the server side.

This patch adds `NoDiscountable` which allows an explicit `false`. This
is modeled after current SDK convention (for exmpale `NoClosed` as part
of `InvoiceParams`).

Fixes #329.

r? @kyleconroy Hey Kyle, would you mind taking a look at this? Thanks!

### DIFF
--- a/invoiceitem.go
+++ b/invoiceitem.go
@@ -11,6 +11,7 @@ type InvoiceItemParams struct {
 	Currency           Currency
 	Invoice, Desc, Sub string
 	Discountable       bool
+	NoDiscountable     bool
 }
 
 // InvoiceItemListParams is the set of parameters that can be used when listing invoice items.
--- a/invoiceitem/client.go
+++ b/invoiceitem/client.go
@@ -39,6 +39,8 @@ func (c Client) New(params *stripe.InvoiceItemParams) (*stripe.InvoiceItem, erro
 
 	if params.Discountable {
 		body.Add("discountable", strconv.FormatBool(true))
+	} else if params.NoDiscountable {
+		body.Add("discountable", strconv.FormatBool(false))
 	}
 
 	params.AppendTo(body)


kjc:
@brandur:  I'm not Kyle.  Same problem in [`stripe-python`'s #253](https://github.com/stripe/stripe-python/pull/253), too.

brandur:
@kjc My apologies. I was typing from (faulty) memory, heh.

r? @kyleconroy Hey Kyle, would you mind taking a look at this? Thanks!

kyleconroy:
👍 

Thanks!

brandur:
Released as 18.3.0.

PR #334, opened 2016-10-13
add order status transitions field

tejasmanohar:


order.go
```diff
@@ -109,6 +110,15 @@ type OrderListParams struct {
 	Status OrderStatus
 }
 
+// StatsuTransitions are the timestamps at which the order status was updated
+// https://stripe.com/docs/api#order_object
+type StatusTransitions struct {
+	Canceled  int64 `json:"canceled"`
+	Fulfilled int64 `json:"fulfilled"`
```
brandur: This is brutal, but I just noticed that the API field for "fulfilled" is actually misspelled:
https://stripe.com/docs/api#order_object-status_transitions

We can fix that, but it'll be a breaking change so the API version tied to these bindings will always be the misspelled version anyway. Can you change the JSON component of this to `fulfiled`? We can leave the actual Go field name spelled correctly.

Thanks for the patch!

tejasmanohar: fixed

tejasmanohar:
@brandur should be read yfor merge

brandur-stripe:
Thanks!

brandur:
Thanks!

Released as 18.10.0.

PR #336, opened 2016-10-14
Support api permissions errors

jflinter:
One note - our error-checking logic doesn't actually look at the HTTP status code, but rather the error code returned by Stripe. Thus we'll need to ensure that `permissions_error` corresponds to 403.

r? @brandur

### DIFF
--- a/error.go
+++ b/error.go
@@ -14,7 +14,7 @@ const (
 	ErrorTypeAuthentication ErrorType = "authentication_error"
 	ErrorTypeCard           ErrorType = "card_error"
 	ErrorTypeInvalidRequest ErrorType = "invalid_request_error"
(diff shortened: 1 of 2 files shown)


jflinter:
@brandur i _think_ this test failure is spurious?

error.go
```diff
@@ -14,6 +14,7 @@ const (
 	ErrorTypeAuthentication ErrorType = "authentication_error"
 	ErrorTypeCard           ErrorType = "card_error"
 	ErrorTypeInvalidRequest ErrorType = "invalid_request_error"
+	ErrorTypePermissions    ErrorType = "permissions_error"
```
brandur: Lets go singular here too.

jflinter:
re-r? @brandur

PR #337, opened 2016-10-18
add the status field to the request for listing subscriptions

sjsnider:
I wanted to be able to request subscriptions that had been canceled (potentially other statuses later) and the docs showed how to do it with the "status" field(https://stripe.com/docs/api/go#list_subscriptions), but that field wasn't present in the Go library, so this adds it.

I tested it out locally on my machine and it works.  I didn't actually run `make tests` though because the last PR I submitted I was told "Unfortunately, a large part of the test suite doesn't currently run in CI and is therefore very prone to failure as it falls out of date for one reason or another. This looks good as is!"

### DIFF
--- a/sub.go
+++ b/sub.go
@@ -40,6 +40,7 @@ type SubListParams struct {
 	ListParams
 	Customer string
 	Plan     string
+	Status   SubStatus
 }
 
 // Sub is the resource representing a Stripe subscription.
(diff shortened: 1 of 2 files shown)


brandur:
Looks good to me. Thanks!

> I tested it out locally on my machine and it works. I didn't actually run make tests though because the last PR I submitted I was told "Unfortunately, a large part of the test suite doesn't currently run in CI and is therefore very prone to failure as it falls out of date for one reason or another. This looks good as is!"

Yes, this is still true unfortunately. Thanks for making the effort to do some sanity checks though.

It looks like Travis is still underwater, but I pulled this down locally to double-check that the basic tests still run for me.

brandur:
Released as 18.5.0.

PR #342, opened 2016-11-03
Add expandable Application to Charge

dpetrovics-stripe:
Adds `Application` to `Charge`. Added some basic deserialization tests and added an expansion for an existing functional test.

r? @brandur

### DIFF
--- a/application.go
+++ b/application.go
@@ -1,6 +1,25 @@
 package stripe
 
+import "encoding/json"
+
 type Application struct {
 	ID   string `json:"id"`
 	Name string `json:"name"`
+}
+
+// UnmarshalJSON handles deserialization of an Application.
+// This custom unmarshaling is needed because the resulting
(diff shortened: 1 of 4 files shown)


brandur:
@dpetrovics-stripe Looking good!

One other thing I think you might want is a custom JSON deserializer that will expand conditionally like this one:
https://github.com/stripe/stripe-go/blob/84484ce43b9b0edbdfde86a4c8f9d082df9e3638/charge.go#L158,L173

That way even if an application is _not_ expanded, the ID will still be injected into an otherwise blank struct.

ptal @dpetrovics-stripe

dpetrovics-stripe:
Ah, yes - my oversight. Added!

ptal @brandur

brandur:
LGTM.

Thanks @dpetrovics-stripe!

dpetrovics-stripe:
Released as `18.9.0`

PR #344, opened 2016-11-07
Detect nil in SetHTTPClient

yyoshiki41:
Detect `nil` when we set our custom `http.Client`.

This patch changes function's return value.
So, let me know what you think.

In other ways, check `nil` in `NewBackends()` like below.

```go
func NewBackends(httpClient *http.Client) *Backends {
        if httpClient == nil {
		// handle error...
		return nil
        }
	return &Backends{
		...
	}
}
```

Thanks!

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -97,8 +97,12 @@ var backends Backends
 // SetHTTPClient overrides the default HTTP client.
 // This is useful if you're running in a Google AppEngine environment
 // where the http.DefaultClient is not available.
-func SetHTTPClient(client *http.Client) {
+func SetHTTPClient(client *http.Client) error {
+	if client == nil {
+		return errors.New("Cannot use empty HTTP client.")
+	}
 	httpClient = client
+	return nil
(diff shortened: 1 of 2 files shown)


brandur:
Hi Yoshiki,

Is there a particular impetus for this change? I would think that although it's possible to pass in a `nil` value here, it's pretty unlikely to be a mistake that will make it past your local development box. Although more checks are probably a good thing, I'm not sure that it's worth introducing a breaking API change over.

yyoshiki41:
@brandur 
thanks for your reply!

> Is there a particular impetus for this change?
I would think that although it's possible to pass in a nil value here, it's pretty unlikely to be a mistake that will make it past your local development box.

No, it's a trivial matter.
After read your comments, I agree with your idea and it's better not to change APIs in this case.

I'll close this PR.
Thanks for your help!

brandur:
@yyoshiki41 No worries. Thanks!

PR #347, opened 2016-11-14
Add paymentsource.Client to api.Client

natethebosch:
api.Client doesn't provide access to paymentsource.Client and
paymentsource.Client doesn't provide a constructor similar to
api.Client.


brandur:
@natethebosch Thanks!

Wow, the Travis queue got really backed up there for a long time, but it finally finished. It looks like you may need to run gofmt on the file to get it passing. Would you mind doing that and recommitting? Thank you!

natethebosch:
updated @brandur

brandur:
Released as 18.13.0.

PR #345, opened 2016-11-10
Add creating a bank account on a customer

prestonmorgan:
Currently only supports creating an external bank account.

PR #348, opened 2016-11-14
Adds ability to add bank account to customer

natethebosch:
Previously only allowed adding to account.


brandur:
Thanks! And oof — there's an almost identical pull open in #345. I'm a little behind on my backlog and haven't had a chance to take a look in depth yet which cause this confusion. Apologies for the delays here. I'll take a look at this very soon.

brandur:
I released as #345 as 18.12.0. Thanks again for this patch!

PR #349, opened 2016-11-22
Added 3DS retrieve endpoint

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @bct 

Adds a method for retrieving 3DS auths.

### DIFF
--- a/threedsecure/client_test.go
+++ b/threedsecure/client_test.go
@@ -14,6 +14,7 @@ func init() {
 }
 
 func TestThreeDSecureNew(t *testing.T) {
+	// Test creating a 3D Secure auth
 	customerParams := &stripe.CustomerParams{}
 	customerParams.SetSource(&stripe.CardParams{
 		Name:   "Stripe Tester",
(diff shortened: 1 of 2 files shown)


threedsecure/client_test.go
```diff
@@ -32,22 +33,32 @@ func TestThreeDSecureNew(t *testing.T) {
 		ReturnURL: "https://test.com",
 	}
 
-	target, err := New(threeDSecureParams)
+	tdsrc, err := New(threeDSecureParams)
```
brandur: Sorry for the nitpick-y feedback, but would you mind changing this naming to something a little more descriptive? IMO, abbreviated local names are suitable for very well-scoped use, but it's better to avoid them more generally. (I'm not sure what `tdsrc` is here for example, I'd guess "3-D Secure ResourCe?)

Something like `threeDSecure` would be fine.

brandur:
Left a very minor comment, but looks good to me!

ptal? @ob-stripe

threedsecure/client_test.go
ob-stripe: Good point. I used `tdsrc` because it's the prefix for this resource's ID, but I'm not too sure what it stands for either!

I replaced with `threeDSecure` per your suggestion.

brandur:
Thanks @ob!

brandur:
Released as 18.14.0.

PR #351, opened 2016-12-05
Truncate tax_percent to four decimals instead of two

### LINKED ISSUE #350: Subscription tax_percent only supports 2 decimal floats.
I need to to set the `tax_percent` value of my subscriptions to a 3 decimal number (eg `8.875`) currently this gets rounded to 2 decimals.

See: 
- https://github.com/stripe/stripe-go/blob/master/sub/client.go#L79
- https://github.com/stripe/stripe-go/blob/master/sub/client.go#L203

brandur:
We used to only support two decimals for tax_percent, but realized later
that we needed four. The API was updated to properly support this, but
the Go bindings continued to truncate to two.

This patch fixes that problem.

Fixes #350.

r? @grey-stripe Hey Marc, can you take a look at this? Thanks!

### DIFF
--- a/customer/client.go
+++ b/customer/client.go
@@ -66,7 +66,7 @@ func (c Client) New(params *stripe.CustomerParams) (*stripe.Customer, error) {
 		}
 
 		if params.TaxPercent > 0 {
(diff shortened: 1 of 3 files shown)


grey-stripe:
LGTM

brandur:
Released as 18.14.1.

PR #353, opened 2016-12-07
Remove the `SubProrationDateNow` parameter

### LINKED ISSUE #352: SubProrationDateNow never gets checked in invoices.GetNext()
stripe-go version: 18.13.0
golang version: 1.7.3

InvoiceParams is [defined](https://github.com/stripe/stripe-go/blob/master/invoice.go#L11) with a field called SubProrationDateNow, which is a bool. I intended on using this with the [invoices.GetNext()](https://github.com/stripe/stripe-go/blob/76efd227ee7e12e332412a515485e96894824efc/invoice/client.go#L169) method, but noticed that it doesn't ever get checked. I would expect that it would check if it was true, then set SubProrationDate to the current time. Was this deprecated at some time?

I know that I can just get the current time manually and set SubProrationDate to that, but I thought it was odd that SubProrationDateNow is left unchecked.

brandur:
This appears to have been introduced accidentally and not actually used.

Fixes #352.

r? @ob-stripe Would you mind taking a look at this one? Thank you!

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -18,7 +18,6 @@ type InvoiceParams struct {
 	SubPlan              string
 	SubNoProrate         bool
 	SubProrationDate     int64
-	SubProrationDateNow  bool
 	SubQuantity          uint64
 	SubTrialEnd          int64
 	TaxPercent           float64


remi-stripe:
@brandur I don't think you want to remove it. I think you want to support it and add it to the Retrieve Upcoming Invoice [API](https://stripe.com/docs/api#upcoming_invoice). This allows you to simulate an update to a subscription and see what the invoice would look like if you were to do it. If your subscription is on a trial, you might want to end the trial during the upgrade and pass [subscription_proration_date](https://stripe.com/docs/api#upcoming_invoice-subscription_proration_date) set to `"now"`.

remi-stripe:
Ah yeah nevermind, I misread the doc and I see what you meant here. There's a completely different parameter to end the trial itself.

remi-stripe:
LGTM

brandur:
@remi-stripe Hah, yeah. I was a little uncertain about this, so I read over that doc field like three times to try and make sure that I wasn't missing something that I might have seen when I originally introduced this field.

Thanks for jumping in! It's really good to have another set of eyes on it.

brandur:
Released as 19.0.0.

@ob-stripe, you're off the hook :)

ob-stripe:
Glad I could help! :P

PR #354, opened 2016-12-08
Fix the Account Update API to support document verification

remi-stripe:
r? @brandur

### DIFF
--- a/account.go
+++ b/account.go
@@ -409,6 +409,10 @@ func (l *LegalEntity) AppendDetails(values *RequestValues) {
 		values.Add("legal_entity[business_vat_id]", l.BusinessVatID)
 	}
 
+	if l.Verification.Document != nil {
+		values.Add("legal_entity[verification][document]", l.Verification.Document.ID)
+	}
+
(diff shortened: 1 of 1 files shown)


account.go
```diff
@@ -409,6 +409,10 @@ func (l *LegalEntity) AppendDetails(values *RequestValues) {
 		values.Add("legal_entity[business_vat_id]", l.BusinessVatID)
 	}
 
+	if l.Verification.Document != nil {
```
brandur: Thanks Remi! Should we do on `nil` check on `l.Verification` as well?

remi-stripe: @brandur I originally did that but `Verification` is not a pointer for some reason and I wasn't sure if there was a "go-like" way to check its presence otherwise.

brandur: Ack you're right. I was reading `Verification` off the `Account` struct instead of the `LegalEntity` struct (the former is a pointer but the latter isn't). That's fine then.

brandur:
Released as 19.0.1.

Issue #355, opened 2016-12-29
TrialPeriod draws from the wrong field for Subscription interactions


kujenga:
When using this client library to create a subscription, the field within `stripe.SubParams` to setup the TrialPeriod is configured incorrectly. Instead of using the actual value of `TrailPeriod`, it uses the value of `TrialEnd`. It looks like this is a simple typo:

https://github.com/stripe/stripe-go/blob/76efd227ee7e12e332412a515485e96894824efc/sub/client.go#L68-L70

```go
if params.TrialPeriod > 0 {
	body.Add("trial_period_days", strconv.FormatInt(params.TrialEnd, 10))
}
```

It looks like the bug was introduced in this commit, which also added the functionality: https://github.com/stripe/stripe-go/commit/fb4dda969c46db1c9d669be2b5bdbfb7e8c1a393

I'll be creating a fork with a fix and submit a PR shortly.

dpetrovics-stripe:
Thanks for catching this, @kujenga! I see that we have some model tests for subscriptions here: https://github.com/stripe/stripe-go/blob/master/sub/client_test.go, but Im not sure if we have a good place to check for Go object to body param mapping. cc @brandur in case he has any opinion on how / if to add a test. Otherwise looks good!

brandur:
Oops, sorry @dpetrovics-stripe, I missed this before merging the other one.

In general, I'd really like to have testing for this kind of thing, but the whole suite is set up so that instead of testing that you're properly serializing parameters, you're instead testing that you're serializing them, sending them up to the server, and testing that the server responds with something that seems reasonable. As such, the tests are very non-specific (if an error occurs it's difficult to know what might have caused it) and incredibly brittle.

We should overhaul the suite at some point to make this better, but I think just having no test for single fields like this might be our wisest course of action right now.

dpetrovics-stripe:
Thanks! Sounds totally reasonable - marking this as resolved.

PR #356, opened 2016-12-29
Fix the Subscription TrialPeriod param to use proper value

kujenga:
This resolves the typo outlined in the following issue: https://github.com/stripe/stripe-go/issues/355

It seems like this might be a thing to add test coverage for? I'm not familiar enough with the library to know the best way to approach that.

### DIFF
--- a/sub/client.go
+++ b/sub/client.go
@@ -66,7 +66,7 @@ func (c Client) New(params *stripe.SubParams) (*stripe.Sub, error) {
 		}
 
 		if params.TrialPeriod > 0 {
-			body.Add("trial_period_days", strconv.FormatInt(params.TrialEnd, 10))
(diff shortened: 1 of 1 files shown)


brandur:
Hi @kujenga, sincere apologies for the delayed response here! It looks like you're exactly right.

> It seems like this might be a thing to add test coverage for? I'm not familiar enough with the library to know the best way to approach that.

Yes, that would be ideal. Unfortunately our test suite isn't in the most optimal state right now for testing this kind of thing right now, so omitting a check for this might be our best course of action for the time being.

Thanks again!

brandur:
Released as 19.0.2.

kujenga:
Fair enough, thanks so much!

PR #359, opened 2017-01-18
Adds verify source endpoint

ob-stripe:
r? @will-stripe
cc @stripe/api-libraries @stan-stripe @remi-stripe 

This PR adds support for the `/v1/sources/src_.../verify` endpoint.

AFAIK, only `ach_debit` sources use this endpoint at the moment, so I hardcoded `Values` as `[2]uint8`. I guess that in the future, other source types might expect different types for the `values` parameter, and we might want to anticipate on that.

Putting the PR up for review, but I'll leave it to you to decide whether we should merge as-is or wait.

### DIFF
--- a/paymentsource.go
+++ b/paymentsource.go
@@ -33,8 +33,9 @@ type CustomerSourceParams struct {
 // For more details see https://stripe.com/docs/guides/ach-beta
 type SourceVerifyParams struct {
 	Params   `form:"*"`
-	Amounts  [2]uint8 `form:"amounts"`
-	Customer string   `form:"-"` // Goes in the URL
+	Amounts  [2]int64 `form:"amounts"` // Amounts is used when verifying bank accounts
+	Customer string   `form:"-"`       // Goes in the URL
+	Values   []string `form:"values"`  // Values is used when verifying sources
 }
 
 // SetSource adds valid sources to a CustomerSourceParams object,
--- a/paymentsource/client.go
+++ b/paymentsource/client.go
@@ -161,8 +161,10 @@ func (s Client) Verify(id string, params *stripe.SourceVerifyParams) (*stripe.Pa
 
 	if len(params.Customer) > 0 {
 		err = s.B.Call("POST", fmt.Sprintf("/customers/%v/sources/%v/verify", params.Customer, id), s.Key, body, &params.Params, source)
+	} else if len(params.Values) > 0 {
+		err = s.B.Call("POST", fmt.Sprintf("/sources/%v/verify", id), s.Key, body, &params.Params, source)
 	} else {
-		err = errors.New("Only customer bank accounts can be verified in this manner.")
+		err = errors.New("Only customer bank accounts or sources can be verified in this manner.")
 	}
 
 	return source, err
--- a/source.go
+++ b/source.go
@@ -35,6 +35,10 @@ const (
 type SourceFlow string
 
 const (
+	// FlowCodeVerification a verification code should be communicated by the
+	// customer to authenticate the source.
+	FlowCodeVerification SourceFlow = "code_verification"
+
 	// FlowNone no particular authentication is involved the source should
 	// become chargeable directly or asyncrhonously.
 	FlowNone SourceFlow = "none"
@@ -45,10 +49,6 @@ const (
 
 	// FlowRedirect a redirect is required to authenticate the source.
 	FlowRedirect SourceFlow = "redirect"
-
-	// FlowVerification a verification code should be communicated by the
-	// customer to authenticate the source.
-	FlowVerification SourceFlow = "verification"
 )
 
 // SourceUsage represents the possible usages of a source object.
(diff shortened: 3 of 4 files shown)


ob-stripe:
After discussing it with @remi-stripe, I updated the PR to use the existing `Amounts` parameter -- the library will send it to the API as `values` for sources.

This will avoid a breaking change if we eventually add a generic `Values` parameter in `SourceVerifyParams`.

will-stripe:
LGTM
Alternatively we can use an array of strings. If we are worried about breaking changes, I would suggest we go in that direction if we can, but I'm not too concerned.

source.go
```diff
@@ -152,8 +152,8 @@ const (
 
 // ReceiverFlow informs of the state of a verification authentication flow.
 type VerificationFlow struct {
-	AttemptsRemaining uint64             `json:"attempts_remaining"`
-	Status            RedirectFlowStatus `json:"status"`
+	AttemptsRemaining uint64                 `json:"attempts_remaining"`
+	Status            VerificationFlowStatus `json:"status"`
```
brandur: Was this just a mistake from before?

ob-stripe: That's my assumption, yes. I should have mentioned it in the PR, sorry about that!

remi-stripe:
@ob-stripe do you know if this still needs to be merged?

ob-stripe:
Yes, this still needs to be merged -- I forgot to remove the "[WIP]" tag in the PR's title.

Assigning to you @brandur-stripe

paymentsource/client.go
```diff
@@ -150,16 +150,20 @@ func Verify(id string, params *stripe.SourceVerifyParams) (*stripe.PaymentSource
...
 	var err error
 
 	if len(params.Customer) > 0 {
+		body.Add("amounts[]", strconv.Itoa(int(params.Amounts[0])))
+		body.Add("amounts[]", strconv.Itoa(int(params.Amounts[1])))
+
 		err = s.B.Call("POST", fmt.Sprintf("/customers/%v/sources/%v/verify", params.Customer, id), s.Key, body, &params.Params, source)
 	} else {
-		err = errors.New("Only customer bank accounts can be verified in this manner.")
+		body.Add("values[]", strconv.Itoa(int(params.Amounts[0])))
```
stan-stripe: So in its most general form, `code_verification` expects strings. We assume here that they are always numbers, which is OK for ACH debits (on bank account or on /v1/sources) but is not general enough for /v1/sources in general?

brandur-stripe: @stan-stripe You have lots of context on this, so would you mind pitching an alternative? Should we keep `Amounts` for customer-specific verification and add a `Values` here? Should we keep the two-element array?

stan-stripe: If you look at the signature of SourceVerifyParams[0], it was really built for bank account verification on customer attached bank account objects (despite the correct intent of making it a generic name)

The right type for the payment agnostic `/v1/sources/:id/verify` endpoint is:
```
type NewNameOrMaybeSameNameSourceVerifyParams struct {
  Values []string
}
```

Unclear to me if we should rename the old one, or rename the new one, but these are two different constructs in the API. Hope that helps?

[0] https://github.com/stripe/stripe-go/blob/8c52b2ccae8b855e33ce2e1339646f54ec791b59/payment_source.go#L37
[1] https://github.com/stripe-internal/pay-server/blob/master/api/lib/api_method/source_verify.rb#L13

brandur-stripe: > Unclear to me if we should rename the old one, or rename the new one, but these are two different constructs in the API. Hope that helps?

Yep, that clears things up. Thanks!

@ob-stripe I'd probably suggest introducing a new struct like the one Stan has above and renaming the old one. The breaking change sucks, but looking at Splunk, it doesn't seem to be too widely used from the Go language right now (< 10 calls in the last week).

ob-stripe:
I updated the PR to handle this with a non-breaking change: I added a `Values` strings array to the `SourceVerifyParams` struct but left the existing `Amounts` integer array.

If the `Customer` param is used, then we assume that the change is for a bank account and use the `Amounts` array. Otherwise we assume that it's for a source and use the `Values` array.

@brandur-stripe @stan-stripe wdyt?

payment_source.go
```diff
@@ -38,6 +38,7 @@ type SourceVerifyParams struct {
 	Params
 	Customer string
 	Amounts  [2]uint8
+	Values   []string
```
brandur-stripe: Would you mind commenting `Amounts` here to indicate that it's for bank accounts only and essentially deprecated.

ob-stripe: `ach_debit` sources are not live yet so I think it's a little bit early to flag bank accounts as deprecated.

That said, I can add the comment and we can wait to merge the PR -- it might be a good idea anyway as there might be other changes before release (e.g. the change from `verification` to `code_verification`).

ob-stripe:
> One thought: should we support both Values and Amount for bank account verification?

Personally, I'd say no: `/v1/customers/cus_.../sources/ba_.../verify` expects an `amounts` parameter, while `/v1/sources/src_.../verify` expects a `values` parameter, and I'd rather not have that sort of magic mapping from one parameter name to another.

payment_source.go
brandur-stripe: Cool. Yeah I was just thinking that if you come into this with very little information, both `Amounts` and `Values` are extremely generic names so it's really not intuitive at all that one would be meant for bank accounts and the other sources. It might be helpful to try and annotate this for users in some way.

brandur-stripe:
> Personally, I'd say no: /v1/customers/cus_.../sources/ba_.../verify expects an amounts parameter, while /v1/sources/src_.../verify expects a values parameter, and I'd rather not have that sort of magic mapping from one parameter name to another.

Fair enough. +1.

brandur-stripe:
@ob-stripe Can we proceed on this one?

remi-stripe:
@ob-stripe could you have a look at this one again rebased on the latest master?

ob-stripe:
Done.

@brandur-stripe ptal

paymentsource.go
```diff
@@ -33,8 +33,9 @@ type CustomerSourceParams struct {
 // For more details see https://stripe.com/docs/guides/ach-beta
 type SourceVerifyParams struct {
 	Params   `form:"*"`
-	Amounts  [2]uint8 `form:"amounts"`
-	Customer string   `form:"-"` // Goes in the URL
+	Amounts  [2]uint8 `form:"amounts"` // Amounts is used when verifying bank accounts
+	Customer string   `form:"-"`       // Goes in the URL
+	Values   []string `form:"values"`  // Values in used when verifying sources
```
remi-stripe: minor typo here

paymentsource/client.go
```diff
@@ -161,8 +161,8 @@ func (s Client) Verify(id string, params *stripe.SourceVerifyParams) (*stripe.Pa
 
 	if len(params.Customer) > 0 {
 		err = s.B.Call("POST", fmt.Sprintf("/customers/%v/sources/%v/verify", params.Customer, id), s.Key, body, &params.Params, source)
-	} else {
-		err = errors.New("Only customer bank accounts can be verified in this manner.")
+	} else if len(params.Values) > 0 {
+		err = s.B.Call("POST", fmt.Sprintf("/sources/%v/verify", id), s.Key, body, &params.Params, source)
 	}
```
remi-stripe: you're likely missing an else that raises an error no?

paymentsource.go
ob-stripe: 👍

paymentsource/client.go
ob-stripe: 👍  Added an `else` clause. The error message is a bit vague, I'm open to suggestions if you can think of something better!

brandur-stripe:
> AFAIK, only ach_debit sources use this endpoint at the moment, so I hardcoded Values as [2]uint8. I guess that in the future, other source types might expect different types for the values parameter, and we might want to anticipate on that.

Thanks for picking this back up again!

@ob-stripe @remi-stripe One another kinda nit: I _think_ one of the things Remi's doing with his big refactoring project right now is to just make all our integers the same type (not sure which one, but probably either `int` or `int64`. Do you think we could replace the `uint8` with either one of these? It's not like you're going to be saving that many system resources by using a smaller integer size on verification (it's an infrequent operation and our modern 64-bit platforms are probably often storing it in 64 bits anyway) so just trying to fall into a convention here is probably more valuable.

Maybe `int64` would be a good choice. We're never going to be storing numbers that are big enough that we'll need `uint64` over `int64`, and it's kind of nice that `int64` is unambiguous 64-bit, although honestly `int` is probably fine too.

ob-stripe:
Rebased on master and updated to use `int64` for `Amounts`.

brandur-stripe:
Released as 28.3.0.

PR #360, opened 2017-01-20
Remove subscriptions on customer

jack-stripe:
r? @brandur-stripe

(not to be released until after API version 2017-01-18; this is why tests are failing)

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -53,7 +53,6 @@ type Customer struct {
 	Discount      *Discount                `json:"discount"`
 	Email         string                   `json:"email"`
 	Meta          map[string]string        `json:"metadata"`
-	Subs          *SubList                 `json:"subscriptions"`
 	Deleted       bool                     `json:"deleted"`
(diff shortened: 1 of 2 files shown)


brandur:
Cool thanks! Similar question to on Java: I thought that you were expecting these changes to happen a little further out in the future. Are they already out as of `2017-01-18`?

Also, since we need to make a major version release for this kind of stuff, is it possible that we could pool a few more of the prospective changes together so that we that we can make as few as possible?

jack-stripe:
@brandur-stripe: good point; I was planning on releasing said customer/subscription changes a couple weeks ago, but then went on vacation instead. I'll update the API version in this PR to the correct thing after the new version's released.

Re: the major version thing - sounds fine to me.

brandur:
@jack-stripe Sounds good! Because (I think) the API change is still a little ways out, I'm going to close this pull for the time being for general repository hygiene. Feel free to reopen as appropriate!

PR #361, opened 2017-01-23
Added support for outcome[rule] on Charge object

remi-stripe:
r? @brandur-stripe

I haven't been able to add tests for this because in Test mode that hash seems to always been empty.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -100,14 +100,21 @@ type FraudDetails struct {
 	StripeReport FraudReport `json:"stripe_report"`
 }
 
+// ChargeOutcomeRule tells you the Radar rule that blocked the charge, if any.
+type ChargeOutcomeRule struct {
+	Action    string `json:"action"`
+	Predicate string `json:"predicate"`
+}
+
 // Outcome is the charge's outcome that details whether a payment
 // was accepted and why.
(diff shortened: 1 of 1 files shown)


brandur:
> I haven't been able to add tests for this because in Test mode that hash seems to always been empty.

I think the Go bindings use a different test account, so it may be some sort of gating issue. Honestly, I wouldn't worry about it too much.

LGTM, but would you mind adding a short comment to the new struct of the form "ChargeOutcomeRule is ...". We're still a little ways off it, but it would be nice to eventually turn the Linter on for this project.

remi-stripe:
> I think the Go bindings use a different test account, so it may be some sort of gating issue. Honestly, I wouldn't worry about it too much.

I think that it's mainly available in Live mode only specifically.

> would you mind adding a short comment to the new struct of the form "ChargeOutcomeRule is ...". 

will do!

remi-stripe:
PTAL

brandur:
> I think that it's mainly available in Live mode only specifically.

Ah yes, that sounds plausible.

Looks great. Thanks!

brandur:
Released as 19.2.0.

PR #364, opened 2017-02-03
Add support for account debits

jnovak-stripe:
Add support for account debits. In reality it already worked and was close to fine as-is -- the only problem I found was jsonifying the response. This addresses that.

cc @stripe/api-libraries r? @brandur-stripe

### DIFF
--- a/payment_source.go
+++ b/payment_source.go
@@ -96,6 +96,10 @@ const (
 	// PaymentSourceCard is a constant representing a payment source which is a
 	// card.
 	PaymentSourceCard PaymentSourceType = "card"
+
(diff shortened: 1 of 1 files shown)


jnovak-stripe:
Also, to be clear, the JSON coming back from 'source' here is just `{id: "acct_xxx", object: "account"}` (only those 2 fields). It's not the full account representation.

brandur:
> Also, to be clear, the JSON coming back from 'source' here is just {id: "acct_xxx", object: "account"} (only those 2 fields). It's not the full account representation.

Thanks for the clarification!

LGTM.

brandur:
Released as 19.4.0.

PR #365, opened 2017-02-07
support destination hash

ccontinanza-stripe:
Don't merge right away.

Changes the calling convention for creating destination charges from specifying a destination string to using a struct that supports `destination[amount]` and `destination[account]` to match the new, public API: https://stripe.com/docs/api#create_charge-destination

Making a Charge with the proposed struct looks like this:

```go
    Amount:        1000,
    Currency:      "usd",
    TransferGroup: group,
    Dest: stripe.DestinationParams{
      Amount:  500,
      Account: accountId,
    },
  }
```

Thoughts:  I called it `DestinationParams` and went with a `string`/`uint64` because that matched some of the conventions: I considered just calling it `Destination` and using an `*Account` but we only appear to do that on the serialization side (which I think makes sense).

Lemme know if you'd like me to:
1) rename anything
2) still support the string via `Dest` and call the field on the `Charge` struct something like `Destination` so you can opt-in to the new syntax (versus getting, albeit easily fixed, compiler error).
3) pass in a reference to the struct instead of a value

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -16,17 +16,17 @@ type FraudReport string
 // For more details see https://stripe.com/docs/api#create_charge and https://stripe.com/docs/api#update_charge.
 type ChargeParams struct {
 	Params
-	Amount                 uint64
-	Currency               Currency
-	Customer, Token        string
-	Desc, Statement, Email string
-	Dest                   DestinationParams
-	NoCapture              bool
-	Fee                    uint64
-	Fraud                  FraudReport
-	Source                 *SourceParams
-	Shipping               *ShippingDetails
-	TransferGroup          string
+	Amount                       uint64
+	Currency                     Currency
+	Customer, Token              string
+	Desc, Statement, Email, Dest string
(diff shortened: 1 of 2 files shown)


ccontinanza-stripe:
cc @stripe/api-libraries r? @brandur-stripe

brandur:
Thanks!

> still support the string via Dest and call the field on the Charge struct something like Destination so you can opt-in to the new syntax (versus getting, albeit easily fixed, compiler error).

Yeah, lets do this. It's good to clean up old code, but we should try not to make breaking changes too liberally, especially when they're not really needed. +1 `Destination` naming too; that's probably the better name anyway.

> pass in a reference to the struct instead of a value

Given that the other two structs on params are pointers, yeah, lets change this to one too.

ptal @ccontinanza-stripe

ccontinanza-stripe:
made requested changes -> PTAL

r? @brandur-stripe

brandur-stripe:
LGTM. Thanks!

PR #366, opened 2017-02-12
Start testing Go 1.7 and latest

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries

Start testing Go 1.7 and latest (with failures allowed for the latter).

Also start testing OS X and Linux separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
 language: go
+sudo: false
 go:
   - 1.6
-sudo: false
+  - 1.7
+  - tip
+matrix:
+  allow_failures:
+    - go: tip
+  fast_finish: true


brandur-stripe:
Nice! And we can add 1.8 in about three days too.

I'm a little bearish on adding Mac into the build matrix. This seems like a good idea conceptually, but Travis often has a hugely backed up queue for Mac boxes which means that your builds start to take _forever_. I'd rather just trust that Go is guaranteeing us our cross-platform compatibility.

ob-stripe:
:+1: I'll update the PR.

brandur-stripe:
Sweet. Thanks!

PR #371, opened 2017-03-09
Add subscription status all as a constant

erictse:
From what I can tell, `all` is a valid status value. https://stripe.com/docs/api/go#list_subscriptions-status

Does this seem reasonable?


brandur-stripe:
Yep, this looks perfectly reasonable. Thank you!

brandur-stripe:
Released as 19.9.0.

Issue #370, opened 2017-03-07
Impossible to remove coupon from subscription via API


layby42:
Hi,
 
Looks like it is impossible to remove coupon from subscription via api.

According to the doc for subscription https://stripe.com/docs/api/go#update_subscription it should be possible: "This can be unset by updating the value to null and then saving."

Currently the code (https://github.com/stripe/stripe-go/blob/master/sub/client.go) does: 	

`
if len(params.Coupon) > 0 {
	body.Add("coupon", params.Coupon)
}
`

Maybe it is possible to extend `SubParams` with `CouponRemove` (or something like this) and handle it similar to `TaxPercentZero` in `func Update(id string, params *stripe.SubParams) (*stripe.Sub, error)`:

`
if len(params.Coupon) > 0 {
     body.Add("coupon", params.Coupon)
} else if params.CouponRemove {
     body.Add("coupon", "")
}
`
?

remi-stripe:
@layby42 This is a pretty recent change and we didn't push the change to the Go bindings yet. We'll be shipping it in the future but if you're up for a pull request that would be awesome!

layby42:
Hi Everybody,

Looks like it is possible to remove coupon from subscription other way around, from discount: https://stripe.com/docs/api/go#delete_subscription_discount. 

I will try to use it.

remi-stripe:
@layby42  yes that's the way that we had until a few weeks ago. We recently added the ability to set `coupon: null` when updating a subscription via the API too but this isn't part of the Go bindings yet.

layby42:
@remi-stripe, OK, I will submit  pull request

brandur-stripe:
Closing this out as it was fixed in #372.

PR #372, opened 2017-03-09
Extended Sub Update with CouponRemove

layby42:
pull request for https://github.com/stripe/stripe-go/issues/370

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -11,6 +11,7 @@ type CustomerParams struct {
 	Balance        int64
 	BalanceZero    bool
 	Token, Coupon  string
+	CouponEmpty    bool
 	Source         *SourceParams
 	Desc, Email    string
 	Plan           string
--- a/customer/client.go
+++ b/customer/client.go
@@ -134,6 +134,8 @@ func (c Client) Update(id string, params *stripe.CustomerParams) (*stripe.Custom
 
 		if len(params.Coupon) > 0 {
 			body.Add("coupon", params.Coupon)
+		} else if params.CouponEmpty {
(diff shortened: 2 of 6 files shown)


brandur-stripe:
Hi @layby42, thank you so much for the PR!

Sorry for the nitpick, but would you mind changing the naming to `CouponEmpty`? I think that way we could then more plausibly apply it to other parameters in the future where we want to allow a string to be set to an empty value. I'm actually a little surprised that we don't seem to have a convention for this already. Thanks!

layby42:
Hi @brandur-stripe, renamed

remi-stripe:
@layby42 we just released a change in the API allowing to do the same thing on the Customer Update [API](https://stripe.com/docs/api#update_customer-coupon). Any chance you could add this feature in your PR too?

layby42:
@brandur-stripe we do not need it for customer, at least not now, but we do need it for subscription and we need it now. I can add the same for customer, I guess.

remi-stripe:
@layby42 if you could bundle it in your PR that would be awesome as it's the same feature just on a different model. No worries if you don't though I can add it later otherwise.

layby42:
@remi-stripe done

remi-stripe:
LGTM

I'll let @brandur-stripe do the merge + release early next week though. Thank you so much for working on this so quickly!

brandur-stripe:
Nice! Thanks for making those changes @layby42! Bringing this in.

brandur-stripe:
Released as 19.10.0.

PR #373, opened 2017-03-11
Add Token to SourceObjectParams

antonholmquist:
Token was missing on SourceObjectParams. Also added it to request body when creating new source.

### DIFF
--- a/source/client.go
+++ b/source/client.go
@@ -81,7 +81,7 @@ func (c Client) New(params *stripe.SourceObjectParams) (*stripe.Source, error) {
 		}
 
 		if params.Token != "" {
-			body.Add("token", token)
+			body.Add("token", params.Token)
 		}
 
 		for k, v := range params.TypeData {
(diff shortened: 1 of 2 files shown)


source/client.go
```diff
@@ -80,6 +80,10 @@ func (c Client) New(params *stripe.SourceObjectParams) (*stripe.Source, error) {
 			}
 		}
 
+		if params.Token != "" {
+			body.Add("token", token)
+		}
```
brandur-stripe: @antonholmquist Thanks for the contribution!

You've got a bit of a syntax problem here. I think this should be `params.Token`. Otherwise this looks good to me. Thanks!

antonholmquist: Too fast, thank you!

brandur-stripe:
Looks good. Thanks!

brandur-stripe:
Released as 19.11.0.

antonholmquist:
Great. Really impressed with the quality of this lib and Stripe stuff in general!

brandur-stripe:
@antonholmquist Very glad to hear it! Thank you for using the service :)

Issue #374, opened 2017-03-15
File Upload requires a file on disk


Oralordos:
Currently the `FileUploadParams` requires that the file you are going to upload is of type `*os.File`. Looking through the code, all that you actually want is the filename, and an `io.Reader`.

If the file is being uploaded by a user to your server, for a managed account for example, there is no file on disk. If you are on AppEngine especially, where you have no write access to the disk, you cannot send the user's identity document using this library.

To keep old code working, a transition period could be put in where the file is supported, as well as the reader and filename codepath, similar to how the `Account` and `StripeAccount` fields are being used now.

In short, `FileUploadParams` should be updated to take a filename and a reader instead of a file.

PR #375, opened 2017-03-15
File upload no longer requires the file to be on disk (#374)

Oralordos:
Added support for files to be uploaded from any generic reader so long as a filename can be provided. Also added a test for the new parameters.

The old parameters still work for backwards compatibility, although there is a comment in the code marking them as deprecated.

### DIFF
--- a/fileupload.go
+++ b/fileupload.go
@@ -15,13 +15,13 @@ type FileUploadParams struct {
 	Params
 	Purpose FileUploadPurpose
 
-	// FileData is a reader with the contents of the file that should be uploaded.
+	// FileReader is a reader with the contents of the file that should be uploaded.
+	FileReader io.Reader
 	// Filename is just the name of the file without path information.
-	FileData io.Reader
 	Filename string
 
-	// File is a deprecated form of FileData and Filename that will do the same thing.
-	// Please use FileData and Filename instead.
+	// File is a deprecated form of FileReader and Filename that will do the same thing, but
+	// allows referencing a file directly. Please prefer the use of FileReader and Filename instead.
 	File *os.File
 }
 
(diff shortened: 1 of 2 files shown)


Oralordos:
Oh, should probably mention this fixes #374, which I made just a bit ago.

fileupload.go
```diff
@@ -14,7 +14,15 @@ import (
 type FileUploadParams struct {
 	Params
 	Purpose FileUploadPurpose
-	File    *os.File
+
+	// FileData is a reader with the contents of the file that should be uploaded.
+	// Filename is just the name of the file without path information.
```
brandur-stripe: Thanks for the documentation! would you mind just moving the `Filename` doc string down a line so that it's directly over the field that it's documenting?

fileupload.go
```diff
@@ -14,7 +14,15 @@ import (
...
+	FileData io.Reader
+	Filename string
+
+	// File is a deprecated form of FileData and Filename that will do the same thing.
+	// Please use FileData and Filename instead.
```
brandur-stripe: Good call documenting the old stuff too. Thanks! Can you tweak this a little bit to read:

> File is a deprecated form of FileData and Filename that will do the same thing, but allows referencing a file directly. Please prefer the use FileData and Filename instead.

fileupload.go
```diff
@@ -59,7 +67,19 @@ func (f *FileUploadParams) AppendDetails(body io.ReadWriter) (string, error) {
...
+		part, err := writer.CreateFormFile("file", filepath.Base(f.Filename))
+		if err != nil {
+			return "", err
+		}
+
+		_, err = io.Copy(part, f.FileData)
+		if err != nil {
+			return "", err
+		}
+		// But allow old value of File field for now.
```
brandur-stripe: I tend to prefer to document my conditions like this too, but gofmt doesn't play nicely with them. Would you mind moving this comment to be combined in with line 70 just because it looks pretty awkward indented into this conditional. Could read something like:

> Support both FileData/Filename and File with the former being the newer preferred version

fileupload.go
```diff
@@ -14,7 +14,15 @@ import (
 type FileUploadParams struct {
 	Params
 	Purpose FileUploadPurpose
-	File    *os.File
+
+	// FileData is a reader with the contents of the file that should be uploaded.
+	// Filename is just the name of the file without path information.
+	FileData io.Reader
```
brandur-stripe: Should we call this `FileReader`?

fileupload/client_test.go
```diff
@@ -56,6 +56,46 @@ func TestFileUploadNewThenGet(t *testing.T) {
 	}
 }
 
+func TestInterfaceFileUploadNewThenGet(t *testing.T) {
```
brandur-stripe: Could you rename this to `TestFileUploadNewThenGet` and then rename the current test to be `TestFileUploadNewThenGetWithFile`? I think it'll help convey that the former is the newer preferred option.

Oralordos:
All changes put in.

brandur-stripe:
Thanks!

brandur-stripe:
Released as 19.12.0.

PR #381, opened 2017-03-25
Added support for the new Payout object

remi-stripe:
The Transfer object used to represent all movements of funds in Stripe. It split in two concepts:
- Transfer: this describes the movement of funds between Stripe accounts and is specific to Stripe Connect.
- Payout: this describes the movement of funds from a Stripe account to a bank account, debit card or any future payout method.

This applies beyond the new object as some properties/methods are removed from Transfer and other properties are renamed on other objects.

We also stopped supporting Recipient creation starting at that API version which means we removed the `New` method and had to fix related tests to not rely on Recipients.

This change is behind an API version. Since the Go library forces an API version it will impact anyone upgrading and they will need to change their code.


remi-stripe:
@brandur-stripe I have some open questions about this PR that I'm not sure how to address/answer. Sorry for the long list.

1. Should we take this as an opportunity to rename all properties aways from short names? Like Tx => BalanceTransaction and Dest => Destination
2. Should we order all properties/parameters alphabetically once and for all?
3. Should we have only one property/parameter per line?
4. Why do you need `params.AddMeta()` instead of passing the object directly in the params on definition?
5. Is it worth splitting the new tests in sub-functions (means creating charges/accounts numerous times)
6. Should we add the new format for `destination[amount]` and `destination[account]` at the same time?
7. What to do with TransferDestination which is now only Account? Do we rip it out or is it useful for expanding?
8. I didn't port Recipient to Payouts. Do you think it's needed if we want to list old payouts' destination/type?
9. Should we remove all tests around recipients beyond the Get? Some tests are failing because of this.

brandur-stripe:
Wow Remi, this is obviously an amazing amount of work. Thanks a lot for tackling this! (And providing context in the PR description — surprisingly rare ;)

Some answers below:

> Should we take this as an opportunity to rename all properties aways from short names? Like Tx => BalanceTransaction and Dest => Destination

Given that this seems pretty breaking change, I'd be +1 this, but it might be better to keep it more localized to properties that are related to transactions and payouts. This allows the scope of the churn to at least be somewhat localized. Thoughts?

> Should we order all properties/parameters alphabetically once and for all?

I think we should do this, but lets save it for another PR. It's not breaking or anything, so one huge alphabetization PR would be perfectly adequate.

> Should we have only one property/parameter per line?

Yes! I hate this arbitrary "sometimes it is and sometimes it's not" thing. Again though, I can see saving this for a future PR given that it doesn't affect the interface.

> Why do you need params.AddMeta() instead of passing the object directly in the params on definition?

So keep in mind I'm not the authoritative source for the original design, but I think the reason here is that `AddMeta` will conveniently initialize the metadata property if it isn't already. This allows you to just add in some new metadata without thinking about whether or not some is already set on the object.

See these lines:
https://github.com/stripe/stripe-go/blob/master/params.go#L192,L194

> Is it worth splitting the new tests in sub-functions (means creating charges/accounts numerous times)

So yes, probably. I think the testing strategy in this library is generally completely pathological and would benefit from some kind of consistency across tests and some kind of modularity within them. i.e. Instead of just testing the entire world in one test case, test _one thing_ or at least a set of closely related things.

I think the reason they're written as is is performance: the tests hit the live API so when setting up objects, so you can get some additional performance gain by just reusing some fixtures that you've already built through an expensive RT that already occurred.

I don't have strong opinions on what we should do immediately because honestly, I think the only way to really move forward is to dump the whole suite and start afresh with some kind of fast local stubbing system.

> Should we add the new format for destination[amount] and destination[account] at the same time?

Sorry, I read through the brief, but would mind elaborating on this one a little bit? I haven't written any code and I don't think I understand all the ramifications here.

> What to do with TransferDestination which is now only Account? Do we rip it out or is it useful for expanding?

This is a bit of a guess, but I think it might be useful to rip it out. There's enough breaking changes in this area here anyway, and this no longer seems like a useful abstraction. Thoughts?

> I didn't port Recipient to Payouts. Do you think it's needed if we want to list old payouts' destination/type?

Hmmm .... I honestly don't know enough about how people use the product to say for sure. It seems like it might be okay to leave it out and add it if the feature is requested?

> Should we remove all tests around recipients beyond the Get? Some tests are failing because of this.

Given that these tests are only nominally useful in that they don't run in CI anyway, I think that'd be fine.

balance.go
```diff
@@ -151,6 +159,10 @@ func (s *TransactionSource) UnmarshalJSON(data []byte) error {
 			json.Unmarshal(data, &s.Dispute)
 		case TransactionSourceFee:
 			json.Unmarshal(data, &s.Fee)
+		case TransactionSourceRecipientTransfer:
+			json.Unmarshal(data, &s.RecipientTransfer)
+		case TransactionSourcePayout:
+			json.Unmarshal(data, &s.Payout)
```
brandur-stripe: Really minor, but mind just keeping this list alphabetized?

remi-stripe: Fixed! I apparently don't know my alphabet because I did check when I added that :p

brandur-stripe:
@remi-stripe Rebased and bumped the pinned version to `2017-04-06`.

recipient/client.go
```diff
@@ -18,49 +18,8 @@ type Client struct {
...
-		body.Add("description", params.Desc)
-	}
-	params.AppendTo(body)
-
-	recipient := &stripe.Recipient{}
-	err := c.B.Call("POST", "/recipients", c.Key, body, &params.Params, recipient)
-
-	return recipient, err
-}
+// Since API version 2017-04-06, new recipients can't be created anymore.
```
brandur-stripe: Thanks for the comment :)

brandur-stripe:
LGTM.

brandur-stripe:
Released as 20.0.0.

payout.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return nil
+}
+
+// UnmarshalJSON handles deserialization of a PayoutDestination.
+// This custom unmarshaling is needed because the specific
+// type of destination it refers to is specified in the JSON
+func (d *PayoutDestination) UnmarshalJSON(data []byte) error {
+	type dest PayoutDestination
+	var dd dest
```
ferhatelmas: Why don't we define it directly as `var dd PayoutDestination` ?

payout.go
```diff
@@ -0,0 +1,156 @@
...
+			json.Unmarshal(data, &d.BankAccount)
+		case PayoutDestinationCard:
+			json.Unmarshal(data, &d.Card)
+		}
+	} else {
+		// the id is surrounded by "\" characters, so strip them
+		d.ID = string(data[1 : len(data)-1])
+	}
+
+	return nil
```
ferhatelmas: Why don't we return `err` here or handle error in `json.Unmarshal` in line 138 or 140 ?

ferhatelmas:
@remi-stripe @brandur-stripe for my own curiosity and learning, I asked two questions, it would be very appreciated if you can answer. Thanks!

payout.go
brandur-stripe: @ferhatelmas Sorry for the late reply here. Somehow these notifications got lost in my big backlog from the holidays.

So the reason that we use this type definition trick is to avoid accidental recursion. We're defining `UnmarshalJSON` for `PayoutDestination`, but we're _also_ calling `UnmarshalJSON` inside the definition! Normally this would just recurse right back into this same function and produce a stack overflow. Here we alias `PayoutDestination` to a new type (`dest`) and call `UnmarshalJSON` on that instead, which falls back to the default `UnmarshalJSON` implementation for a struct. Make sense?

payout.go
brandur-stripe: Good question. So you may notice that a few lines above, `data` is already run through another `json.Unmarshal`:

```
	err := json.Unmarshal(data, &dd)
```

Given that it's already been validated there (i.e. it's likely to be valid JSON, etc.), it's fairly unlikely that those other `Unmarshal` invocations are going to throw an error, and I think that's why an error check on them was originally excluded (it turns out that this is/was a common convention in the library and exists throughout a number of files).

That said, even though it's unlikely to be needed it's still good practice to include an error check anyway. I've added a patch in #509 that puts error checks in all these places.

PR #383, opened 2017-03-29
Stripe Balance Transaction Is a String

mikepigott:
The Stripe Go API has the Charge's Balance Transaction as a JSON object, but in reality it's the String ID of that JSON object.


mikepigott:
I didn't want to merge directly to Stripe's master.

remi-stripe:
@mikepigott We implemented this way because the transaction could have been [expanded](https://stripe.com/docs/api#expanding_objects). This means that the resource can be a string if not expanded or the entire resource if it is. Our Go bindings handle this gracefully by expanding or not as needed and you'd get your id in `charge.Tx.ID`

PR #386, opened 2017-04-06
Upgrade the pinned API version to 2017-02-14

brandur-stripe:
Upgrades the pinned API version to [2017-02-14](https://stripe.com/docs/upgrades#2017-02-14) and introduces appropriate changes. This mostly includes adding an `UnmarshalJSON` implementation for dispute and charge outcome's rule, both of which have become fields that are expandable but collapsed by default.

cc @remi-stripe

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -112,6 +112,7 @@ type FraudDetails struct {
 // ChargeOutcomeRule tells you the Radar rule that blocked the charge, if any.
 type ChargeOutcomeRule struct {
 	Action    string `json:"action"`
+	ID        string `json:"id"`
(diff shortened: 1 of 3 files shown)


brandur-stripe:
This basically takes us right up to the point where payouts/transfers are split. The plan is to merge this, then bump the API version one more time in #381.

r? @remi-stripe

remi-stripe:
LGTM

brandur-stripe:
FWIW, I range through the tests in `./charge` and `./dispute` manually and they're passing along with CI.

Thanks!

PR #387, opened 2017-04-07
Balance history can be filtered by Payout and not Transfer

remi-stripe:
This fixes something I missed when doing the Payout/Transfer split. When listing all balance transactions bundled in a payout you need to pass the payout id, po_XXXXX, in the `payout` parameter instead of `transfer.

r? @brandur-stripe


brandur-stripe:
Nice catch @remi-stripe. Thanks!

brandur-stripe:
Released as 21.0.0.

PR #388, opened 2017-04-08
Remove raw PANs from our test suite

remi-stripe:
This PR aims to use the new test tokens such as tok_visa instead or hardcoding raw PANs in the test suite.

There's one last PAN we can't remove for disputes as it hasn't been released yet as a token ready-to-use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,12 +50,7 @@ params := &stripe.CustomerParams{
 	Desc:  "Stripe Developer",
 	Email: "gostripe@stripe.com",
 }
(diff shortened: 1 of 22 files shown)


remi-stripe:
Similar to the PHP one, the test environments run in parallel and try to re-use the token which leads to errors. Waiting for the internal fix.

remi-stripe:
@brandur-stripe  PTAL!

brandur-stripe:
Thanks!

PR #389, opened 2017-04-12
Coupon allowed on order creation and optional description

remi-stripe:
Adds support for the `coupon` parameter on order [creation](https://stripe.com/docs/api#create_order-coupon). Also only send the description for an Order Item if it's set. Today, if you don't set it you get an error even though it's [optional](https://stripe.com/docs/api#create_order-items-description).

### DIFF
--- a/order/client_test.go
+++ b/order/client_test.go
@@ -144,6 +144,8 @@ func TestOrder(t *testing.T) {
 	if o.Meta["foo"] != "bar" {
 		t.Error("Order metadata not set")
 	}
+
+	coupon.Del(c.ID)
 }
 
 func TestOrderUpdate(t *testing.T) {
--- a/order/client_test.go
+++ b/order/client_test.go
@@ -59,7 +59,16 @@ func CreateTestProductAndSku(t *testing.T) *stripe.SKU {
 func TestOrder(t *testing.T) {
 	sku := CreateTestProductAndSku(t)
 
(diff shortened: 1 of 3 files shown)


order/client_test.go
```diff
@@ -59,7 +59,16 @@ func CreateTestProductAndSku(t *testing.T) *stripe.SKU {
 func TestOrder(t *testing.T) {
 	sku := CreateTestProductAndSku(t)
 
+	couponParams := &stripe.CouponParams{
+		Amount:   99,
+		Currency: currency.USD,
+		Duration: coupon.Once,
+	}
+
+	c, err := coupon.New(couponParams)
```
brandur-stripe: Minor, but currently the Go tests make at least a half-hearted attempt to clean up after themselves when they run. Maybe add this to the end of the test case:

``` go
coupon.Del(c.ID)
```

You can see another example of this in the [order update test case here](https://github.com/stripe/stripe-go/blob/2aea243be0e339e6a5687c6fa0b9675339221139/order/client_test.go#L226).

brandur-stripe:
Minor comment with regards to testing, but looks good to me. Thanks Remi!

order/client_test.go
remi-stripe: Fixed! Nice catch!

brandur-stripe:
LGTM.

brandur-stripe:
Released as 21.1.0.

PR #390, opened 2017-04-18
Remove user_id field from event object.

leonsim:
According to https://stripe.com/docs/api/go#event_object , there is not a field called `user_id`


ob-stripe:
Hi @leonsim, thanks for your contribution. The `user_id` field actually does exist: it is used when events are sent to [Connect webhook endpoints](https://stripe.com/docs/connect/authentication#webhooks) and contains the ID of the account that emitted the event.

I'm going to close the PR, but I'll make sure to send a note to our docs team to see if it'd be possible to document the field in the API reference.

PR #391, opened 2017-04-18
Encode and send `X-Stripe-Client-User-Agent`

brandur-stripe:
It turns out that the go bindings were not sending our JSON `X-Stripe-Client-User-Agent` like the other bindings. This PR fixes that to put it inline with everything else.

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -10,6 +10,8 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
+	"runtime"
 	"strings"
 	"time"
 )
@@ -33,6 +35,10 @@ const defaultHTTPTimeout = 80 * time.Second
 // TotalBackends is the total number of Stripe API endpoints supported by the
 // binding.
 const TotalBackends = 2
+
+// UnknownPlatform is the string returned as the system name if we couldn't get
+// one from `uname`.
+const UnknownPlatform = "unknown platform"
 
 // Backend is an interface for making calls against a Stripe service.
 // This interface exists to enable mocking for during testing if needed.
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -348,3 +382,24 @@ func (s *BackendConfiguration) ResponseToError(res *http.Response, resBody []byt
...
+	var out bytes.Buffer
+	cmd.Stderr = nil // goes to os.DevNull
+	cmd.Stdout = &out
+	err = cmd.Run()
+	if err != nil {
+		return UnknownPlatform
+	}
+
+	return out.String()
+}
```
brandur-stripe: So one note here: the way we're trying to fetch unames across libraries isn't consistent. Where the language has a built-in function (PHP, Python) we usually use that. stripe-ruby does something a little special in that it tries extra hard to get a uname based on a whole bunch of different switches. I didn't really want to reimplement that, so I'm just using the stripe-node technique of trying to run `uname -a`.

I think this should generally be suitable because it would be pretty surprising for people to be running server side Go on non-Unix non-Mac OS systems (I'm sure it's done somewhere, but it's probably very unusual).

stripe_test.go
```diff
@@ -38,6 +40,66 @@ func TestCheckinBackendConfigurationNewRequestWithStripeAccount(t *testing.T) {
 	}
 }
 
+func TestCheckinUserAgent(t *testing.T) {
```
brandur-stripe: I also added a basic test for `User-Agent` because I want to use it on my next PR for `application` support.

stripe.go
ob-stripe: Seems very reasonable

ob-stripe:
👍

brandur-stripe:
Thanks @ob-stripe!

PR #392, opened 2017-04-18
Add app info to stripe-go

brandur-stripe:
Adds app info the Go libraries (the idea is already in PHP and Ruby).

r? @ob-stripe 
cc @stripe/api-libraries 

---

~~Note: currently targets the branch in #391 until that's merged.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,22 @@ if err := i.Err(); err != nil {
 }
 ```
 
+### Writing a Plugin
+
+If you're writing a plugin that uses the library, we'd appreciate it if you
+identified using `stripe.SetAppInfo`:
+
+```go
+stripe.SetAppInfo(&stripe.AppInfo{
+    Name:    "MyAwesomePlugin",
+    URL:     "https://myawesomeplugin.info",
+    Version: "1.2.34",
+})
+```
+
+This information is passed along when the library makes calls to the Stripe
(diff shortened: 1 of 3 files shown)


ob-stripe:
My Go-fu is very weak, but lgtm!

OOC, when instantiating an `AppInfo` struct, all three fields must be provided, but users can provide empty strings as values for `URL` and `Version`, right? If that's the case, do you think it might be worth clarifying this in the README?

brandur-stripe:
Thanks @ob-stripe!

> OOC, when instantiating an AppInfo struct, all three fields must be provided, but users can provide empty strings as values for URL and Version, right? If that's the case, do you think it might be worth clarifying this in the README?

So that's basically true, but actually the user doesn't have to specify all three fields. They're allowed to do this:

```go
	appInfo := &stripe.AppInfo{
		Name: "MyAwesomePlugin",
	}
```

Which is why I need to make empty string checks while encoding. In the case where these values aren't passed, `URL` and `Value` will end up as the string type's "zero value", which is an empty string.

ob-stripe:
Makes perfect sense, thanks for explaining!

brandur-stripe:
NP!

Released this and the `X-Stripe-Client-User-Agent` change as 21.2.0.

PR #396, opened 2017-05-02
Move STRIPE_KEY into code

brandur-stripe:
Right now `STRIPE_KEY` gets pulled from an encrypted Travis variable
which has a few disadvantages:

* It's really opaque. I have to go to Splunk to reverse engineer which
  merchant it belongs to every time I need to look at it (the Travis secret
  is not reversible, even to us).

* It means there's no easy local testbed. Everyone who tries to run the
  Go test suite needs to go find their own merchant to use. This is
  painful.

Storing keys in plaintext like this is of course a little weird, but
we're already doing it in pretty much every other library, and it's a
testmode key on an account that's test-only.

r? @ob-stripe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,20 @@
+# This is a testmode key for this test account:
+#
+#     ID:    cuD9Rwx8pgmRZRpVe02lsuR9cwp2Bzf7
+#     Email: test+bindings@stripe.com
+#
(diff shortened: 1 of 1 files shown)


ob-stripe:
Seems totally peaceful -- the secret key for the bindings test account is not so secret :)

:+1:

brandur-stripe:
Thanks @ob-stripe!

Going to pull this in and delete the `STRIPE_KEY` in our Travis dashboard then.

PR #400, opened 2017-05-11
Pass extra parameters on bank account deletion

remi-stripe:
This ensures that we send extra parameters such as the Account on bank account deletion on a customer saved on a connected account.

r? @brandur-stripe


brandur-stripe:
Doh. LGTM.

brandur-stripe:
Released as 21.4.1.

PR #402, opened 2017-05-13
Add 'Get' method to RequestValues

### LINKED ISSUE #401: Add query method on RequestValues
I'm writing a `Backend` object so i can fake the responses from Stripe in a test.  It's a little annoying that the only accessor method for `RequestValues` is `Encode`...would you be amenable to me adding a `Get(key string) []string` method?

edenman:
fixes #401 
@brandur-stripe I wasn't able to run the tests so I'm not 100% sure this works, but it seems right :)

the error I get when I do a `make test`:
```
./params_test.go:38: values.Get undefined (type *stripe.RequestValues has no field or method Get)
[a few more of the same]
```
presumably because I'm building against the released version of stripe-go, and not the modified one that has my new method?  Sorry if this is an obvious thing, I'm a Go n00b.

### DIFF
--- a/params.go
+++ b/params.go
@@ -62,11 +62,11 @@ func (f *RequestValues) Set(key, val string) {
 }
 
 // Get retrieves the list of values for the given key.  If no values exist
-// for the key, an empty list will be returned.
+// for the key, nil will be returned.
 //
 // Note that Get is O(n) and may be quite slow for a very large parameter list.
 func (f *RequestValues) Get(key string) []string {
-	results := []string{}
+	var results []string
 	for i, v := range f.values {
 		if v.Key == key {
 			results = append(results, f.values[i].Value)
--- a/params.go
+++ b/params.go
@@ -61,6 +61,20 @@ func (f *RequestValues) Set(key, val string) {
 	f.Add(key, val)
 }
 
+// Get retrieves the list of values for the given key.  If no values exist
+// for the key, an empty list will be returned.
+//
+// Note that Get is O(n) and may be quite slow for a very large parameter list.
+func (f *RequestValues) Get(key string) []string {
+	results := []string{}
+	for i, v := range f.values {
+		if v.Key == key {
+			results = append(results, f.values[i].Value)
+		}
+	}
+	return results
+}
+
 // ToValues converts an instance of RequestValues into an instance of
 // url.Values. This can be useful in cases where it's useful to make an
 // unordered comparison of two sets of request values.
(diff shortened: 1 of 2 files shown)


params.go
```diff
@@ -61,6 +61,20 @@ func (f *RequestValues) Set(key, val string) {
 	f.Add(key, val)
 }
 
+// Get retrieves the list of values for the given key.  If no values exist
+// for the key, an empty list will be returned.
+//
+// Note that Get is O(n) and may be quite slow for a very large parameter list.
+func (f *RequestValues) Get(key string) []string {
+	results := []string{}
```
brandur-stripe: Fairly minor, but can we return `nil` instead of an empty list when `Get` doesn't find anything?

You should be able to do this by just changing this line to:

``` go
var results []string
```

`append` will automatically take care of list initialization for you.

(Also, the doc string will need an update too. Thanks!)

brandur-stripe:
@edenman Thanks for the pull!

> presumably because I'm building against the released version of stripe-go, and not the modified one that has my new method? Sorry if this is an obvious thing, I'm a Go n00b.

Yeah, it looks like you're building against your installed version of stripe-go. The usual reason for this is that unlike most languages, Go is very picky about where code you're working on is stored, and although it kind of works without one, they heavily recommend setting up [a Go workspace](https://golang.org/doc/code.html#Workspaces) which (defaults to `~/go` in newer versions). You're basically trying to make sure that your code is located at `$GOPATH/src/github.com/stripe/stripe-go`.

You probably have that path already, so what you want to do is set up a new Git remote there that points back to your own fork of the repository, and push to that. Make sense?

brandur-stripe:
Left a minor comment above, but otherwise looks good to me. We have a very weird CI setup on this project still, so it'd be nice to double check that those tests are passing locally before bringing this in.

edenman:
Ok so I got the tests to run (had a to fix a silly compile issue) but a bunch of them fail for apparently unrelated reasons.  If I just run my test:
```
[Erics-MBP]~/Go/src/github.com/stripe/stripe-go master= go test -run "TestRequestValues" .
ok  	github.com/stripe/stripe-go	0.012s
```

brandur-stripe:
> Ok so I got the tests to run (had a to fix a silly compile issue) but a bunch of them fail for apparently unrelated reasons. If I just run my test:

Yeah, that's fine. Most of the suite doesn't run in CI, and that's had exactly the kind of dilapidation effect that you might expect. Thanks for the changes!

brandur-stripe:
Released as 21.5.0.

edenman:
thanks @brandur-stripe!

PR #404, opened 2017-05-23
Use interface for logger

discordianfish:
Hi,

I want to propose this simple change which allows people to specify their own logger. All calls to the logger are already only using Printf, so that's all a logger is required to implement.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -84,7 +84,12 @@ var LogLevel = 2
 // Logger controls how stripe performs logging at a package level. It is useful
 // to customise if you need it prefixed for your application to meet other
 // requirements
-var Logger *log.Logger
+var Logger Printfer
+
+// Printfer is an interface to be implemented by Logger.
+type Printfer interface {
+	Printf(format string, v ...interface{})
+}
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
> I want to propose this simple change which allows people to specify their own logger.

Can you elaborate on this further? It seems like you can already specify your own logger.

I'm somewhat amenable to the change if there's a good reason, but by default we should probably stick with standard Go conventions here.

discordianfish:
@brandur-stripe Right now you can only specify an instance of log.Logger, while there are many other, very common logger implementations (like logrus which I'm using). I don't feel very strongly since I could also use a log.Logger instance which writes to my 3rd party logger, but using interfaces seems to me the more idiomatic way. Then again, not sure how others are doing this.

brandur-stripe:
Thanks for the context! I poked around with a few of our Go people here, and it seems that this is fairly uncontroversial given a lack of a better solution from the standard library. Thanks for the patch.

brandur-stripe:
Released as 22.1.0.

PR #406, opened 2017-05-24
Rename `user_id` to `account`

brandur-stripe:
This field is being renamed on a new API version for tomorrow (and we need to bump the version used by stripe-go before bringing it in).

Given that this is mostly a change in how we map our JSON serialization, I haven't changed any tests for this one.

r? @remi-stripe


remi-stripe:
LGTM

PR #408, opened 2017-05-25
Add support for the new account type

remi-stripe:
When creating an account via the API, you should now pass the `type` parameter explicitly instead of `managed`.

r? @brandur


brandur-stripe:
@remi-stripe Same on this one! (gofmt)

remi-stripe:
@brandur PTAL

PR #409, opened 2017-05-25
Added support for login_link on the Account resource

remi-stripe:
r? @brandur


brandur-stripe:
@remi-stripe Can you check the build here too? I think you need to run gofmt. I'll look into getting the API version bumped.

remi-stripe:
@brandur I think the last build failure is just bad luck and unrelated.

PR #410, opened 2017-05-25
Make event's request an object

brandur-stripe:
We made a change in the latest API version that makes an event's request an object that contains an idempotency key along with a request ID. Here we reflect that in the Go code.

I also opportunistically renamed `Req` to `Request` to kill another one of these silly abbreviations since this is a breaking change anyway.

r? @remi-stripe


remi-stripe:
LGTM

This is awesome!!!!!

PR #412, opened 2017-05-28
Fix account type comment copy paste errors

jbowes:


brandur-stripe:
Thank you for the fix!

PR #413, opened 2017-05-28
Implement webhook.ConstructEvent.

### LINKED ISSUE #405: Webhook signature verification
Hi, I think it would be useful if this library verified webhook event signatures.  Here's how I implemented it:

```go
import (
	"crypto/hmac"
	"crypto/sha256"
	"encoding/hex"
	"errors"
	"strings"
)

// checkSignature verifies a Stripe-Signature header from webhook.
// It returns an error if the signature is missing or invalid, or nil if the signature is OK.
// NOTE: stripe only sends signatures when you have clicked to reveal your secret
// The secret is a hex byte slice of the string generated by stripe.
// The body is the webhook request's body.
// The header is the webhook request's Stripe-Signature header.
func checkSignature(secret, body []byte, header string) error {
        // First extract the timestamp
	var timestamp string
	pairs := strings.Split(header, ",")
	for _, pair := range pairs {
		parts := strings.Split(pair, "=")
		if parts[0] == "t" {
			timestamp = parts[1]
			break
		}
	}

	if timestamp == "" {
		return errors.New("Missing timestamp")
	}

	mac := hmac.New(sha256.New, secret)
	mac.Write([]byte(timestamp))
	mac.Write([]byte("."))
	mac.Write(body)
	res := mac.Sum(nil)

	// Check all v1 signatures (multiple v1 can happen in case of rolled secret).
	for _, pair := range pairs {
		parts := strings.Split(pair, "=")
		if parts[0] != "v1" {
			continue
		}

		sig, err := hex.DecodeString(parts[1])
		if err != nil {
			return err
		}

		if hmac.Equal(res, sig) {
			return nil
		}
	}

	return errors.New("Correct signature not found")
}
```

Usage looks like:

```go
func stripeWebhooks(w http.ResponseWriter, r *http.Request) {
        secret := byte[]("whsec_XYZ")
	body, _ := ioutil.ReadAll(r.Body)
        header := r.Header.Get("Stripe-Signature")
        if err := checkSignature(secret, body, header); err != nil {
                http.Error(w, "Missing/invalid signature", http.StatusBadRequest)
                return
        }
```

Happy to make a PR if this looks OK and if you can guide on how API should look :).

froatsnook:
Fixes #405.

webhook.ConstructEvent creates a stripe.Event and verifies the signature and timestamp.  Since go doesn't have default parameters, the `tolerance` param is required.  I added two constants `ToleranceDefault` with the suggested 5 minute tolerance, and `ToleranceIgnoreTimestamp` which disables checking the timestamp.

### DIFF
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -42,10 +42,12 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 	pairs := strings.Split(sigHeader, ",")
 	for _, pair := range pairs {
 		parts := strings.Split(pair, "=")
-		if parts[0] == "t" {
-			t = parts[1]
-			break
+		if len(parts) != 2 || parts[0] != "t" {
+			continue
 		}
+
+		t = parts[1]
+		break
 	}
 
 	if t == "" {
@@ -80,7 +82,7 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 	// Check all given v1 signatures since multiple v1 can happen in case of rolled secret.
 	for _, pair := range pairs {
 		parts := strings.Split(pair, "=")
-		if parts[0] != "v1" {
+		if len(parts) != 2 || parts[0] != "v1" {
 			continue
 		}
 
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -17,6 +17,11 @@ import (
 const ToleranceDefault int64 = 300
 const ToleranceIgnoreTimestamp int64 = 0
 
+var ErrNotSigned error = errors.New("Webhook has no Stripe-Signature header")
+var ErrNoTimestamp error = errors.New("Webhook has no timestamp")
+var ErrTooOld error = errors.New("Webhook had valid signature but timestamp wasn't within tolerance")
+var ErrNoValidSignature error = errors.New("Webhook had no valid signature")
+
 // ConstructEvent initializes an Event object from a JSON payload.
 // It returns an non-nil error when the payload is not valid JSON or when signature verification fails.
 // payload is the webhook request body, i.e. `ioutil.ReadAll(r.Body)`.
@@ -32,7 +37,7 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 	}
 
 	if sigHeader == "" {
-		return e, errors.New("Missing Stripe-Signature header")
+		return e, ErrNotSigned
 	}
 
 	// sigHeader looks like "t=1495999758,v1=ABC,v1=DEF,v0=GHI"
@@ -51,24 +56,25 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 	}
 
 	if t == "" {
-		return e, errors.New("Webhook has no timestamp")
+		return e, ErrNoTimestamp
 	}
 
+	invalidTimestamp := false
 	if tolerance > 0 {
 		timestamp, err := strconv.ParseInt(t, 10, 64)
-		if err != nil {
-			return e, errors.New("Webhook has invalid timestamp")
-		}
+		if err == nil {
+			currentTimestamp := time.Now().Unix()
 
-		currentTimestamp := time.Now().Unix()
+			diff := timestamp - currentTimestamp
+			if diff < 0 {
+				diff = -diff
+			}
 
-		diff := timestamp - currentTimestamp
-		if diff < 0 {
-			diff = -diff
-		}
-
-		if diff > tolerance {
-			return e, errors.New("Webhook is too old")
+			if diff > tolerance {
+				invalidTimestamp = true
+			}
+		} else {
+			invalidTimestamp = true
 		}
 	}
 
@@ -88,14 +94,19 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 
 		sig, err := hex.DecodeString(parts[1])
 		if err != nil {
-			return e, fmt.Errorf("Invalid v1 signature: %s", err.Error())
+			// Ignore signature which isn't valid hex.
+			continue
 		}
 
 		if hmac.Equal(res, sig) {
+			if invalidTimestamp {
+				return e, ErrTooOld
+			}
+
 			// OK
 			return e, nil
 		}
 	}
 
-	return e, errors.New("Correct signature not found")
+	return e, ErrNoValidSignature
 }
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -1,0 +1,99 @@
+package webhook
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/stripe/stripe-go"
+)
+
+const ToleranceDefault int64 = 300
+const ToleranceIgnoreTimestamp int64 = 0
+
+// ConstructEvent initializes an Event object from a JSON payload.
+// It returns an non-nil error when the payload is not valid JSON or when signature verification fails.
+// payload is the webhook request body, i.e. `ioutil.ReadAll(r.Body)`.
+// sigHeader is the webhook Stripe-Signature header, i.e. `r.Header.Get("Stripe-Signature")`.
+// secret is your Signing Secret, i.e. `"whsec_XYZ"`.  See https://dashboard.stripe.com/webhooks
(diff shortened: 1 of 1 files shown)


webhook/client.go
```diff
@@ -0,0 +1,99 @@
...
+		return e, errors.New("Missing Stripe-Signature header")
+	}
+
+	// sigHeader looks like "t=1495999758,v1=ABC,v1=DEF,v0=GHI"
+
+	// First extract the timestamp
+	var t string
+	pairs := strings.Split(sigHeader, ",")
+	for _, pair := range pairs {
+		parts := strings.Split(pair, "=")
```
ob-stripe: In order to make sure that you're really returning pairs, you should use [`SplitN`](https://golang.org/pkg/strings/#SplitN) with `n=2` here.

webhook/client.go
```diff
@@ -0,0 +1,99 @@
...
+		if parts[0] != "v1" {
+			continue
+		}
+
+		sig, err := hex.DecodeString(parts[1])
+		if err != nil {
+			return e, fmt.Errorf("Invalid v1 signature: %s", err.Error())
+		}
+
+		if hmac.Equal(res, sig) {
```
ob-stripe: You should use [`ConstantTimeCompare`](https://golang.org/pkg/crypto/subtle/#ConstantTimeCompare) from the `crypto/subtle` package to compare signatures. This protects against timing attacks that could be used to retrieve the secret and forge valid signatures.

webhook/client.go
```diff
@@ -0,0 +1,99 @@
...
+	mac := hmac.New(sha256.New, []byte(secret))
+	mac.Write([]byte(t))
+	mac.Write([]byte("."))
+	mac.Write(payload)
+	res := mac.Sum(nil)
+
+	// Check all given v1 signatures since multiple v1 can happen in case of rolled secret.
+	for _, pair := range pairs {
+		parts := strings.Split(pair, "=")
+		if parts[0] != "v1" {
```
ob-stripe: It'd be better to have `"v1"` set as a constant somewhere, to make it easier to upgrade the library if/when we need to bump the signature scheme.

webhook/client.go
```diff
@@ -0,0 +1,99 @@
...
+
+		currentTimestamp := time.Now().Unix()
+
+		diff := timestamp - currentTimestamp
+		if diff < 0 {
+			diff = -diff
+		}
+
+		if diff > tolerance {
+			return e, errors.New("Webhook is too old")
```
ob-stripe: Ideally, we should first test for the presence of a valid signature before checking the timestamp. The timestamp can only be trusted after a valid signature was found (since timestamps are part of the signed payload).

ob-stripe:
Hey @froatsnook, thank you very much for tackling this!

I have left a few comments and I'm no Go expert, but this looks like a very good first step to me.

We'd need some tests to go along with this too. There are basically 4 possible scenarios in which the signature verification can fail:
- the `Signature-Header` is unparseable
- the header is parseable, but there are no signatures with the expected scheme (currently `"v1"`)
- there are signatures with the expected scheme, but none of them match the expected signature for the received payload
- a valid signature was found, but the timestamp for the event is too old

For reference, here are the tests in Ruby: https://github.com/stripe/stripe-ruby/blob/master/test/stripe/webhook_test.rb. All other languages have a similar set of tests.

Do you think you could add these tests? Otherwise, we'll handle this internally.

cc @brandur-stripe

webhook/client.go
froatsnook: [`hmac.Equal`](https://golang.org/pkg/crypto/hmac/#Equal) is a constant-time compare

> Equal compares two MACs for equality without leaking timing information.

webhook/client.go
froatsnook: That doesn't really make sense to me since you won't bump the version for fun -- you would only change the version if the hash or something changed, and this code would have to be rewritten anyway.

webhook/client.go
froatsnook: Sounds good, will do.

webhook/client.go
froatsnook: So you're saying you would rather have one error message than another?  Why would that matter?

froatsnook:
> Hey @froatsnook, thank you very much for tackling this!

You're welcome :).  Thanks for providing go bindings for stripe.

> Do you think you could add these tests? Otherwise, we'll handle this internally.

I'll take a look and get back to you.

webhook/client.go
ob-stripe: We need to ensure that there is at least one valid signature in the header AND that the timestamp is within the tolerance. So strictly speaking, you are correct that the order in which the condition are verified does not matter: if one of the conditions isn't met, then the signature verification fails.

Functionally speaking though, I believe that "no valid signatures" is more important than "timestamp not within the tolerance" and so should be checked first. When that's the case, the "timestamp not within the tolerance" error becomes a strong indicator of a replay attack (because a valid signature was found).

It's also how it was implemented in all of our official libraries and I'd much rather have the behavior be as consistent as possible across all languages.

webhook/client.go
ob-stripe: The way it was implemented in other libraries is that if we ever need to change the signature scheme, only two things would need be updated:

- the constant containing the value of the expected scheme (e.g. `"v1"` -> `"v2"`)

- the `compute_signature` method that computes the expected signature with the appropriate scheme (e.g. we'd switch from HMAC-SHA256 to another algorithm)

So, yeah, I should have mentioned that too. It'd be better to have a dedicated method that takes a payload and secret and computes the expected signature. Having such a method will also make testing the entire thing easier since you can reuse the method to compute signatures in your tests to craft `Stripe-Signature` header values.

webhook/client.go
ob-stripe: My bad, sorry!

webhook/client.go
```diff
@@ -0,0 +1,99 @@
...
+	// Check all given v1 signatures since multiple v1 can happen in case of rolled secret.
+	for _, pair := range pairs {
+		parts := strings.Split(pair, "=")
+		if parts[0] != "v1" {
+			continue
+		}
+
+		sig, err := hex.DecodeString(parts[1])
+		if err != nil {
+			return e, fmt.Errorf("Invalid v1 signature: %s", err.Error())
```
ob-stripe: Upon re-reading this, wouldn't an error be raised if an invalid signature is found?

The way the signature verification needs to be implemented is that the verification succeeds as long as one valid signature key is found, even if there are additional invalid signatures.

This is useful when rolling an endpoint's signing secret: you can leave the old secret active for a period of time, and during this time Stripe will generate one signature per active secret, so the header would have two different signatures, each with a `v1=` prefix.

This would let you update your code to replace the old secret with the new with no downtime.

webhook/client.go
froatsnook: That makes sense.

webhook/client.go
froatsnook: OK, I will change this

webhook/client.go
froatsnook: "Invalid" signatures are fine so long as they are properly formatted.  This error only happens if there's a v1 signature that can't be parsed as a hex string.  So a rolling secret should work fine.

OTOH might as well ignore an invalid v1 and then return an error when no valid signature is found.

webhook/client.go
froatsnook: Actually this doesn't help.  SplitN returns **at most** `n` substrings. So rather we need to check if len(pairs) is 2.

ob-stripe: So this is what we want: an `a=b=c` entry should be decoded as a pair (prefix) `a` / (value) `b=c`.

That said, you're correct that we should also verify that there _are_ two elements and not just one. An entry without any `=` character would be equally invalid and the code would fail when trying to access `parts[1]`.

webhook/client.go
ob-stripe: Ah yes, sorry, I completely misread the condition here.

I agree that non-hex values for signatures should simply be ignored and fall back to the general "no valid signature found" error case if needed.

froatsnook:
Hi, I've made your requested changes.  Would you like me to squash?

Can you guys implement the tests internally?  I'm sorry to say I don't have the time to implement them.  Trying to go live ASAP.

ob-stripe:
@froatsnook Sure thing. I'm terrible at Go myself and only pitching in because I'm familiar with webhook signatures, but I'll find someone to complete the review and add tests. Thanks for your contributions!

cc @brandur-stripe @remi-stripe for awareness and in case one of you wants to volunteer :)

PR #414, opened 2017-06-02
Added support to delete a plan on a connected account

mrosentr:
Modified plan/client.go Del functions to accept params.
Modified plan.go struct with Customer String.

### DIFF
--- a/plan/client.go
+++ b/plan/client.go
@@ -130,8 +130,9 @@ func (c Client) Del(id string, params *stripe.PlanParams) (*stripe.Plan, error) 
 	}
 
 	plan := &stripe.Plan{}
+	qid := url.QueryEscape(id) //Added query escape per commit 9821176
 
-	err := c.B.Call("DELETE", fmt.Sprintf("/plans/%v", id), c.Key, body, commonParams, plan)
+	err := c.B.Call("DELETE", fmt.Sprintf("/plans/%v", qid), c.Key, body, commonParams, plan)
 
 	return plan, err
 }
--- a/plan/client.go
+++ b/plan/client.go
@@ -2,6 +2,7 @@
 package plan
 
 import (
+	"fmt"
 	"net/url"
 	"strconv"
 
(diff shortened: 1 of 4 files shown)


brandur-stripe:
@mrosentr Thanks for the pull!

Can you clarify what the new `Customer` parameter is for? I don't think this is going to do anything right now.

mrosentr:
@brandur-stripe Correct, it isn't doing anything right now.  I added it in to be consistent with the api call from the platform based on other functions and also future proofing if needed later on.

brandur-stripe:
> @brandur-stripe Correct, it isn't doing anything right now. I added it in to be consistent with the api call from the platform based on other functions and also future proofing if needed later on.

Okay, I'm not sure I fully follow this. There aren't any plan endpoints that take `customer` right now as far as I'm aware.

You know that usage for Connect would be something more akin to `params.SetAccount()` right?

mrosentr:
Your right.  With no endpoints it doesn't make any sense.  I have removed it.  Thank you.

plan/client.go
```diff
@@ -113,13 +114,24 @@ func (c Client) Update(id string, params *stripe.PlanParams) (*stripe.Plan, erro
...
+		body = &stripe.RequestValues{}
+
+		params.AppendTo(body)
+		commonParams = &params.Params
+	}
+
 	plan := &stripe.Plan{}
-	err := c.B.Call("DELETE", "/plans/"+url.QueryEscape(id), c.Key, nil, nil, plan)
+
+	err := c.B.Call("DELETE", fmt.Sprintf("/plans/%v", id), c.Key, body, commonParams, plan)
```
brandur-stripe: Can we make sure that we keep the `QueryEscape` around here? It looks like it came in for a reason in 98211760beb526dbc58cd91b408488c0ce6929ed.

brandur-stripe:
@mrosentr Thanks! Can you check the test suite run in CI? It looks like there are two more test failures:

```
client/checkin_test.go:126: not enough arguments in call to c.Plans.Del
client/checkin_test.go:171: not enough arguments in call to c.Plans.Del
```

mrosentr:
@brahn-stripe I have corrected the last two test failures.

brahn-stripe:
cc: @brandur-stripe

brandur-stripe:
Thanks!

brandur-stripe:
Released as 23.0.0.

PR #415, opened 2017-06-08
Add webhook signing

### LINKED ISSUE #405: Webhook signature verification

fred-stripe:
Fixes #405 based on @froatsnook's excellent work in #413. Added a test suite based on other libraries' tests, refactored the public interface, and updated the godocs.

r? @brandur-stripe @ob-stripe 

I want to add one additional test as I think it's not fully testing multiple signatures behavior.

### DIFF
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -127,14 +127,14 @@ func ValidateEventIgnoringTolerance(payload []byte, header string, secret string
 	return validateEvent(payload, header, secret, 0*time.Second, false)
 }
 
-func validateEvent(payload []byte, header string, secret string, tolerance time.Duration, enforceTolerance bool) (stripe.Event, error) {
+func validateEvent(payload []byte, sigHeader string, secret string, tolerance time.Duration, enforceTolerance bool) (stripe.Event, error) {
 	e := stripe.Event{}
 
 	if err := json.Unmarshal(payload, &e); err != nil {
 		return e, fmt.Errorf("Failed to parse webhook body json: %s", err.Error())
 	}
 
-	header, err := parseSignatureHeader(header)
+	header, err := parseSignatureHeader(sigHeader)
 	if err != nil {
 		return e, err
 	}
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -42,10 +42,12 @@ func ConstructEvent(payload []byte, sigHeader string, secret string, tolerance i
 	pairs := strings.Split(sigHeader, ",")
 	for _, pair := range pairs {
 		parts := strings.Split(pair, "=")
-		if parts[0] == "t" {
-			t = parts[1]
-			break
+		if len(parts) != 2 || parts[0] != "t" {
+			continue
 		}
+
+		t = parts[1]
+		break
 	}
 
 	if t == "" {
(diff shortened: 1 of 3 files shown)


ob-stripe:
Awesome! Thanks @fred-stripe (and thanks again @froatsnook).

- I don't feel too strongly about this but I think the `testmode` part based on detecting `v0` signatures is unnecessary. The event itself already has a `livemode` attribute, and while it's true that `v0` signatures are only sent for test mode events, their purpose is not specifically to indicate test mode but rather to allow you to verify that your integration ignores signatures with unexpected schemes

-  could we rename the methods from `ValidateEvent` to `ConstructEvent`, to be consistent with other languages implementations?

Aside from that, lgtm!

brandur-stripe:
Nice work! Looks good to me as well.

Quick question: is `ConstructEventIgnoringTolerance` expected to be useful for something? It seems like it might be something that could be used accidentally, and whether we should consider just not even providing it.

fred-stripe:
Hmm, the only way this would be useful would be validating stale signed requests long after the fact (say you stored all webhook requests in a file or database somewhere). I'm not sure that it's a really valid usecase, as the better approach would be to note that they had valid signatures when they arrived and reject those that aren't helpful.

The ability to ignore timestamp tolerances seems to be available in `stripe-ruby` as well, so I'm tempted to keep it in this implementation:

https://github.com/stripe/stripe-ruby/blob/master/lib/stripe/webhook.rb#L69-L73

I also rather fancy `ConstructEventIgnoringTolerance` as a method name, similar to `dangerouslySetInnerHTML` in React—that is, "you _can_ do this but you should think about what you're doing."

brandur-stripe:
Alright, works for me!

brandur-stripe:
Released as 22.5.0.

PR #417, opened 2017-06-08
Add support for ephemeral keys

grey-stripe:
This adds the `EphemeralKey` resource to the bindings.  Currently, the only supported operations are creation and deletion.

Creation is a bit of a special case, in that it requires `Stripe-Version` to be set explicitly (because this will be requested by the backend for use in the frontend, which may be expecting an arbitrary API version).  This means that it is an error if the header is set by the globally configured API version, and it must be provided explicitly in the `create` call.  As such, `create` will return an error if `StripeVersion` is not specified.

In addition, the Go bindings do not typically handle different API version formats, so there is some different behavior in `UnmarshalJSON` for the `EphemeralKey` resource.  We preserve the raw JSON in `EphemeralKey.RawJSON` so that it may be passed to the frontend unchanged, even if the format is not supported by the bindings.


ephemeralkey.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+// EphemeralKey is the resource representing a Stripe ephemeral key.
+// For more details see https://stripe.com/docs/api#ephemeral_keys.
+type EphemeralKey struct {
+	ID      string `json:"id"`
+	Created int64  `json:"created"`
+	Expires int64  `json:"expires"`
+	Live    bool   `json:"livemode"`
+    AssociatedObjects []struct {
```
brandur-stripe: You may not have been told, but the Go people _really_ like tabs :)

You should probably change this in your editor (there's some good Go plugins for Vim if you're on that), but the toolchain also ships with a tool called `gofmt` to fix this stuff for you:

```
gofmt -w *.go
```

We have a check in CI on this, so you'll need to fix it before the build will pass.

ephemeralkey.go
```diff
@@ -0,0 +1,42 @@
...
+        ID   string `json:"id"`
+        Type string `json:"type"`
+    } `json:"associated_objects"`
+    RawJSON []byte
+}
+
+// UnmarshalJSON handles deserialization of an EphemeralKey.
+// This custom unmarshaling is needed because the resulting
+// property may be an id or the full struct if it was expanded.
+func (e *EphemeralKey) UnmarshalJSON(data []byte) error {
```
brandur-stripe: So these custom deserializers are generally only needed when we have properties that may be expandable. Looking at what you have here, I think you can get away without one.

ephemeralkey/client.go
```diff
@@ -0,0 +1,53 @@
...
+	return getC().New(params)
+}
+
+func (c Client) New(params *stripe.EphemeralKeyParams) (*stripe.EphemeralKey, error) {
+	if params.StripeVersion == "" {
+		return nil, fmt.Errorf("params.StripeVersion must be specified")
+	}
+
+	body := &stripe.RequestValues{}
+    body.Add("customer", params.Customer)
```
brandur-stripe: So admittedly our convention here is far from perfect, but would you mind changing this to be nil checked?

``` go
	if len(params.Customer) > 0 {
		body.Add("customer", params.Customer)
	}
```

ephemeralkey/client_test.go
```diff
@@ -0,0 +1,94 @@
...
+    })
+    if err != nil {
+        t.Fatalf("%+v", err)
+    }
+
+    k, err = ephemeralkey.Del(k.ID)
+    if err != nil {
+        t.Fatalf("%+v", err)
+    }
+}
```
brandur-stripe: I think I already ranted to you about our bad CI situation with these bindings in particular, but could you just confirm that these tests are passing for you locally?

params.go
```diff
@@ -124,6 +124,11 @@ type Params struct {
 	// Account is deprecated form of StripeAccount that will do the same thing.
 	// Please use StripeAccount instead.
 	Account string
+
+    // StripeVersion sets the version for the API to use when parsing requests
+    // and rendering responses.  This string refers to a date, such as
+    // "2017-05-25"
+    StripeVersion string
```
brandur-stripe: Ugh, so this is going to make implementation a little more awkward, but we shouldn't add a top-level `StripeVersion` here.

stripe-go ties itself to a particular version of the Stripe API, and this is probably something that we want to eventually emulate from every library. Even if that doesn't happen though, we should make sure that this continues to be the case for stripe-go because otherwise this makes versioning available on every endpoint and makes people's use of the library much less deterministic for us.

I think we'll have to move this down to the `EphemeralKeyParams` and figure out another way to inject it into the request headers.

ephemeralkey/client_test.go
grey-stripe: They were passing when I checked them in - I'll rerun them after fixing the review comments.

ephemeralkey.go
grey-stripe: So the top comment was copy/paste incorrect, but I think we still need this because of the `RawJSON` field (which is needed to pass back an unparseable EphemeralKey to the frontend - which could happen due to incompatible api versions).

params.go
grey-stripe: OK, I added a general `Headers` field here instead, and `StripeVersion` is now on `EphemeralKeyParams`.

params.go
```diff
@@ -124,6 +124,10 @@ type Params struct {
 	// Account is deprecated form of StripeAccount that will do the same thing.
 	// Please use StripeAccount instead.
 	Account string
+
+	// This may be used to provide a new header or override an existing one on
+	// the HTTP request.
+	Headers map[string]string
```
brandur-stripe: A couple fair minor comments here:

1. One of the core values of the Go language is that the standard library's types/functions/conventions should be re-used as often as possible. Instead of defining your own map here, can you make this a `Header` type [from `net/http`](https://godoc.org/net/http#Header)? It's basically just an alias for what you have already, but allows for each key to hold more than one value, which is more correct than what you have here because HTTP allows you to repeat headers multiple times.
2. `Header` also exports an `Add` function, so lets drop your `AddHeader` implementation below, and just prefer the use of that.
3. Convention for Go docstrings is to start out with the name of the thing that they're documenting (e.g. `// Headers may be used to provide new headers or override existing ones on HTTP requests made from stripe-go.`). We don't have a linter turned on for the project the project right now (because there's so many existing problems), but golint would've called this out as a violation. Can you reword accordingly?

Thanks!

brandur-stripe:
@grey-stripe Requested a few other minor changes here, but starting to look very mergeable.

params.go
grey-stripe: Great comments!
1) Switched to `net/http.Header`
2) `http.Header` doesn't automatically initialize on `Add`, so I added a check in `ephemeralkey.New` to initialize it.
3) `golint` actually didn't seem to catch this (probably because it's not a top-level export), but fixed anyway (and a few other things `golint` turned up)

brandur-stripe: Thanks!

> golint actually didn't seem to catch this (probably because it's not a top-level export), but fixed anyway (and a few other things golint turned up)

Weird! It should have. Either way though, I probably should [have linked to this document on it instead](https://blog.golang.org/godoc-documenting-go-code):

> Notice this comment is a complete sentence that begins with the name of the element it describes. This important convention allows us to generate documentation in a variety of formats, from plain text to HTML to UNIX man pages, and makes it read better when tools truncate it for brevity, such as when they extract the first line or sentence.

Excellent though. This is good to go I think.

ephemeralkey.go
```diff
@@ -0,0 +1,42 @@
...
+type EphemeralKey struct {
+	ID                string `json:"id"`
+	Created           int64  `json:"created"`
+	Expires           int64  `json:"expires"`
+	Live              bool   `json:"livemode"`
+	AssociatedObjects []struct {
+		ID   string `json:"id"`
+		Type string `json:"type"`
+	} `json:"associated_objects"`
+	RawJSON []byte
```
brandur-stripe: Actually, one other ask here: would you mind documenting the `RawJSON` type field here? I don't think it's intuitive as to why it's needed, and I'm a little afraid that it might confuse people as they're reading the code (like as to why this model has it, but no others do).

PR #418, opened 2017-06-12
Handling subscription's item deletion when updating a subscription

alecha:
I’ve noticed that the `Deleted` field of an `SubItem` is not taken into account while updating a subscription


sub/client.go
```diff
@@ -167,6 +167,9 @@ func (c Client) Update(id string, params *stripe.SubParams) (*stripe.Sub, error)
 				} else if item.QuantityZero {
 					body.Add(key+"[quantity]", "0")
 				}
+				if item.Deleted {
+					body.Add(key+"[deleted]", "true")
+				}
```
alecha: not sure if the syntax is this one, please have a look 👍

brandur-stripe:
Thanks @alecha! Released as 22.1.1.

PR #419, opened 2017-06-13
Support setting the sub's quantity to 0 for an upcoming invoice

remi-stripe:
When previewing an upcoming invoice, you can also simulate changes to the subscription. Changing the plan for example works. Setting the `quantity` to 0 was not working because this parameter is ignored in the code.

This adds `SubQuantityZero` as a valid parameter, mimicking what we do for the subscription update in the rest of the library.

r? @brandur


brandur-stripe:
Good call. Thanks Remi!

brandur-stripe:
Released as 22.2.0.

PR #421, opened 2017-06-14
Subitem List Fix

nathanjsweet:
1. Update Subitem List client to use the SubItemListParams
2. Pass subscription to the request params directly from the SubItemListParams rather than the Filters.


brandur-stripe:
Thank you!

Definitely copy pasta at its worst. I'm going to release this as a minor version bump even though it's technically breaking since no one could have been using this successfully anyway.

brandur-stripe:
Released as 22.3.0.

nathanjsweet:
No problem. Happy to help.

PR #424, opened 2017-06-22
Start test suite refactor + power tests with external API stub

brandur:
In the style of https://github.com/stripe/stripe-ruby/pull/553 starts to move toward an external API stub [1] executable to power the test suite. This will allow us to start turning more tests back on in CI (currently the vast majority of them don't run).

I've also left some inline notes that annotate how I've also changed our testing strategy in a few places. The end goal are tests that are _much_ faster, _much_ more reliable, and _not_ totally broken (because they don't get to run in CI).

This is an experimental prototype for demonstrative purposes and should not be merged.

[1] https://github.com/stripe/stripe-mock

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,26 +13,25 @@ sudo: false
 
 env:
   global:
-    - STRIPE_STUB_PORT=6065
-    - STRIPE_STUB_VERSION=0.1.1
+    - STRIPE_MOCK_VERSION=0.1.13
 
 cache:
   directories:
-    - stripestubs
+    - stripe-mock
 
 before_install:
   - go get -u github.com/stretchr/testify/require
 
   # Unpack and start the Stripe API stub so that the test suite can talk to it
   - |
-    if [ ! -d "stripestubs/stripestub_${STRIPE_STUB_VERSION}" ]; then
-      mkdir -p stripestubs/stripestub_${STRIPE_STUB_VERSION}/
-      curl -L "https://github.com/brandur/stripestub/releases/download/v${STRIPE_STUB_VERSION}/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz" -o "stripestubs/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz"
-      tar -zxf "stripestubs/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz" -C "stripestubs/stripestub_${STRIPE_STUB_VERSION}/"
+    if [ ! -d "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}" ]; then
+      mkdir -p stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/
+      curl -L "https://github.com/brandur/stripe-mock/releases/download/v${STRIPE_MOCK_VERSION}/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz" -o "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz"
+      tar -zxf "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz" -C "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/"
     fi
   - |
-    stripestubs/stripestub_${STRIPE_STUB_VERSION}/stripestub -port ${STRIPE_STUB_PORT} &
-    STRIPE_STUB_PID=$!
+    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock > /dev/null &
+    STRIPE_MOCK_PID=$!
 
 script:
   - make
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,30 @@ matrix:
   allow_failures:
     - go: tip
   fast_finish: true
+sudo: false
+
+env:
+  global:
+    - STRIPE_STUB_PORT=6065
+    - STRIPE_STUB_VERSION=0.1.1
+
+cache:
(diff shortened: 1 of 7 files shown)


plan_test.go
```diff
@@ -0,0 +1,40 @@
+package stripe
+
+import (
+	"strconv"
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestPlanParamsAppendTo(t *testing.T) {
```
brandur: Note that a major change to testing strategy here is that parameter serialization starts getting its own first class tests. This represents a huge share of stripe-go by line count and is very prone to error (we regularly get user patches for this stuff), and yet has very little in the way of direct testing.

This test serializes parameters to a form body, and then immediately checks the result.

plan/client_test.go
```diff
@@ -1,193 +1,56 @@
...
-	if target.TrialPeriod != planParams.TrialPeriod {
-		t.Errorf("Trial period %v does not match expected trial period %v\n", target.TrialPeriod, planParams.TrialPeriod)
-	}
-
-	if target.Statement != planParams.Statement {
-		t.Errorf("Statement %q does not match expected statement %q\n", target.Statement, planParams.Statement)
-	}
-
-	Del(planParams.ID)
+func TestPlanDel(t *testing.T) {
```
brandur: The client tests represent one of the largest sections of change: instead of doing elaborate and brittle setup to make slow calls against the API that don't run in CI because they fail intermittently and are slow, we just do the simplest possible calls against the API stub.

The stub will perform some basic parameter validation, but in general is far more forgiving: it'll give us back a resource no matter what ID we asked for (and generally most of all operations will succeed), so we can significantly shorten these tests.

~200 lines down to ~50 and probably on the order of 3-4 orders of magnitude faster. The only downside is that we don't get to test the API itself, which is fine.

plan/client_test.go
```diff
@@ -1,193 +1,56 @@
 package plan
 
 import (
-	"fmt"
 	"testing"
 
+	assert "github.com/stretchr/testify/require"
```
brandur: I pulled in testify so we can get assertions and stop all this manually constructed boilerplate everywhere which is slow to write, brittle, and gives horrible error messages. Testify's become an internal convention with other Stripe Go projects, so bringing it in here too should be fine.

plan/client.go
```diff
@@ -29,23 +28,7 @@ func New(params *stripe.PlanParams) (*stripe.Plan, error) {
...
-	}
-
-	if params.TrialPeriod > 0 {
-		body.Add("trial_period_days", strconv.FormatUint(params.TrialPeriod, 10))
-	}
-
-	if len(params.Statement) > 0 {
-		body.Add("statement_descriptor", params.Statement)
-	}
+	params.Params.AppendTo(body)
```
brandur: Lets move away from constructing request bodies from clients and into `AppendTo` functions implemented by parameter structs. This way we can test the logic much more safely (and it's also reusable which is another big advantage).

Makefile
```diff
@@ -13,6 +13,10 @@ check-gofmt:
 checkin:
 	STRIPE_KEY=$(STRIPE_KEY) go test -run "TestCheckin*" ./client
 
+	# Whitelisted sets of tests that use stripestub
+	go test -run "TestPlan*"
+	go test ./plan
```
brandur: And once again we start testing things in CI! What a novel idea!

.travis.yml
```diff
@@ -1,11 +1,38 @@
...
+  # Unpack and start the Stripe API stub so that the test suite can talk to it
+  - |
+    if [ ! -d "stripestubs/stripestub_${STRIPE_STUB_VERSION}" ]; then
+      mkdir -p stripestubs/stripestub_${STRIPE_STUB_VERSION}/
+      curl -L "https://github.com/brandur/stripestub/releases/download/v${STRIPE_STUB_VERSION}/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz" -o "stripestubs/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz"
+      tar -zxf "stripestubs/stripestub_${STRIPE_STUB_VERSION}_linux_amd64.tar.gz" -C "stripestubs/stripestub_${STRIPE_STUB_VERSION}/"
+    fi
+  - |
+    stripestubs/stripestub_${STRIPE_STUB_VERSION}/stripestub -port ${STRIPE_STUB_PORT} &
+    STRIPE_STUB_PID=$!
```
brandur: This looks daunting, but it's simple: we fetch a stripestub release, unpack it, and run it. This also does some caching to make setup a bit faster.

plan/client.go
```diff
@@ -89,19 +73,7 @@ func (c Client) Update(id string, params *stripe.PlanParams) (*stripe.Plan, erro
...
-
-		if len(params.Statement) > 0 {
-			body.Add("statement_descriptor", params.Statement)
-		}
-
-		if params.TrialPeriod > 0 {
-			body.Add("trial_period_days", strconv.FormatUint(params.TrialPeriod, 10))
-		}
-
+		params.Params.AppendTo(body)
```
kjc-stripe: This feels like a breaking change to me. If someone has been passing in a `params` object with the `Currency` field set, we'll now append that to the request and cause a 400 error. We may want to make this an explicit breaking change and have different types of `PlanParams` objects per request.

brandur-stripe: Yeah that's a good point ... a major version bump seems to be in order.

> We may want to make this an explicit breaking change and have different types of PlanParams objects per method.

Yeah, that would probably be the better way to go. We should maybe consider just how impactful even a major version bump will be (they're all annoying, but some can be _really_ annoying) because that would be the kind of change where you'd have to update literally every invocation of Stripe that you have in your codebase. Maybe worth the risk though.

brandur-stripe:
Closing in favor of #446.

PR #428, opened 2017-06-26
Fix retrieving a bank account with a customer

### LINKED ISSUE #426: Bug in BankAccount client.go
Hi,
There is a bug on line 100 in bankaccount/client.go. The GET request should take params.Customer but is being passed params.AccountID

`if len(params.Customer) > 0 {
		err = c.B.Call("GET", fmt.Sprintf("/customers/%v/bank_accounts/%v", params.AccountID, id), c.Key, body, commonParams, ba)
	} else if len(params.AccountID) > 0 {
		err = c.B.Call("GET", fmt.Sprintf("/accounts/%v/bank_accounts/%v", params.AccountID, id), c.Key, body, commonParams, ba)
	} else {
		err = errors.New("Invalid bank account params: either Customer or AccountID need to be set")
	}`

I have tested this locally and replacing params.Customer fixes the issue.

brandur:
As mentioned in #426, when retrieving a bank account with a customer ID,
it should be passed into the URL, but we are incorrectly passing an
account ID instead.

Fixes #426.

### DIFF
--- a/bankaccount/client.go
+++ b/bankaccount/client.go
@@ -97,7 +97,7 @@ func (c Client) Get(id string, params *stripe.BankAccountParams) (*stripe.BankAc
 	var err error
 
 	if len(params.Customer) > 0 {
-		err = c.B.Call("GET", fmt.Sprintf("/customers/%v/bank_accounts/%v", params.AccountID, id), c.Key, body, commonParams, ba)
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Released as 23.2.1.

Issue #425, opened 2017-06-23
All `Del` functions should take a params


brandur-stripe:
Deleting a resource doesn't technically anything besides an ID in most cases, but a problem with the majority of our `Del` functions right now is that there's no way to specify request-level information like `Stripe-Account` or `Idempotency-Key`. In other action types, this can be specified using the params type for that resource which inherits fields from `stripe.Params`.

We should fix this by just adding a params as a second argument to every `Del` function. It's going to be an annoying breaking change, but the migration is easy at least, and it'll leave things in better shape.

Currently we've got about five `Del` functions that take a params and the rest don't: 

```
$ ag 'func Del'
account/client.go
191:func Del(id string) (*stripe.Account, error) {

applepaydomain/client.go
52:func Del(id string) (*stripe.ApplePayDomain, error) {

bankaccount/client.go
145:func Del(id string, params *stripe.BankAccountParams) (*stripe.BankAccount, error) {

card/client.go
131:func Del(id string, params *stripe.CardParams) (*stripe.Card, error) {

coupon/client.go
113:func Del(id string) (*stripe.Coupon, error) {

customer/client.go
167:func Del(id string) (*stripe.Customer, error) {

discount/client.go
18:func Del(customerID string) (*stripe.Discount, error) {
31:func DelSub(subscriptionID string) (*stripe.Discount, error) {

ephemeralkey/client.go
53:func Del(id string) (*stripe.EphemeralKey, error) {

invoiceitem/client.go
113:func Del(id string) (*stripe.InvoiceItem, error) {

paymentsource/client.go
94:func Del(id string, params *stripe.CustomerSourceParams) (*stripe.PaymentSource, error) {

plan/client.go
117:func Del(id string, params *stripe.PlanParams) (*stripe.Plan, error) {

product/client.go
234:func Delete(id string) error {

recipient/client.go
105:func Del(id string) (*stripe.Recipient, error) {

sku/client.go
259:func Delete(id string) error {

subitem/client.go
116:func Del(id string, params *stripe.SubItemParams) (*stripe.SubItem, error) {
```

Tagging this as future for now.

brandur-stripe:
Also ... lets renamed product and sku's `Delete` to `Del` while we're at it ...

PR #429, opened 2017-06-27
Add parameters when calling `pay` on an invoice

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @ark-stripe

Users can now provide a `source` parameter when paying invoices. Previously, the Go library let users pass creation/update params when calling `Pay` which is not what we want. This PR adds a new `InvoicePayParams` struct to handle parameters for this specific API request. Currently, the only accepted parameter is `Source`.

### DIFF
--- a/invoice/client_test.go
+++ b/invoice/client_test.go
@@ -488,11 +488,11 @@ func TestAllInvoicesScenarios(t *testing.T) {
 		t.Errorf("Invoice was not reponed as expected and its value is %v", targetInvoiceOpened.Closed)
 	}
 
-	paidInvoice := &stripe.InvoiceParams{
-		Paid: true,
+	payInvoice := &stripe.InvoicePayParams{
+		Source: cust.Sources.Values[0].Card.ID,
 	}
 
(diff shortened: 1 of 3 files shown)


invoice/client_test.go
```diff
@@ -502,7 +502,7 @@ func TestAllInvoicesScenarios(t *testing.T) {
 		t.Errorf("Updated invoice paid status %v does not match expected true\n", targetInvoiceUpdated.Paid)
 	}
 
-	_, err = plan.Del(planParams.ID)
+	_, err = plan.Del(planParams.ID, nil)
```
ob-stripe: This is unrelated to the PR, but I had to make this modification otherwise the test wouldn't compile. I guess it's related to #425.

brandur-stripe: Yeah, I think more specifically #414 probably broke it.

brandur-stripe:
LGTM.

brandur-stripe:
Released as 24.0.0.

PR #432, opened 2017-07-14
Fix subitem test compilation

brandur:
Fixes compilation of subitem tests, which is currently broken. The tests
may still not actually run ...


siddarth:
lgtm

PR #433, opened 2017-07-17
subitem: add support for metadata

siddarth-stripe:
r? @brandur


brandur-stripe:
Should do the trick. Thanks!

brandur-stripe:
Released as 24.1.0.

PR #435, opened 2017-07-19
Add support for range queries on list endpoints

### LINKED ISSUE #434: `event.List`'s params do not match the REST API:
Specifically, the `created` param is an int64 in the client package, but according to the [API docs](https://stripe.com/docs/api#list_events), it should be a dictionary

https://github.com/stripe/stripe-go/blob/master/event/client.go#L43
```
if params.Created > 0 {
  body.Add("created", strconv.FormatInt(params.Created, 10))
}
```

brandur:
Adds support for "range queries" (the naming comes from the API's
internals) for list endpoints. These are for timestamp parameters that
filter a list by time based on fields like greater than, lesser than or
equal, etc.

~~This patch in particular adds support to event's `created` parameter,
but it's generic enough to extend into other resources.~~ I tried to go
through and add range queries everywhere that they belong. The one 
exception are the four ranges under order's `status_transitions`
subparam, which I decided to skip for now.

This is a breaking change.

Fixes #434.

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -72,7 +72,10 @@ type TxParams struct {
 // For more details see https://stripe.com/docs/api/#balance_history.
 type TxListParams struct {
 	ListParams
-	Created, Available    int64
+	Created               int64
+	CreatedRange          *RangeQueryParams
+	Available             int64
+	AvailableRange        *RangeQueryParams
 	Currency, Src, Payout string
 	Type                  TransactionType
 }
--- a/balance/client.go
+++ b/balance/client.go
@@ -94,8 +94,16 @@ func (c Client) List(params *stripe.TxListParams) *Iter {
 			body.Add("created", strconv.FormatInt(params.Created, 10))
 		}
 
+		if params.CreatedRange != nil {
+			params.CreatedRange.AppendTo(body, "created")
+		}
+
 		if params.Available > 0 {
 			body.Add("available_on", strconv.FormatInt(params.Available, 10))
+		}
+
+		if params.AvailableRange != nil {
+			params.AvailableRange.AppendTo(body, "available_on")
 		}
 
 		if len(params.Currency) > 0 {
--- a/charge.go
+++ b/charge.go
@@ -47,6 +47,7 @@ func (cp *ChargeParams) SetSource(sp interface{}) error {
 type ChargeListParams struct {
 	ListParams
 	Created       int64
+	CreatedRange  *RangeQueryParams
 	Customer      string
 	TransferGroup string
 }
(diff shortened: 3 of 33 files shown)


ob-stripe:
lgtm, though unless I'm missing something the PR doesn't cover all listable resources. It covers coupons, events, orders, payouts, plans and transfers, but there are still a bunch of other listable resources that support `created` ranges (charges, customers, etc.) and in some cases other specific ranges (e.g. balance_transaction support listing by `available_on` range).

brandur-stripe:
Thanks OB!

> though unless I'm missing something the PR doesn't cover all listable resources. It covers coupons, events, orders, payouts, plans and transfers, but there are still a bunch of other listable resources that support created ranges (charges, customers, etc.) and in some cases other specific ranges (e.g. balance_transaction support listing by available_on range).

Doh! I realize now that I was querying the code by only looking for `RangeQuery` which is a symbol that only works on "new" abstract API methods. Let me do another pass and put in the rest. I noticed while doing this that at least some range list parameters are improperly implemented as `int64` right now, so it'll be nice to get them in all at once so that we only have to release one breaking major version.

brandur-stripe:
ptal @ob-stripe Would you mind taking a look at this one again?

So I ended up re-evaluating my strategy here after I realized that all "range" parameters also support a simple usage, i.e. `created=123` instead of `created[gte]=123`. To support this simple case, I ended up teasing the range apart into a convention that has two separate parameters like this:

``` go
type TxListParams struct {
	ListParams
	Created               int64
	CreatedRange          *RangeQueryParams
...
```

And with the corresponding client code using:

``` go
		if params.Created > 0 {
			body.Add("created", strconv.FormatInt(params.Created, 10))
		}

		if params.CreatedRange != nil {
			params.CreatedRange.AppendTo(body, "created")
		}
```

I think there's a few nice things about this approach:

1. The API for the simple case gets a lot simpler (simply `Created: 123` instead of a struct initialization).
2. It's backwards compatible so this change no longer requires a major version release.
3. We have a pretty well-established existing convention of doubling up parameters when it's necessary to do so (e.g. `Quantity` vs. `QuantityZero`).

I also went back through our API methods and added range support to all parameters in here that are supported, so there are quite a few more change resources than my first push.

ob-stripe:
> all "range" parameters also support a simple usage, i.e. `created=123` instead of `created[gte]=123`.

TIL! I didn't know about this either. I like that your PR supports both and is backwards-compatible!

lgtm

brandur-stripe:
Thanks OB!

brandur-stripe:
Released as 24.2.0.

PR #438, opened 2017-07-27
Compile all tests in CI (even if they're not run)

brandur:
This patch changes the CI based test command from `./client` to `./...`.
For now, the `TestCheckin*` filter is still applied, meaning that no
additional tests are run, but the change is that at least now we try to
compile every single subpackage in the project.

It turns out that there were quite a few basic syntax errors that had
creeped into various packages, and even on some of our newer ones, which
suggests that people probably aren't even checking tests locally before
sending to CI (meaning that they have never successfully run even one
time).

Anyway, unfortunately this is still far from a perfect fix, but brings
us a step closer to something a little more sane.

r? @remi-stripe

### DIFF
--- a/reversal/client_test.go
+++ b/reversal/client_test.go
@@ -1,14 +1,7 @@
 package reversal
 
 import (
-	"fmt"
-	"testing"
-
 	stripe "github.com/stripe/stripe-go"
-	"github.com/stripe/stripe-go/charge"
-	"github.com/stripe/stripe-go/currency"
-	"github.com/stripe/stripe-go/recipient"
-	"github.com/stripe/stripe-go/transfer"
 	. "github.com/stripe/stripe-go/utils"
 )
 
@@ -16,6 +9,9 @@ func init() {
 	stripe.Key = GetTestKey()
 }
 
+// These tests need to be refactored to no longer use recipients. Create
+// transfers for a target account instead.
+/*
 func TestReversalNew(t *testing.T) {
 	chargeParams := &stripe.ChargeParams{
 		Amount:   1000,
@@ -98,3 +94,4 @@ func TestReversalGet(t *testing.T) {
 
 	fmt.Printf("%+v\n", target)
 }
+*/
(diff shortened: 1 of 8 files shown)


reversal/client_test.go
```diff
@@ -1,21 +1,24 @@
...
 	. "github.com/stripe/stripe-go/utils"
 )
 
 func init() {
 	stripe.Key = GetTestKey()
 }
 
+// These tests need to be refactored to no longer used recipients. Create
+// transfers for a target account instead.
+/*
```
brandur-stripe: I feel really bad about this one, but basically this entire suite needs to be refactored because `recipient.New` no longer exists, and it depends on that function to create new transfers. It shouldn't be too bad to get working again, but I'm hoping to catch it on an upcoming pass of the entire test suite for stripe-mock.

reversal/client_test.go
```diff
@@ -1,21 +1,24 @@
 package reversal
 
 import (
-	"fmt"
-	"testing"
+	//"fmt"
+	//"testing"
```
remi-stripe: should we remove instead?

reversal/client_test.go
```diff
@@ -1,21 +1,24 @@
...
 
 	stripe "github.com/stripe/stripe-go"
-	"github.com/stripe/stripe-go/charge"
-	"github.com/stripe/stripe-go/currency"
-	"github.com/stripe/stripe-go/recipient"
-	"github.com/stripe/stripe-go/transfer"
+	//"github.com/stripe/stripe-go/charge"
+	//"github.com/stripe/stripe-go/currency"
+	//"github.com/stripe/stripe-go/recipient"
+	//"github.com/stripe/stripe-go/transfer"
```
remi-stripe: Should we also remove those?

reversal/client_test.go
remi-stripe: Yeah it's not great but I think it's okay for now, we will soon delete Recipients from the lib completely I think

reversal/client_test.go
brandur-stripe: Yeah, good call. We may as well just take them out.

brandur-stripe:
Thanks Remi!

PR #440, opened 2017-07-28
Introduce new reflection package to do form encoding

brandur-stripe:
Builds on top of #424, but adds a new reflection package that's able to form-encode a parameters struct using tags on its fields (similar to how the `json` package works).

There's a fairly extensive test suite, but we also use the parameter test suite introduced in #424 to verify that we get the results we expect for `PlanParams` and `ParamListParams`.

Still do to:

* ~~Handled integer-indexed arrays like for `legal_entity[additional_owners]`.~~ Done.
* ~~Handle "zero" fields (e.g. `QuantityZero` versus `Quantity`) that we're introduced to work around the fact that we can't detect an unset and zeroed value.~~ Done.
* ~~Handle "empty" fields (e.g. `AdditionalOwnersEmpty` vs `AdditionalOwners`.~~ Done.
* ~~Handle "no" fields (e.g. `NoProration`).~~ Done.

cc @remi-stripe @ob-stripe @stripe/api-libraries Thoughts on this sort of approach?

---

Note: targets the branch in #424 until we can get that merged.

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -5,12 +5,20 @@ import (
 	"net/url"
 	"reflect"
 	"strconv"
+	"strings"
 )
 
 const tagName = "form"
 
+type formOptions struct {
+	// IndexedArray indicates that contrary to standard "Rack-style" form
+	// encoding, array items should be index like `arr[0]=...&arr[1]=...`
+	// (normally it'd be `arr[]=...`).
+	IndexedArray bool
+}
+
 func AppendTo(values *RequestValues, i interface{}) {
-	reflectValue(values, reflect.ValueOf(i), nil)
+	reflectValue(values, reflect.ValueOf(i), nil, nil)
 }
 
 func formatName(names []string) string {
@@ -25,7 +33,7 @@ func formatName(names []string) string {
 	return fullName
 }
 
-func reflectValue(values *RequestValues, val reflect.Value, names []string) {
+func reflectValue(values *RequestValues, val reflect.Value, names []string, options *formOptions) {
 	t := val.Type()
 
 	// Also do nothing if this is the type's zero value
@@ -40,7 +48,14 @@ func reflectValue(values *RequestValues, val reflect.Value, names []string) {
 		arrNames := append(names, "")
 
 		for i := 0; i < val.Len(); i++ {
-			reflectValue(values, val.Index(i), arrNames)
+			// The one exception to the above is when options have requested
+			// that this array/slice be indexed. In that case we produce a hash
+			// keyed with integers which the Stripe API knows how to interpret.
+			if options != nil && options.IndexedArray {
+				arrNames = append(names, strconv.Itoa(i))
+			}
+
+			reflectValue(values, val.Index(i), arrNames, nil)
 		}
 
 	case reflect.Bool:
@@ -59,7 +74,7 @@ func reflectValue(values *RequestValues, val reflect.Value, names []string) {
 		if val.IsNil() {
 			return
 		}
-		reflectValue(values, val.Elem(), names)
+		reflectValue(values, val.Elem(), names, nil)
 
 	case reflect.Map:
 		for _, keyVal := range val.MapKeys() {
@@ -67,14 +82,14 @@ func reflectValue(values *RequestValues, val reflect.Value, names []string) {
 				panic("Don't support serializing maps with non-string keys")
 			}
 
-			reflectValue(values, val.MapIndex(keyVal), append(names, keyVal.String()))
+			reflectValue(values, val.MapIndex(keyVal), append(names, keyVal.String()), nil)
 		}
 
 	case reflect.Ptr:
 		if val.IsNil() {
 			return
 		}
-		reflectValue(values, val.Elem(), names)
+		reflectValue(values, val.Elem(), names, options)
 
 	case reflect.String:
 		values.Add(formatName(names), val.String())
@@ -86,13 +101,28 @@ func reflectValue(values *RequestValues, val reflect.Value, names []string) {
 			if tag == "" {
 				continue
 			}
-			formName := tag
-			reflectValue(values, val.Field(i), append(names, formName))
+			formName, options := parseTag(tag)
+			reflectValue(values, val.Field(i), append(names, formName), options)
 		}
 
 	case reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64:
 		values.Add(formatName(names), strconv.FormatUint(val.Uint(), 10))
 	}
+}
+
+func parseTag(tag string) (string, *formOptions) {
+	parts := strings.Split(tag, ",")
+	formName := parts[0]
+	options := &formOptions{}
+
+	for i := 1; i < len(parts); i++ {
+		switch parts[i] {
+		case "indexed":
+			options.IndexedArray = true
+		}
+	}
+
+	return formName, options
 }
 
 // ---
--- a/form/form.go
+++ b/form/form.go
@@ -67,7 +67,19 @@ func reflectValue(values *RequestValues, val reflect.Value, names []string) {
 		values.Add(formatName(names), strconv.FormatInt(val.Int(), 10))
 
 	case reflect.Interface:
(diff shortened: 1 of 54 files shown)


form/form.go
```diff
@@ -0,0 +1,191 @@
...
+		values.Add(formatName(names), strconv.FormatUint(val.Uint(), 10))
+	}
+}
+
+// ---
+
+// RequestValues is a collection of values that can be submitted along with a
+// request that specifically allows for duplicate keys and encodes its entries
+// in the same order that they were added.
+type RequestValues struct {
```
brandur-stripe: Unfortunately, I had to copy in the `RequestValues` struct from `params.go` for now, which makes this incompatible with our clients because these are of different values. Moving `RequestValues` will probably end up making this a breaking change, but hopefully not too many people will be depending on it directly.

form/form.go
```diff
@@ -0,0 +1,279 @@
...
+}
+
+func parseTag(tag string) (string, *formOptions) {
+	var options *formOptions
+	parts := strings.Split(tag, ",")
+	name := parts[0]
+
+	for i := 1; i < len(parts); i++ {
+		switch parts[i] {
+		case "empty":
```
remi-stripe: can you end up with a tag that would have more than one option? Like can you have a field that is `indexed,invert,empty`?

form/form.go
```diff
@@ -0,0 +1,279 @@
...
+//     arr[][foo]=foo0&arr[][bar]=bar0&arr[][foo]=foo1&arr[][bar]=bar1
+//
+// Because url.Values is a map, values will be handled in a way that's grouped
+// by their key instead of in the order they were added. Therefore the above
+// may by encoded to something like (maps are unordered so the actual result is
+// somewhat non-deterministic):
+//
+//     arr[][foo]=foo0&arr[][foo]=foo1&arr[][bar]=bar0&arr[][bar]=bar1
+//
+// And thus result in an incorrect request to Stripe.
```
remi-stripe: I'm a little bit confused by the comment. It seems to imply that `ToValues` is not really needed but you do call it in this PR.

form/form_test.go
```diff
@@ -0,0 +1,257 @@
...
+	Uuint32    uint32  `form:"uint32"`
+	Uuint32Ptr *uint32 `form:"uint32_ptr"`
+	Uuint64    uint64  `form:"uint64"`
+	Uuint64Ptr *uint64 `form:"uint64_ptr"`
+
+	Zeroed bool `form:"zeroed,zero"`
+}
+
+type testSubStruct struct {
+	SubSubStruct testSubSubStruct `form:"subsubstruct"`
```
remi-stripe: Does it work if the substruct is just an abstraction but not part of the naming scheme in the URL parameters itself? Something where we want to abstract away something to make it easier to re-use in the library for example but the parameter itself would have no mention of the "level".

brandur-stripe:
Hey Remi, thanks for the feedback here! Responses inline.

> One thing I was wondering was how complex it would be to add JSON support if the Stripe API ever supported this in the future. Would this be mostly a rewrite or would it just be replacing the core of reflectValue with a new/different logic?

So honestly we could _mostly_ just use the built in `encoding/json` package, with the one exception being some kind of special consideration for the inverted/emptied/zero special types that we have. I think we could implement those with just a slightly customized `MarshalJSON` method (see how expandable types work today) for each type though and get out of writing our own reflection package.

> Separately, I was wondering how slow all of this dynamic logic was in reality? There seems to be a lot of string parsing and a ton of mini loops to build all the right names for each form param. It's likely to be negligible but that's the kind of thing that always bother me so wanted to raise. (and I see you added some notes about it while reading the PR further)

So there's no question it's slower, but also that it's still pretty fast. I'll try to write a benchmark before we ship anything so that have a good idea about this — I don't want to be the slowest component in people's stacks.

I examined the implementation of the `encoding/json` package while writing this, and they use a similar approach to mind, but with an "type encoder cache" that "detailed" reflection for any given type only needs to happen one time. We should be able to use a similar approach to get good results. Go's JSON encoding is slow compared to other modern languages (e.g. Rust), but still fast enough.

form/form.go
```diff
@@ -0,0 +1,279 @@
...
+	// Empty indicates that a field's value should be emptied in that its value
+	// should be an empty string. It's used to workaround the fact that an
+	// empty string is a string's zero value and wouldn't normally be encoded.
+	Empty bool
+
+	// Invert indicates that a boolean field's value should be inverted. False
+	// is the zero value for a boolean so it's convention in the library to
+	// specify a `No*` field to allow a false to be passed to the API. These
+	// fields should be annotated with `invert`.
+	Invert bool
```
brandur-stripe: Ah! So it's sort of horrible, but we actually kind of need this given the current implementation. If you look at something like `AccountParams`:

``` go
type AccountParams struct {
...
	DebitNegativeBal     bool                          `form:"debit_negative_balances"`
	NoDebitNegativeBal   bool                          `form:"debit_negative_balances,invert"`
...
}
```

Without `NoDebitNegativeBal`, there's no way to detect whether a `false` was specified for `DebitNegativeBal` (because `false` is `bool`'s zero type), and therefore no way to set `DebitNegativeBal` to `false`. I've seen other SDKs user pointers (`*bool`) to work around this problem, which is awkward in its own way, but possibly better. I think for the time being we'll need to stick to this convention.

form/form.go
```diff
@@ -0,0 +1,279 @@
...
+			values.Add(formatName(names), strconv.FormatBool(val.Bool()))
+		}
+
+	case reflect.Float32:
+		values.Add(formatName(names), strconv.FormatFloat(val.Float(), 'f', 4, 32))
+
+	case reflect.Float64:
+		values.Add(formatName(names), strconv.FormatFloat(val.Float(), 'f', 4, 64))
+
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
```
brandur-stripe: Yeah we should do that (we currently have a mix of all kinds of `int`, `int64` and `uint64` when basically everything could just be `int`), but lets try to establish that separately. It seems like just good hygiene for our encoder here to support all possible types that could be reflected.

form/form.go
```diff
@@ -0,0 +1,279 @@
...
+	case reflect.Interface:
+		if val.IsNil() {
+			return
+		}
+		reflectValue(values, val.Elem(), names, nil)
+
+	case reflect.Map:
+		for _, keyVal := range val.MapKeys() {
+			if keyVal.Kind() != reflect.String {
+				panic("Don't support serializing maps with non-string keys")
```
brandur-stripe: I don't think there's anything in our library besides string-keyed maps, but you could have a `map[int]string` or whatever. I'm trying to help ensure that people just don't try to do this by panicking so that we can catch the problem early :)

I could see the argument for returning an `error` though, which would be fairly conventional. I think we could refactor that over, but I'm basically following the convention set by today's `AppendTo` functions which also don't return errors.

form/form.go
brandur-stripe: @remi-stripe Yeah. Only one of them would be honored though because the combination isn't conceptually sound.

form/form.go
brandur-stripe: Yeah ... we can't use it live, but it's relatively useful for inspecting and working with `RequestValues`. Note that this code is all copied from `params.go` though, so we shouldn't scrutinize it too much on this PR probably.

form/form_test.go
brandur-stripe: Basically anything that's not a leaf primitive type with a `form` tag will get ignored.

So while a `string` under `testSubSubStruct` gets serialized as expected here, these would be ignored:

1. A `struct` without a `form` tag.
2. A `string` (or other primitive) with a `form` tag that's nested under a struct _without_ a `form` tag (we need a valid parent name to assemble a form key hierarchy like `param[struct][substruct][foo]...`).

brandur-stripe:
Implemented a fully functional encoder for account, which I believe is our most complex resource! See #442.

brandur-stripe:
Closing in favor of #446.

PR #443, opened 2017-08-08
Application fee can be set back to 0 on Invoice and Subscription

remi-stripe:
Via Stripe Connect, you can set an `application_fee_percent` on a subscription or an `application_fee` on an invoice. This can be set to 0 but as usual the Go library does not support this.

This PR fixes the bug by introducing `FeePercentZero` and `FeeZero`.

r? @brandur

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -17,6 +17,7 @@ type InvoiceParams struct {
 	Customer             string
 	Desc, Statement, Sub string
 	Fee                  uint64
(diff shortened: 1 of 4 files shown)


brandur-stripe:
Thanks Remi!

The build failure is unrelated. Pulling this in.

brandur-stripe:
Released as 24.3.0.

PR #444, opened 2017-08-17
Fix Del methods to always take an extra parameter and be named Del

remi-stripe:
This PR fixes https://github.com/stripe/stripe-go/issues/425 and ensures that all Del methods behave the same way.

First, Product/SKU Delete methods were renamed to Del and now return the deleted resource too.

All Del methods now properly take an extra parameter. This lets you pass custom values such as an IdempotencyKey and where relevant the id of the connected account that you're making the request on behalf of.

Also update the Readme to better reflect how Del works.

### DIFF
--- a/product/client.go
+++ b/product/client.go
@@ -231,14 +231,27 @@ func (i *Iter) Product() *stripe.Product {
 
 // Delete deletes a product
 // For more details see https://stripe.com/docs/api#delete_product.
-func Delete(id string) error {
-	return getC().Delete(id)
+func Del(id string, params *stripe.ProductParams) (*stripe.Product, error) {
+	return getC().Del(id, params)
 }
 
 // Delete deletes a product.
 // For more details see https://stripe.com/docs/api#delete_product.
-func (c Client) Delete(id string) error {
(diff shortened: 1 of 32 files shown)


remi-stripe:
This is a work in progress because that's a big change that we need to get right the first time. I'm getting various build failures for some tests locally and trying to debug why.

cc @brandur  for awareness

remi-stripe:
Okay, I figured out my testing issues (my GOPATH was again pointing to a different version of the code) so I think it's ready for review/feedback.

r? @brandur-stripe

product/client.go
```diff
@@ -231,14 +231,27 @@ func (i *Iter) Product() *stripe.Product {
 
 // Delete deletes a product
 // For more details see https://stripe.com/docs/api#delete_product.
-func Delete(id string) error {
-	return getC().Delete(id)
+func Del(id string, params *stripe.ProductParams) (*stripe.Product, error) {
```
brandur-stripe: Arg. Thanks for fixing that too.

brandur-stripe:
Released as 25.0.0.

PR #448, opened 2017-09-12
Add missing OnBehalfOf

jzs:
Reading the documentation, there is a functionality of adding an OnBehalfOf parameter When creating separate charges:
https://stripe.com/docs/connect/charges-transfers#on-behalf-of

This is also the case for creating subscriptions, however it is not currently documented in the official documentation.

### DIFF
--- a/sub.go
+++ b/sub.go
@@ -29,6 +29,7 @@ type SubParams struct {
 	Items                                           []*SubItemsParams
 	Billing                                         SubBilling
 	DaysUntilDue                                    uint64
+	OnBehalfOf                                      string
 }
 
 // SubItemsParams is the set of parameters that can be used when creating or updating a subscription item on a subscription
--- a/charge.go
+++ b/charge.go
@@ -27,6 +27,7 @@ type ChargeParams struct {
 	Source                       *SourceParams
 	Shipping                     *ShippingDetails
 	TransferGroup                string
+	OnBehalfOf                   string
 }
 
 type DestinationParams struct {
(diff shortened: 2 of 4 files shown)


sub.go
```diff
@@ -29,6 +29,7 @@ type SubParams struct {
 	Items                                           []*SubItemsParams
 	Billing                                         SubBilling
 	DaysUntilDue                                    uint64
+	OnBehalfOf                                      string
```
remi-stripe: @jzs This is undocumented, the `OnBehalfOf` is only supported on Charge creation.

remi-stripe:
@jzs Thanks for the PR! I just added a quick comment but it looks good to me. We might wait a bit to merge this one because we want to land https://github.com/stripe/stripe-go/pull/446 first which is a big refactor of the library and then we'll merge yours in afterwards. I'll defer to @brandur-stripe  on the decision though.

brandur-stripe:
> We might wait a bit to merge this one because we want to land #446 first which is a big refactor of the library and then we'll merge yours in afterwards. I'll defer to @brandur-stripe on the decision though.

If the pull's updated before #446 comes in, I think it's fine to merge. I've got enough practice rebasing at this point that it's not a big deal :)

sub.go
jzs: @remi-stripe @brandur-stripe The server side api does actually support `on_behalf_of` when creating subscriptions and we do need that feature to keep charges in correct currency.
However if there are reasons it can't be included we would just have to fork the client which is always not the preferred option :)

remi-stripe: I'm never a big fan of merging things that are not really supported but if you do need it I don't see a safer/better solution in this case.

remi-stripe:
LGTM

@brandur-stripe can you merge as is?

brandur-stripe:
Yeah, works for me.

brandur-stripe:
Released as 25.2.0.

Issue #449, opened 2017-09-15
Library cleanup


brandur-stripe:
During the review of #446, quite a few possibilities for improvements and cleanup of the library came up. Here's the list:

* [x] ~~Remove `Email` from token.go now that Stripe Checkout doesn't seem to use this library anymore.~~ (Actually, still in use.)
* [x] Remove `Display` function `BankAccount` and the like. It doesn't add very much (see #451).
* [ ] Move ChargeParam.Fraud to be its own subparameter struct called something like `FraudDetails` (see #452).
* [ ] Drop recipients entirely given that it's all deprecated (and drop `form`'s `AppendToPrefixed` which is only needed for recipients).
* [ ] Optimize performance of the `form` package's `AppendTo` (see `BenchmarkAppendTo` for help with this).
* [x] Order all parameter and resource struct fields alphabetically (see #450).
* [ ] Replace weird truncates (e.g. `Src`) with expanded words (e.g. `Source`). These should be as close to the name of the API field/param as much as possible for least confusion.
* [ ] Get rid of all "shared" parameter/resource structs. There should be no structs that have fields annotated with both `json` and `form` tags.

brandur-stripe:
@remi-stripe Doh. I was wrong about Checkout no longer using this library. The JS bundle itself doesn't of course, but its backend API still appears to. Unfortunately we won't be able to remove `Email` for now.

brandur-stripe:
Alright most of these are done, and I've opened two new fields (#456 and #457) for those that weren't.

PR #450, opened 2017-09-15
Alphabetize the world

brandur:
Alphabetizes all API parameter and resource structs. I'm not going to pretend like this was a great use of time, but at least it will allow us to have some kind of convention to point to so that we can help control consistency in the project.

This is technically breaking because you can initialize structs with positional order in Go (e.g. `MyStruct{1, 2}`), but I doubt it'll break almost anyone because (1) most people won't use the positional initialization for these sorts of big structs, and (2) even if they were, type mismatches in the new order will let the compiler tell them about the problem. That said, there are a few potential danger spots that I noticed like `PackageDimemsions` for Relay which is four fields with all exactly the same type.

Let's put a note in the changelog but ship it as a major version. I haven't released #446 yet, so I'd like to group it in with that.

Fixes part of #449.

r? @remi-stripe Sorry for another huge diff, but I think this one is relatively benign. Are you okay with it?

### DIFF
--- a/account.go
+++ b/account.go
@@ -133,37 +133,39 @@ type Account struct {
 	Email                string               `json:"email"`
 	ExternalAccounts     *ExternalAccountList `json:"external_accounts"`
 	ID                   string               `json:"id"`
-	LegalEntity          *LegalEntity         `json:"legal_entity"`
-	Meta                 map[string]string    `json:"metadata"`
-	Name                 string               `json:"display_name"`
-	PayoutSchedule       *PayoutSchedule      `json:"payout_schedule"`
-	PayoutStatement      string               `json:"payout_statement_descriptor"`
-	PayoutsEnabled       bool                 `json:"payouts_enabled"`
-	ProductDesc          string               `json:"product_description"`
-	Statement            string               `json:"statement_descriptor"`
-	SupportAddress       *Address             `json:"support_address"`
-	SupportEmail         string               `json:"support_email"`
-	SupportPhone         string               `json:"support_phone"`
-	SupportUrl           string               `json:"support_url"`
-	Timezone             string               `json:"timezone"`
-	Type                 AccountType
 
 	Keys *struct {
 		Publish string `json:"publishable"`
 		Secret  string `json:"secret"`
 	} `json:"keys"`
 
-	Verification *struct {
-		DisabledReason string   `json:"disabled_reason"`
-		Due            *int64   `json:"due_by"`
-		Fields         []string `json:"fields_needed"`
-	} `json:"verification"`
+	LegalEntity     *LegalEntity      `json:"legal_entity"`
+	Meta            map[string]string `json:"metadata"`
+	Name            string            `json:"display_name"`
+	PayoutSchedule  *PayoutSchedule   `json:"payout_schedule"`
+	PayoutStatement string            `json:"payout_statement_descriptor"`
+	PayoutsEnabled  bool              `json:"payouts_enabled"`
+	ProductDesc     string            `json:"product_description"`
+	Statement       string            `json:"statement_descriptor"`
+	SupportAddress  *Address          `json:"support_address"`
+	SupportEmail    string            `json:"support_email"`
+	SupportPhone    string            `json:"support_phone"`
+	SupportUrl      string            `json:"support_url"`
+	Timezone        string            `json:"timezone"`
 
 	TOSAcceptance *struct {
 		Date      int64  `json:"date"`
 		IP        string `json:"ip"`
 		UserAgent string `json:"user_agent"`
 	} `json:"tos_acceptance"`
+
+	Type AccountType
+
+	Verification *struct {
+		DisabledReason string   `json:"disabled_reason"`
+		Due            *int64   `json:"due_by"`
+		Fields         []string `json:"fields_needed"`
+	} `json:"verification"`
 }
 
 // UnmarshalJSON handles deserialization of an account.
--- a/account.go
+++ b/account.go
@@ -27,58 +27,58 @@ type IdentityVerificationStatus string
 type Interval string
 
 const (
+	// Company is a constant value representing a company legal entity type.
+	Company LegalEntityType = "company"
+
+	// Day is a constant value representing a daily payout interval.
+	Day Interval = "daily"
+
 	// Individual is a constant value representing an individual legal entity
 	// type.
 	Individual LegalEntityType = "individual"
-
-	// Company is a constant value representing a company legal entity type.
-	Company LegalEntityType = "company"
 
 	// IdentityVerificationPending is a constant value indicating that identity
 	// verification status is pending.
 	IdentityVerificationPending IdentityVerificationStatus = "pending"
 
-	// IdentityVerificationVerified is a constant value indicating that
-	// identity verification status is verified.
-	IdentityVerificationVerified IdentityVerificationStatus = "verified"
-
 	// IdentityVerificationUnverified is a constant value indicating that
 	// identity verification status is unverified.
 	IdentityVerificationUnverified IdentityVerificationStatus = "unverified"
 
+	// IdentityVerificationVerified is a constant value indicating that
+	// identity verification status is verified.
+	IdentityVerificationVerified IdentityVerificationStatus = "verified"
+
 	// Manual is a constant value representing a manual payout interval.
 	Manual Interval = "manual"
 
-	// Day is a constant value representing a daily payout interval.
-	Day Interval = "daily"
+	// Month is a constant value representing a monthly payout interval.
+	Month Interval = "monthly"
 
 	// Week is a constant value representing a weekly payout interval.
 	Week Interval = "weekly"
-
-	// Month is a constant value representing a monthly payout interval.
-	Month Interval = "monthly"
 )
 
 // AccountParams are the parameters allowed during account creation/updates.
 type AccountParams struct {
 	Params               `form:"*"`
-	Country              string                        `form:"country"`
-	Email                string                        `form:"email"`
-	DefaultCurrency      string                        `form:"default_currency"`
-	Statement            string                        `form:"statement_descriptor"`
 	BusinessName         string                        `form:"business_name"`
-	BusinessUrl          string                        `form:"business_url"`
 	BusinessPrimaryColor string                        `form:"business_primary_color"`
-	SupportPhone         string                        `form:"support_phone"`
-	SupportEmail         string                        `form:"support_email"`
-	SupportUrl           string                        `form:"support_url"`
-	FromRecipient        string                        `form:"from_recipient"`
-	PayoutStatement      string                        `form:"payout_statement_descriptor"`
-	ExternalAccount      *AccountExternalAccountParams `form:"external_account"`
-	LegalEntity          *LegalEntity                  `form:"legal_entity"`
-	PayoutSchedule       *PayoutScheduleParams         `form:"payout_schedule"`
+	BusinessUrl          string                        `form:"business_url"`
+	Country              string                        `form:"country"`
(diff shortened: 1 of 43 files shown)


account.go
```diff
@@ -120,47 +120,50 @@ func (p *PayoutScheduleParams) AppendTo(body *form.Values, keyParts []string) {
...
+	SupportPhone         string               `json:"support_phone"`
 	SupportUrl           string               `json:"support_url"`
-	ProductDesc          string               `json:"product_description"`
-	DebitNegativeBal     bool                 `json:"debit_negative_balances"`
+	Timezone             string               `json:"timezone"`
 	Type                 AccountType
-	Keys                 *struct {
-		Secret  string `json:"secret"`
+
+	Keys *struct {
```
remi-stripe: ooc why isn't Keys ordered alphabetically?

account.go
```diff
@@ -120,47 +120,50 @@ func (p *PayoutScheduleParams) AppendTo(body *form.Values, keyParts []string) {
...
-		Due            *int64   `json:"due_by"`
 		DisabledReason string   `json:"disabled_reason"`
+		Due            *int64   `json:"due_by"`
+		Fields         []string `json:"fields_needed"`
 	} `json:"verification"`
-	LegalEntity    *LegalEntity    `json:"legal_entity"`
-	PayoutSchedule *PayoutSchedule `json:"payout_schedule"`
-	TOSAcceptance  *struct {
+
+	TOSAcceptance *struct {
```
remi-stripe: same as Keys, why isn't it ordered above?

bitcoinreceiver.go
```diff
@@ -37,24 +37,24 @@ type BitcoinReceiverUpdateParams struct {
...
+	ID                    string                  `json:"id"`
 	InboundAddress        string                  `json:"inbound_address"`
-	RefundAddress         string                  `json:"refund_address"`
-	BitcoinUri            string                  `json:"bitcoin_uri"`
 	Meta                  map[string]string       `json:"metadata"`
-	Email                 string                  `json:"email"`
 	Payment               string                  `json:"payment"`
-	Customer              string                  `json:"customer"`
+	RejectTransactions    bool                    `json:"reject_transactions"`
+	RefundAddress         string                  `json:"refund_address"`
```
remi-stripe: nitpicky but `RefundAddress` should be before `RejectTransactions`

card.go
```diff
@@ -33,29 +37,25 @@ type TokenizationMethod string
...
-	Recipient string `form:"-"`
-	Name      string `form:"name"`
-	Number    string `form:"number"`
-	Month     string `form:"exp_month"`
-	Year      string `form:"exp_year"`
-	CVC       string `form:"-"`
-	Currency  string `form:"currency"`
 	Address1  string `form:"address_line1"`
 	Address2  string `form:"address_line2"`
+	CVC       string `form:"-"`
```
remi-stripe: CVC should be after Customer (I guess it's because of casing?

brandur-stripe:
I forgot to note is that the one downside of this is that it moves our `No`-prefixed fields away from their inverted field. So for example, `Prorate` and `NoProrate` might have a few fields separating them when ideally they'd be side-by-side.

We should probably consider renaming those to a suffix instead of a prefix at some point so that they'll play a little nicer, e.g. `Prorate` and `ProrateNo`. This would fit well with how our "zero" fields already look (e.g. `Quantity` and `QuantityZero`).

card.go
```diff
@@ -148,40 +148,30 @@ type CardListParams struct {
...
-	Recipient          *Recipient         `json:"recipient"`
-	DynLastFour        string             `json:"dynamic_last4"`
-	Deleted            bool               `json:"deleted"`
-	ThreeDSecure       *ThreeDSecure      `json:"three_d_secure"`
-	TokenizationMethod TokenizationMethod `json:"tokenization_method"`
+	Address1      string       `json:"address_line1"`
+	Address1Check Verification `json:"address_line1_check"`
+	Address2      string       `json:"address_line2"`
+	Brand         CardBrand    `json:"brand"`
+	CVCCheck      Verification `json:"cvc_check"`
```
remi-stripe: CVC should be after Currency

ephemeralkey.go
```diff
@@ -30,6 +26,11 @@ type EphemeralKey struct {
 	// from the version of these bindings, we can still pass back a compatible
 	// key.
 	RawJSON []byte `json:"-"`
+
+	AssociatedObjects []struct {
```
remi-stripe: this should be at the top of the struct

account.go
brandur-stripe: Yeah, I alphabetized most things, but went back and forth on this one. These unnamed structs are a little messy so I ended up leaving them at the bottom of the block.

That said, I just changed my mind and ordered them in a5e4dd1. There aren't very many of these so I don't think it matters very much.

bitcoinreceiver.go
brandur-stripe: Good catch. Fixed in 2a1fcea. I figure that I probably missorted a couple others too, but we can fix them as we go.

card.go
brandur-stripe: Yeah. Generally with lexigraphic sorting, the capitals come first:

```
irb(main):004:0> ["Customer", "CVC"].sort
=> ["CVC", "Customer"]
```

I followed this convention here.

ephemeralkey.go
brandur-stripe: Hoisted in 5140873.

brandur-stripe:
@remi-stripe Addressed feedback above. How do you feel about pulling this in?

feerefund.go
```diff
@@ -22,13 +22,13 @@ type FeeRefundListParams struct {
 // FeeRefund is the resource representing a Stripe fee refund.
 // For more details see https://stripe.com/docs/api#fee_refunds.
 type FeeRefund struct {
-	ID       string            `json:"id"`
 	Amount   uint64            `json:"amount"`
+	ID       string            `json:"id"`
```
remi-stripe: ID should be later in the list

fileupload.go
```diff
@@ -12,18 +12,18 @@ import (
...
 	// FileReader is a reader with the contents of the file that should be uploaded.
 	FileReader io.Reader
 
-	// File is a deprecated form of FileReader and Filename that will do the same thing, but
-	// allows referencing a file directly. Please prefer the use of FileReader and Filename instead.
-	File *os.File
+	// Filename is just the name of the file without path information.
+	Filename string
+
+	Params  `form:"*"`
```
remi-stripe: In other classes you kept Params at the top. I think it's "worth" keeping always at the top since it's not really a parameter

invoice.go
```diff
@@ -14,93 +14,94 @@ type InvoiceBilling string
 // For more details see https://stripe.com/docs/api#create_invoice, https://stripe.com/docs/api#update_invoice.
 type InvoiceParams struct {
 	Params         `form:"*"`
+	Billing        InvoiceBilling `form:"billing"`
 	Customer       string         `form:"customer"`
+	Closed         bool           `form:"closed"`
```
remi-stripe: Closed should be before Customer

invoice.go
```diff
@@ -14,93 +14,94 @@ type InvoiceBilling string
...
 	// These are all for exclusive use by GetNext.
 
-	SubPlan          string            `form:"subscription_plan"`
 	SubNoProrate     bool              `form:"subscription_prorate,invert"`
+	SubPlan          string            `form:"subscription_plan"`
 	SubProrationDate int64             `form:"subscription_proration_date"`
 	SubQuantity      uint64            `form:"subscription_quantity"`
 	SubQuantityZero  bool              `form:"subscription_quantity,zero"`
-	SubTrialEnd      int64             `form:"subscription_trial_end"`
 	SubItems         []*SubItemsParams `form:"subscription_items,indexed"`
```
remi-stripe: Should be earlier in the list

invoice.go
```diff
@@ -14,93 +14,94 @@ type InvoiceBilling string
...
 	TaxPercent    float64           `json:"tax_percent"`
-	Fee           uint64            `json:"application_fee"`
-	Charge        *Charge           `json:"charge"`
-	Desc          string            `json:"description"`
-	Discount      *Discount         `json:"discount"`
-	ReceiptNumber string            `json:"receipt_number"`
-	EndBalance    int64             `json:"ending_balance"`
-	NextAttempt   int64             `json:"next_payment_attempt"`
+	Total         int64             `json:"total"`
 	Statement     string            `json:"statement_descriptor"`
```
remi-stripe: Should be earlier in the list, same for Sub (next line)

order.go
```diff
@@ -46,17 +46,17 @@ type OrderReturnParams struct {
...
-	Name    string  `json:"name"`
 	Address Address `json:"address"`
+	Name    string  `json:"name"`
 	Phone   string  `json:"phone"`
 }
 
 type ShippingMethod struct {
-	ID               string            `json:"id"`
 	Amount           int64             `json:"amount"`
+	ID               string            `json:"id"`
```
remi-stripe: Should be later in the list

orderreturn.go
```diff
@@ -3,13 +3,13 @@ package stripe
...
 type OrderReturn struct {
-	ID       string      `json:"id"`
 	Amount   int64       `json:"amount"`
 	Created  int64       `json:"created"`
 	Currency Currency    `json:"currency"`
+	ID       string      `json:"id"`
 	Items    []OrderItem `json:"items"`
-	Live     bool        `json:"livemode"`
 	Order    Order       `json:"order"`
+	Live     bool        `json:"livemode"`
```
remi-stripe: Live should be before Order

recipienttransfer.go
```diff
@@ -48,17 +48,15 @@ const (
 // The Type should indicate which object is fleshed out
 // For more details see https://stripe.com/docs/api/go#recipient_transfer_object
 type RecipientTransferDestination struct {
-	Type        RecipientTransferDestinationType `json:"object"`
-	ID          string                           `json:"id"`
-	BankAccount *BankAccount                     `json:"-"`
 	Card        *Card                            `json:"-"`
+	BankAccount *BankAccount                     `json:"-"`
```
remi-stripe: BankAccount should be before

sub.go
```diff
@@ -18,29 +18,29 @@ type SubBilling string
...
 	Items                 []*SubItemsParams `form:"items,indexed"`
-	Billing               SubBilling        `form:"billing"`
-	DaysUntilDue          uint64            `form:"days_until_due"`
+	NoProrate             bool              `form:"prorate,invert"`
 	OnBehalfOf            string            `form:"on_behalf_of"`
+	Plan                  string            `form:"plan"`
+	ProrationDate         int64             `form:"proration_date"`
+	Quantity              uint64            `form:"quantity"`
+	QuantityZero          bool              `form:"quantity,zero"`
+	Token                 string            `form:"card"`
```
remi-stripe: should be after

sub.go
```diff
@@ -64,49 +64,49 @@ func (p *SubParams) AppendTo(body *form.Values, keyParts []string) {
...
 	Ended        int64             `json:"ended_at"`
+	FeePercent   float64           `json:"application_fee_percent"`
+	ID           string            `json:"id"`
+	Items        *SubItemList      `json:"items"`
 	Meta         map[string]string `json:"metadata"`
+	PeriodEnd    int64             `json:"current_period_end"`
+	PeriodStart  int64             `json:"current_period_start"`
+	Plan         *Plan             `json:"plan"`
+	Quantity     uint64            `json:"quantity"`
+	Status       SubStatus         `json:"status"`
```
remi-stripe: Status should be after Start

brandur-stripe:
Thanks a lot Remi! Fixed all your feedback in 16c900d.

PR #453, opened 2017-09-15
Drop ChargeParams.Dest

brandur:
Drop `ChargeParams.Dest` in favor of `ChargeParams.Destination.Account`,
which is aligned with how the API is designed. This also allows us to
completely drop a custom `AppendTo` implementation on `Charge`.

Fixes part of #449.

r? @remi-stripe Sorry about the boatload of reviews, but I'd previously
discussed this one with you as well.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -2,8 +2,6 @@ package stripe
 
 import (
 	"encoding/json"
-
-	"github.com/stripe/stripe-go/form"
 )
 
 // Currency is the list of supported currencies.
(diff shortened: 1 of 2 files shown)


remi-stripe:
ha you need to remove the extra form require

brandur-stripe:
> ha you need to remove the extra form require

Arg, totally right. Rebased with that in 7ec7666.

brandur-stripe:
Thanks Remi!

PR #454, opened 2017-09-16
Fix bug where Params.AppendTo not called if leaf had its own implementation

brandur:
In Go you can "embed" a struct into another struct [1]:

    type ChargeParams struct {
        Params
    }

Embeds mostly behave like normal fields, except that they have the
special property that they allow the leaf struct to inherit any
functions implemented on the embedded struct.

For `Params`, this means that most parameter structs inherited the
`AppendTo` implementation of basic `Params`, which is where we encoded
our "extra" information.

However, when a leaf declares its own function by the same name, it no
longer inherits its embedded struct's implementation. This means that
any params that declare their own `AppendTo` will not currently encode
extras, because the `AppendTo` of `Params` doesn't get called.

This patch shifts things around a little bit so that we move encoding of
extras into an `AppendTo` on a new struct `ExtraValues`, which is an
alias for `url.Values`. This is breaking if a user was initializing
`Extra` directly, but hopefully most users were using the preferred
`params.AddExtra` instead. Regardless, this is the least invasive way I
could think of to fix this, so I think we'll have to go through with it.

We also do something similar for filters in `ListParams` (although no
`ListParams` implements an `AppendTo` so this is more of a hedge against
possible future problems) and change some tests around to better detect
these conditions (it wasn't very well tested before which is why the bug
happened).

r? @remi-stripe

[1] https://golang.org/doc/effective_go.html#embedding

### DIFF
--- a/params.go
+++ b/params.go
@@ -23,8 +23,8 @@ type Params struct {
 	// Please use StripeAccount instead.
 	Account string `form:"-"` // Passed as header
 
-	Exp   []string   `form:"expand"`
-	Extra url.Values `form:"-"` // See custom encoding in AppendTo
+	Exp   []string     `form:"expand"`
+	Extra *ExtraValues `form:"*"`
 
 	// Headers may be used to provide extra header lines on the HTTP request.
 	Headers http.Header `form:"-"`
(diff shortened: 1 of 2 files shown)


brandur-stripe:
It also looks like that the original `form` implementation might have a fun where `AppendTo` might get called twice, and this is failing one of the tests that I tweaked.

It might be best to review this along with #455, which fixes that bug as well optimizing, and then bring them in simultaneously.

remi-stripe:
I looked at the build error but I'm not sure it's related here and I don't understand why the keys are duplicated

brandur-stripe:
Thanks Remi!

Yeah, I don't think the build error is related to any changes in this PR — it dropped out because I changed a test, and that revealed a bug elsewhere. #455 targets this branch and the problem disappears though, so the reimplementation has fixed it I think.

PR #455, opened 2017-09-16
Build a type encoder caching system for the `form` package

brandur:
Here we optimize the `AppendTo` function of the form package and its
associated helpers. As #446 came in, there was some concerns expressed
around its performance. This implementation might not be as quick as a
bunch of `if` statements, but it's pretty fast.

Before:

    $ make bench
    go test -bench . -run "Benchmark" ./form
    goos: darwin
    goarch: amd64
    pkg: github.com/stripe/stripe-go/form
    BenchmarkAppendTo-4       500000              2767 ns/op
    PASS
    ok      github.com/stripe/stripe-go/form        1.480s

After:

    $ make bench
    go test -bench . -run "Benchmark" ./form
    goos: darwin
    goarch: amd64
    pkg: github.com/stripe/stripe-go/form
    BenchmarkAppendTo-4      2000000               922 ns/op
    PASS
    ok      github.com/stripe/stripe-go/form        2.867s

Which shows roughly a 3x speedup. Encoding a fairly normal struct 
takes < 1 ms on my commodity PC.

The implementation here looks pretty exotic, but it's very inspired by
Go's own `encoding/json` which is where the ideas of type and struct 
encoding caches comes from.

Testing is pretty extensive, so I have pretty good confidence that this
works as expected based off of the passing suite.

Fixes part of #449.

r? @remi-stripe 

---

~~Currently targets #454 because we want this params-related bug fix.~~

### DIFF
--- a/account.go
+++ b/account.go
@@ -340,10 +340,10 @@ type Owner struct {
 
 // IdentityVerification is the structure for an account's verification.
 type IdentityVerification struct {
-	Details     *string                         `json:"details"`
-	DetailsCode IdentityVerificationDetailsCode `json:"details_code"`
+	Details     *string                         `json:"details" form:"-"`
+	DetailsCode IdentityVerificationDetailsCode `json:"details_code" form:"-"`
 	Document    *IdentityDocument               `json:"document" form:"document"`
-	Status      IdentityVerificationStatus      `json:"status"`
+	Status      IdentityVerificationStatus      `json:"status" form:"-"`
 }
 
 // IdentityDocument is the structure for an identity document.
--- a/form/form.go
+++ b/form/form.go
@@ -7,9 +7,37 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 )
 
 const tagName = "form"
+
+// Appender is the interface implemented by types that can append themselves to
+// a collection of form values.
+//
+// This is usually something that shouldn't be used, but is needed in a few
+// places where authors deviated from norms while implementing various
+// parameters.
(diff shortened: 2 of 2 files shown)


account.go
```diff
@@ -340,10 +340,10 @@ type Owner struct {
 
 // IdentityVerification is the structure for an account's verification.
 type IdentityVerification struct {
-	Details     *string                         `json:"details"`
-	DetailsCode IdentityVerificationDetailsCode `json:"details_code"`
+	Details     *string                         `json:"details" form:"-"`
+	DetailsCode IdentityVerificationDetailsCode `json:"details_code" form:"-"`
 	Document    *IdentityDocument               `json:"document" form:"document"`
-	Status      IdentityVerificationStatus      `json:"status"`
+	Status      IdentityVerificationStatus      `json:"status" form:"-"`
```
brandur-stripe: Not sure why my `panic` check didn't catch this one before, but I found these untagged fields while running tests on this branch.

brandur-stripe:
Also worth noting that way more optimization work could be done here — I didn't even pull out a profiler for this one. I think it's worth pulling in the 80% win first though, and then moving onto to coaxing out the last 20%.

brandur-stripe:
> This looks good to me but I'll admit it goes a bit beyond what I could grasp just reviewing so I'm not confident I would catch an edge-case or something you missed.

Yeah I totally get it. I think in this sort of case we need to be relying heavily on the test suite, which did catch quite a few problems as I was writing it.

Thanks for taking a look!

brandur-stripe:
Released all recent changes as 26.0.0.

PR #458, opened 2017-09-17
Remove Charge.Token

brandur:
Unfortunately I forgot about this one before releasing, but it doesn't
appear that `Token` on `ChargeParams` is used anywhere and I can't find
it in any of the charge-related documentation on the API reference.

@remi-stripe Thoughts?


remi-stripe:
LGTM. I have never heard of this parameter before and it's unlikely to break anyone.

brandur-stripe:
Thanks Remi.

brandur-stripe:
Alright, just for good hygiene, released this as 27.0.0 (even though it's not likely to break anyone).

Issue #460, opened 2017-09-20
Payment Source creation results in Bad Request


kostaskoukouvis:
The new `form` package does not fully support payment sources. Especially I ran into the problem when trying to create a new `three_d_secure` source.

What the `/v1/sources` endpoint **expects**:
```
type=three_d_secure&three_d_secure[card]=src_66666666&three_d_secure[customer]=cus_66666666
```
What is actually being sent:
```
type=three_d_secure&card=src_6666666&customer=cus_666666666
```

The problem lies to the fact that the `TypeData` struct attribute has a wildcard tag `form:"*"` which results in it being encoded on the same level as the rest of the struct and not as subfields of the type name. I guess this means that there should be a special case in the `structEncoder's encode` function or on some other level that depending on the payment source `Type` it gives the correct key name for the `TypeData` map.

Unless I totally missed something and there already is a proper way of making those requests that hasn't been documented yet.

Go version: 1.8.3

brandur-stripe:
Released as 27.0.1. Thank you for reporting!

PR #461, opened 2017-09-20
Fix Source creation so that TypeData is passed properly

remi-stripe:
The TypeData values are specific to each source type. When the parameters are sent, they have to be passed to the API under the source's type's name so for 3DS you pass `three_d_secure[card]` for example while for Sofort you pass `sofort[country]`.

This uses a custom `AppendTo` for `SourceObjectParams` so that the `TypeData` properties are looped over and encoded with the right level based on the current Source's type.

r? @brandur-stripe 

I've confirmed it works for 3DS now but I was not sure if there was an easier way that looping over `p.TypeData` ourselves.

Fixes https://github.com/stripe/stripe-go/issues/460

### DIFF
--- a/source.go
+++ b/source.go
@@ -191,6 +191,10 @@ type Source struct {
 // AppendTo implements custom encoding logic for SourceObjectParams so that the special
 // "TypeData" value for is sent as the correct parameter based on the Source type
 func (p *SourceObjectParams) AppendTo(body *form.Values, keyParts []string) {
+	if len(p.TypeData) > 0 && len(p.Type) == 0 {
+		panic("You can not fill TypeData if you don't explicitly set Type")
+	}
+
 	for k, vs := range p.TypeData {
 		body.Add(form.FormatKey(append(keyParts, p.Type, k)), vs)
 	}
--- a/source.go
+++ b/source.go
@@ -2,6 +2,8 @@ package stripe
 
 import (
 	"encoding/json"
+
+	"github.com/stripe/stripe-go/form"
 )
 
 // SourceStatus represents the possible statuses of a source object.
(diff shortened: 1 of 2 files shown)


kostaskoukouvis:
Sorry for commenting here and being a bit off-topic probably, but since you touched that file i'll go ahead and take the liberty to ask. Why is the `Amount` in the `SourceObjectParams` struct a `uint64` while in the `Source` struct a `int64`?

remi-stripe:
@kostaskoukouvis Mostly for historical reasons I think. Most amounts should be `uint64` though there are exceptions (invoice items can have a negative amount). We likely chose the wrong type originally and wanted to avoid changing the type silently. I'm mostly guessing though.

brandur-stripe:
@remi-stripe Thanks for the fix!

The one problem I see with this one is that if `Type` is left empty, then you're going to see some really weird encoding like `[][foo]` (an empty string sent to `form.FormatKey` will be treated as `[]`). Should we maybe wrap that `for` loop in a conditional that checks `if Type != ""`?

brandur-stripe:
> Mostly for historical reasons I think. Most amounts should be uint64 though there are exceptions (invoice items can have a negative amount). We likely chose the wrong type originally and wanted to avoid changing the type silently. I'm mostly guessing though.

Yep :/ all legacy. I see that `uint64` probably seemed like a good idea, but we probably should have just standardized on one type like `int64`.

remi-stripe:
> The one problem I see with this one is that if Type is left empty, then you're going to see some really weird encoding like [][foo] (an empty string sent to form.FormatKey will be treated as []). Should we maybe wrap that for loop in a conditional that checks if Type != ""?

Ha nice catch, https://stripe.com/docs/api#create_source says it's required except for shared Sources. But in that case `TypeData` would have to be empty I think.

remi-stripe:
@brandur-stripe  pushed a fix, is this the right approach?

brandur-stripe:
> @brandur-stripe pushed a fix, is this the right approach?

Yeah, it's usually preferable not to panic of course, but it's probably helpful in this sort of case where we find an obvious usage error.

PR #462, opened 2017-09-22
Remove Bitcoin receiver tests and mark package deprecated

brandur:
Like we're doing elsewhere, this removes Bitcoin receiver tests and
marks the package deprecated in favor of using the new sources API.

These tests still work for now, but they'll break the next time we
upgrade stripe-mock which doesn't provide create/update endpoints. I've
left in the get/list tests because those will continue to work for the
time being.

r? @remi-stripe


brandur-stripe:
Thanks Remi!

PR #465, opened 2017-09-26
Add form annotation for `cvc` in `CardParams`

### LINKED ISSUE #464: Card CVC value omitted from form serialization
The CVC value is "usually required" by Stripe (I am testing from within Turkey), yet it is omitted from the CardParams serialization instructions: 

type CardParams struct {
	Params    `form:"*"`
	Account   string `form:"-"`
	Address1  string `form:"address_line1"`
	Address2  string `form:"address_line2"`
	CVC       string `form:"-"`

I believe this should be:
	CVC       string `form:"cvc"`

Without this change I get this error from Stripe when using their 4242 test Visa card:
 {"code":"invalid_cvc","status":402,"message":"You must provide a value for 'cvc'.","param":"cvc","request_id":"req_NsaoYSTvjhsH0L","type":"card_error"}

go version go1.9 linux/amd64
github.com/stripe/stripe-go to v27.0.1.

brandur:
Adds missing form annotation for `cvc` in CardParams. This isn't used
when creating cards directly (that gets passed through a special
`AppendTo*` that already encodes a `cvc`), but is important when
creating a token with a card.

Fixes #464.

### DIFF
--- a/card.go
+++ b/card.go
@@ -39,7 +39,7 @@ type CardParams struct {
 	Account   string `form:"-"`
 	Address1  string `form:"address_line1"`
 	Address2  string `form:"address_line2"`
-	CVC       string `form:"-"`
+	CVC       string `form:"cvc"`
 	City      string `form:"address_city"`
 	Country   string `form:"address_country"`
 	Currency  string `form:"currency"`


brandur-stripe:
I think this one is pretty uncontentious because it's right in the ["create token" example for Go](https://stripe.com/docs/api/go#create_card_token), so going to self-approve since I'm out a little later today.

Issue #466, opened 2017-09-26
Headers not passed in, in possibly multiple cases, preventing connected account interaction


natdm:
In the most current version of the Go library, the headers are not passed in to the requests.

As an example:
https://github.com/stripe/stripe-go/blob/master/product/client.go#L24
The `params *stripe.ProductParams` that are passed in has a `params.Params` that has a field, `StripeAccount`. This is needed to pass in as a header to signify you are dealing with connected accounts.

It looks like a struct is created here:
https://github.com/stripe/stripe-go/blob/master/product/client.go#L24

and passed in here:
https://github.com/stripe/stripe-go/blob/master/product/client.go#L32

If the code on line 26 (https://github.com/stripe/stripe-go/blob/master/product/client.go#L26-L29)
was made to utilize these parameters by doing 

```go
        var commonParams *stripe.Params
        
	if params != nil {
		body = &form.Values{}
		form.AppendTo(body, params)
                commonParams = &params.Params
	}
```

or something similar, they would actually get passed in.

natdm:
Looks like the sku client doesn't do it either?
https://github.com/stripe/stripe-go/blob/master/sku/client.go#L22-L35

natdm:
Thanks @remi-stripe. Do you guys allow PR's to the codebase at all?

remi-stripe:
@natdm yes we definitely do so if you ever spot anything else wrong, submitting a PR would work. Fixing that one was really easy since you had already found the exact issue :)

brandur-stripe:
Fixed released as 28.0.0.

PR #467, opened 2017-09-26
Properly pass the Params parameters for sku and product

remi-stripe:
This properly sends the extra parameters in params.Params for sku and for product
This also adds the ability to pass `ProductParams` to the Get for the product resource

Fixes https://github.com/stripe/stripe-go/issues/466

r? @brandur-stripe


brandur-stripe:
Thanks Remi!

LGTM.

PR #459, opened 2017-09-19
Renaming all properties, structs and methods to match the API naming

remi-stripe:
This PR ensures that all properties in the library matches the official API. Names are not shortened anymore making it easier to associated a given parameter to its API counterpart.

We can't list all the changes one by one as those are extensive but here are the ones likely to impact most integrations:
* All properties with the word url in it now have the same case such as `BusinessURL`.
* All properties with the word four in it now use a digit such as `DynamicLast4` or `SSNLast4`.
* Renamed confusing properties such as `Account` on `BankAccount` becoming `AccountNumber` to avoid confusing with its Connect counterpart which is now named `Account` instead of `AccountID`.
* Properties such as `MonthAnchor` or constants such as `Month` are now renamed to match the corresponding parameter/value in the API `MonthlyAnchor` and `Monthly`.
* `Meta` and `Live` have been renamed to `Metadata` and `Livemode` on all classes
* `Values` on List objects has been renamed to `Data`
* Some classes have been renamed. 
  * `Owner` is now `AdditionalOwner`,
  * `Transaction` is now `BalanceTransaction` with all constants and corresponding classes renamed to match this
  * references to `Tx` were also replaced by `BalanceTransaction` in classes, properties and constants
  * `Fee` and `FeeRefund` have been renamed to `ApplicationFee` and `ApplicationFeeRefund` along with all related classes, properties and constants.
  * `Sub` has been renamed to `Subscription` along with all related classes, properties and constants.
* Card parameters have been renamed such as `City` becoming `AddressCity` and `Month` becoming `ExpMonth.
* Errors codes have been prefixed with `ErrorCode` such as `ErrorCodeCardDeclined` or `ErrorCodeIncorrectZip`.
* Event properties are now named `Object` and `PreviousAttributes` and methods `GetObjectValue()` and `GetPreviousValue()`.
* Pagination properties have been renamed to `StartingAfter`, `EndingBefore` and `HasMore`.
* `AddMeta()` has been renamed to `AddMetadata()`, `Expand()` to `AddExpand()`

PR #468, opened 2017-09-28
Parse zero-val pointers

natdm:
If a boolean value is meant to be `true` by default, `stripe-go` uses a pointer to a bool, and uses `nil` to render no url parameter, which ends up being `true` to the stripe-api.

Unfortunately, there was no ability to pass a pointer to a boolean of `true`. This attempts to fix it in a non-breaking, non-major change or refactor way.

I made some smaller changes as well to assist in the readability (since I had to do a lot of that -- reading).

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -61,6 +61,11 @@ type formOptions struct {
 	// properly encode an explicit 0. It indicates that an explicit zero should
 	// be sent.
 	Zero bool
+
+	// EncodeZeroVal indicates the need to encode the zero value of a pointer.
+	// It is not parsed from the form, it's meant to be passed while encoding
+	// boolean values, to make sure 'false' is not ignored.
+	EncodeZeroVal bool
 }
 
 type structEncoder struct {
@@ -141,20 +146,21 @@ func FormatKey(parts []string) string {
 // ---
 
 func boolEncoder(values *Values, v reflect.Value, keyParts []string, options *formOptions) {
-	if !v.Bool() {
+	if !v.Bool() && !(options != nil && options.EncodeZeroVal) {
 		return
 	}
 
 	if options != nil {
-		if v.Bool() {
-			switch {
-			case options.Empty:
-				values.Add(FormatKey(keyParts), "")
-			case options.Invert:
-				values.Add(FormatKey(keyParts), strconv.FormatBool(false))
-			case options.Zero:
-				values.Add(FormatKey(keyParts), "0")
-			}
+		switch {
+		case options.Empty:
+			values.Add(FormatKey(keyParts), "")
+		case options.Invert:
+			values.Add(FormatKey(keyParts), strconv.FormatBool(false))
+		case options.Zero:
+			values.Add(FormatKey(keyParts), "0")
+		case options.EncodeZeroVal:
+			fmt.Println(strconv.FormatBool(v.Bool()))
+			values.Add(FormatKey(keyParts), strconv.FormatBool(v.Bool()))
 		}
 	} else {
 		values.Add(FormatKey(keyParts), strconv.FormatBool(v.Bool()))
@@ -195,6 +201,9 @@ func buildPtrEncoder(t reflect.Type) encoderFunc {
 	return func(values *Values, v reflect.Value, keyParts []string, options *formOptions) {
 		if v.IsNil() {
 			return
+		}
+		if options == nil {
+			options = &formOptions{EncodeZeroVal: true}
 		}
 		elemF(values, v.Elem(), keyParts, options)
 	}
@@ -357,9 +366,12 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 			continue
 		}
 
+		fldTyp := reflectField.Type
+		fldKind := fldTyp.Kind()
+
 		if Strict && options != nil &&
 			(options.Empty || options.Invert || options.Zero) &&
-			reflectField.Type.Kind() != reflect.Bool {
+			fldKind != reflect.Bool {
 
 			panic(fmt.Sprintf(
 				"Cannot specify `empty`, `invert`, or `zero` for non-boolean field; on: %s/%s",
@@ -367,20 +379,15 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 			))
 		}
 
-		var isPtr bool
-		if reflectField.Type.Kind() == reflect.Ptr {
-			isPtr = true
-		}
-
 		se.fields = append(se.fields, &field{
 			formName:   formName,
 			index:      i,
-			isAppender: isAppender(reflectField.Type),
-			isPtr:      isPtr,
+			isAppender: isAppender(fldTyp),
+			isPtr:      fldKind == reflect.Ptr,
 			options:    options,
 		})
 		se.fieldEncs = append(se.fieldEncs,
-			getCachedOrBuildTypeEncoder(reflectField.Type))
+			getCachedOrBuildTypeEncoder(fldTyp))
 	}
 
 	return se
--- a/form/form.go
+++ b/form/form.go
@@ -27,7 +27,15 @@ type Appender interface {
 	AppendTo(values *Values, keyParts []string)
 }
 
-type encoderFunc func(values *Values, v reflect.Value, keyParts []string, options *formOptions)
+// encoderFunc is used to encode any type from a request.
+//
+// A note about encodeZeroVal:
+// Since some types in the Stripe API are defaulted to non-zero values, and Go defaults types to their zero values,
+// any type that has a Stripe API default of a non-zero value is defined as a go pointer, meaning nil defaults to
+// the Stripe API non-zero value. To override this, a check is made to see if the value is the zero-value for that
+// type. If it is, and encodeZeroVal is true, it's encoded. This is ignored as a parameter when dealing with types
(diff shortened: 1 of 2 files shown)


natdm:
@remi-stripe or someone, since I can't review or assign.

remi-stripe:
@natdm thanks for the PR! Don't worry it pings the whole team whenever one is created. I'll leave this one to Brandur (assigning him) as it's about the internals of `form.go`

brandur-stripe:
@natdm Thanks and sorry about the trouble here!

This patch is looking pretty good. Looking through this though, I think it might be a good idea to shift the behavior a little more broadly because although booleans are _mostly_ what was broken here, it looks like they weren't the only thing. `Quantity` on `OrderItemParams` would have previously allowed a zero, but that will also no longer be encoded.

What do you think about tweaking this slightly so that _if_ a field is a pointer and _if_ that pointer has a kind that's a scalar value (i.e. bool, int, string, unit, float, etc., _not_ array, interface, struct), we encode the value even if it's zero value? This would address `Quantity` as well, and I think the change would be a little nicer.

I realize this is a fairly tall order though, so if you like I can take over the branch and work it through to completion. Thoughts?

brandur-stripe:
@remi-stripe BTW, I'd totally overlooked that we had a few places that do use pointers to encode params information. I think what happened is that as the Relay stuff was going in, it kind of deviated from the normal conventions randomly because no one was familiar enough with the kit at the time to let the author know about it.

This does represent some precedent though ... since we've got both types of use, it might be worthwhile considering a pointer-based approach elsewhere in the library too (e.g. `Prorate *bool` instead of `Prorate bool` + `ProrateNo bool`). We'd probably want to add some helpers as well (e.g. `func Bool(b bool) *bool {`) to make usage more convenient than it is now, but it's plausible at least.

natdm:
@brandur-stripe I noticed that too, and I appreciate the opportunity to take a stab at it. I'll look tonight and see what I can come up with.

brandur-stripe:
> I noticed that too, and I appreciate the opportunity to take a stab at it. I'll look tonight and see what I can come up with.

Thank you! Let me know if you want any other kind of support.

remi-stripe:
@brandur 
> This does represent some precedent though ... since we've got both types of use, it might be worthwhile considering a pointer-based approach elsewhere in the library too (e.g. Prorate *bool instead of Prorate bool + ProrateNo bool). We'd probably want to add some helpers as well (e.g. func Bool(b bool) *bool {) to make usage more convenient than it is now, but it's plausible at least.

Agreed that's what I was trying to raise internally on deciding for the right approach on the pointer. You pointed out that passing the value in code becomes harder this way though which rang true.

I'm happy to take any approach though I agree that consolidating this once and for all across the library would be valuable. The `ProrateNo` option is not the greatest but I'm not sure that the `Prorate *bool` is cleaner Go-wise. Do you have a personal preference?

brandur-stripe:
> Agreed that's what I was trying to raise internally on deciding for the right approach on the pointer. You pointed out that passing the value in code becomes harder this way though which rang true.

Ah sorry. I may not have been reading carefully enough — I didn't realize at the time that we were already doing pointer stuff in some places.

> I'm happy to take any approach though I agree that consolidating this once and for all across the library would be valuable. The ProrateNo option is not the greatest but I'm not sure that the Prorate *bool is cleaner Go-wise. Do you have a personal preference?

That's where it gets a little harder. Honestly, if we were doing this from scratch, I'd absolutely just say: pointers.

However, given that we have quite a few existing installations, we should try to minimize upgrade churn so it's not obscenely painful.

I just did some accounting and right now we have:

* 11 `form:invert` fields for booleans (e.g. `NoProrate bool`).
* 4 `form:empty` fields for strings (e.g. `CouponEmpty bool`).
* 11 `form:zero` fields for integers (e.g. `QuantityZero bool).

If we switched to pointers, we could get rid of all of these and they'd become just `x *bool`, `x *string`, and `x *integer`. We could also remove the `form` special casing, and we could also get everything on the same system (as opposed to Relay parameters doing their own thing).

All in all, I'm pretty on the fence right now. In total there are ~600 parameter fields in the library, so changing these would affect ~5-10%. It'd be painful for existing users, but probably an improvement for future users. What do you think?

remi-stripe:
@brandur-stripe with the scope of https://github.com/stripe/stripe-go/pull/459 and how widely it changes struct and params names I'm assuming those would be as painful (or as straightforward) to change as the rest for users right?

As in, as long as we tackle this in the same major version it's one big hurdle (for the best) and then the library is cleaner?

If so, I'm definitely in favor of changing this

brandur-stripe:
> If so, I'm definitely in favor of changing this

Cool. Yeah I can get behind that.

Just for inspiration, here's how the [AWS Golang SDK]() looks with an invocation (note `aws.String`):

``` go
// Uploads the object to S3. The Context will interrupt the request if the
// timeout expires.
_, err := svc.PutObjectWithContext(ctx, &s3.PutObjectInput{
	Bucket: aws.String(bucket),
	Key:    aws.String(key),
	Body:   os.Stdin,
 })
```

The first [Go-based GitHub SDK to appear on Google](https://github.com/google/go-github) also does something very similar (note `github.String`):

``` go
// create a new private repository named "foo"
repo := &github.Repository{
	Name:    github.String("foo"),
	Private: github.Bool(true),
}
client.Repositories.Create(ctx, "", repo)
```

I think people are relatively used to this pattern in Go and could probably easily get behind it.

natdm:
For what it's worth, here at Brightcove, we use the AWS libraries for everything and use `aws.Bool` (or what have you) everywhere. I also use it with Stripe for dealing with these issues.

How I found this particular bug: 
```go
// CreateProduct creates a product based on a database item.
func (b *Billing) CreateProduct(acct string, i database.Item) (*stripe.Product, error) {
	meta := make(map[string]string)
	var desc string
	if i.Description != nil {
		desc = *i.Description
	}
	meta["event_id"] = strconv.Itoa(i.EventID)
	params := stripe.ProductParams{Name: i.Name, Desc: desc, Shippable: aws.Bool(false)}
	params.Meta = meta
	params.SetStripeAccount(acct)
	return product.New(&params)
}
```

brandur-stripe:
> For what it's worth, here at Brightcove, we use the AWS libraries for everything and use aws.Bool (or what have you) everywhere. I also use it with Stripe for dealing with these issues.

Ah, good to know! Well, we should probably be shipping our own, but there's no particular reason that the `aws` package couldn't also do the job.

natdm:
Absolutely. Just pointing out that I believe you're correct -- it's an adopted pattern to use a helper library of sorts to create boolean values.

natdm:
Ok, took a quick stab at it.

I added a lot of `{"sometype_ptr", &testStruct{}, ""}` tests, mainly for documentation, although doing it once would  suffice. I can remove them if you wish.

natdm:
@brandur-stripe - Let me know how you would do this to clean it up. Feel free to branch the repo or create another PR, no feelings hurt -- I don't feel that this is as clean as it can get, and I'm not sure how far I should be going with it.

form/form.go
```diff
@@ -357,30 +378,28 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 			continue
 		}
 
+		fldTyp := reflectField.Type
```
brandur-stripe: It looks like "typ" for types is the abbreviation used in Go core's source code too. Nice!

form/form.go
```diff
@@ -357,30 +378,28 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
...
-			isPtr = true
-		}
-
 		se.fields = append(se.fields, &field{
 			formName:   formName,
 			index:      i,
-			isAppender: isAppender(reflectField.Type),
-			isPtr:      isPtr,
+			isAppender: isAppender(fldTyp),
+			isPtr:      fldKind == reflect.Ptr,
```
brandur-stripe: Nice improvement. Somehow I'd convinced myself that you weren't allowed to specify an expression like this.

form/form_test.go
```diff
@@ -128,21 +128,30 @@ func BenchmarkAppendTo(b *testing.B) {
 func TestAppendTo(t *testing.T) {
 	var arrayVal = [3]string{"1", "2", "3"}
 
-	var boolVal = true
+	var boolValT = true
+	var boolValF = false
 
 	var float32Val float32 = 1.2345
+	var float32Val0 float32
```
brandur-stripe: I love the `*0` convention. Thanks for the considerable number of new tests too!

natdm:
Ok, @brandur-stripe, I took a go at it. It got a little tricky but I was able to do it without touching the tests. I also added a note to the `encoderFunc` type since I think the explanation was necessary, and that seemed to be the only place for it.

I agree it didn't seem like it belonged in `formOptions`. The only reason I stuck it there was because it wasn't called `formTags`, so I convinced myself it could be allowed to collect more information. :-X

brandur-stripe:
Nice one @natdm! Thanks again for the hard work on this one.

natdm:
Glad to commit to one of my favorite products to work with.

brandur-stripe:
> Glad to commit to one of my favorite products to work with.

Just glad you're still happy after running into all these bugs :)

Released as 28.0.1!

PR #473, opened 2017-10-09
Adds support for detaching sources from customers

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @stan-stripe 

Adds support for detaching sources from customers.

### DIFF
--- a/source/client.go
+++ b/source/client.go
@@ -1,6 +1,9 @@
 package source
 
 import (
+	"errors"
+	"fmt"
+
 	stripe "github.com/stripe/stripe-go"
 	"github.com/stripe/stripe-go/form"
 )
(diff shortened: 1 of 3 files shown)


source/client.go
```diff
@@ -83,3 +86,30 @@ func (c Client) Update(id string, params *stripe.SourceObjectParams) (*stripe.So
...
+		commonParams = &params.Params
+		body = &form.Values{}
+		form.AppendTo(body, params)
+	}
+
+	source := &stripe.Source{}
+	var err error
+
+	if len(params.Customer) > 0 {
+		err = c.B.Call("DELETE", fmt.Sprintf("/customers/%v/sources/%v", params.Customer, id), c.Key, body, commonParams, source)
```
remi-stripe: Since `SourceObjectParams` has it defined as `Customer string             `form:"customer"`` won't the parameter be posted anyway to the API? I guess you might need to create a `SourceDetachParams` instead to avoid this issue.

ob-stripe: Good point, though that probably also applies to the `Del` method on cards and bank accounts then. I'll check!

ob-stripe: Okay, you were entirely correct. Cards and bank accounts aren't affected since `Customer` has `form:"-"` in their respective structures.

Fixed by defining a new `SourceObjectDetachParams` struct.

brandur-stripe:
LGTM. Thanks guys!

brandur-stripe:
Released as 28.2.0.

PR #484, opened 2017-10-23
Error early when params not specified for card functions

### LINKED ISSUE #483: Card client panics on nil CardParams
https://github.com/stripe/stripe-go/blob/master/card/client.go#L86

If CardParams cannot be nil, there should be a check and an error.

brandur:
Card-related API calls are a bit of an unusual case because we need to
have an instantiated params struct in order to determine the correct URL
to which to make the request. Previously, if a `nil` was passed to any
of these functions, we'd panic as we tried to access one of its members.

This patch just makes the params requirement a little more explicit by
returning an error if any of card functions is called with `nil` params.

Fixes #483.

r? @remi-stripe

### DIFF
--- a/card/client.go
+++ b/card/client.go
@@ -41,6 +41,10 @@ func New(params *stripe.CardParams) (*stripe.Card, error) {
 }
 
 func (c Client) New(params *stripe.CardParams) (*stripe.Card, error) {
+	if params == nil {
+		return nil, errors.New("params should not be nil")
+	}
+
 	body := &form.Values{}
 
 	// Note that we call this special append method instead of the standard one
@@ -71,6 +75,10 @@ func Get(id string, params *stripe.CardParams) (*stripe.Card, error) {
 }
 
 func (c Client) Get(id string, params *stripe.CardParams) (*stripe.Card, error) {
+	if params == nil {
+		return nil, errors.New("params should not be nil")
+	}
+
 	var body *form.Values
 	var commonParams *stripe.Params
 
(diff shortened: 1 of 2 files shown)


card/client.go
```diff
@@ -129,6 +141,10 @@ func Del(id string, params *stripe.CardParams) (*stripe.Card, error) {
 }
 
 func (c Client) Del(id string, params *stripe.CardParams) (*stripe.Card, error) {
+	if params == nil {
```
remi-stripe: We're already doing `if params != nil {` below (not visible) so it's redundant. Should you fix the one below in that case?

card/client.go
```diff
@@ -166,14 +182,20 @@ func (c Client) List(params *stripe.CardListParams) *Iter {
...
+		form.AppendTo(body, params)
+		lp = &params.ListParams
+		p = params.ToParams()
+	}
 
 	return &Iter{stripe.GetIter(lp, body, func(b *form.Values) ([]interface{}, stripe.ListMeta, error) {
 		list := &stripe.CardList{}
 		var err error
 
+		if params == nil {
```
remi-stripe: Why not just error above like you do in the other functions `if params == nil`?

card/client.go
brandur-stripe: Yeah good point. Unfortunately the other one would still result in a panic because of the check on `params.Account` immediately afterwards, so I've taken that one out in favor of my addition.

card/client.go
brandur-stripe: Ah, basically because `List` functions only return an `*Iter` (and not an error). The first opportunity we get to return an error is the first time the iterator iterates, and so the check on `params` is further down than you'd expect.

remi-stripe: ah gotcha, that makes sense, thanks for clarifying!

brandur-stripe:
Thanks Remi. Released as 28.4.1.

Issue #485, opened 2017-10-25
Issues when passing bank account details directly in external account creations


ob-stripe:
cc @thorsten-stripe @gh-stripe 

@thorsten-stripe noticed a couple of issues when trying to add an external account to a custom account:

1. When using `ExternalAccountParams` and providing the bank account details directly, no `object` parameter is sent, causing the API request to fail.

2. When using `BankAccountParams` (as shown in the API reference [here](https://stripe.com/docs/api/go#account_create_bank_account)) and providing the bank account details directly, the library will send empty values for `account_holder_name` and `account_holder_type` if they are not provided. This causes the API request to fail with ""The bank account type you have added is invalid. You must pass either 'company' or 'individual'.""

    These parameters are only required when creating a bank account for payins (i.e. ACH). They are optional when creating a bank account as an external payouts account.

brandur-stripe:
Totally missed that Remi was already working on this and had assigned it to himself. Sorry :/

brandur-stripe:
#487 squashed the rest of this one. Closing.

PR #486, opened 2017-10-25
Send `object=bank_account` when using `AccountExternalAccountParams`

brandur:
When using `AccountExternalAccountParams` for creating a bank account
from an account, sending an `object` field is required. Here we give the
params struct a custom `AppendTo` so that we can comply with this logic.

Partial fix for #485.

r? @ob-stripe

### DIFF
--- a/account.go
+++ b/account.go
@@ -104,9 +104,13 @@ type AccountExternalAccountParams struct {
 
 // AppendTo implements custom encoding logic for AccountExternalAccountParams
 // so that we can send the special required `object` field up along with the
-// other specified parameters.
+// other specified parameters or the token value
 func (p *AccountExternalAccountParams) AppendTo(body *form.Values, keyParts []string) {
-	body.Add(form.FormatKey(append(keyParts, "object")), "bank_account")
+	if len(p.Token) > 0 {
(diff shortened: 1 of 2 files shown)


ob-stripe:
lgtm in the specific context of #485, but ideally we should also be able to handle the following cases:
- sending a token in `external_account` instead of bank account details
- sending card details instead of bank account details

At the moment I think the library is not capable of either of these. This can be handled in a future PR though.

remi-stripe:
> sending card details instead of bank account details

We explicitly refuse to support this one since it's not PCI compliant and would be blocked for almost all users by default.

remi-stripe:
Worked with @brandur-stripe on this and pushed a fix for the token part so that you can send it as needed.

ob-stripe:
Nice! Mind adding a test for the token case?

remi-stripe:
I looked unfortunately I don't understand the mock logic for this test case. I get a `panic` that does not make sense to me since the code does work as expected. Will defer to @brandur-stripe

brandur-stripe:
Good call on the additional functionality @ob-stripe, and thanks for the patch @remi-stripe!

Added an additional test in f7de6b1 and am bringing this one in.

PR #487, opened 2017-10-25
Make `account_holder_name`/`account_holder_type` truly optional

brandur:
When encoding a bank account we shouldn't send `account_holder_name` or
`account_holder_type` unless they're given values. Sending an empty
value will cause an API error in cases where the request would have
otherwise been valid.

Fixes the other part of #485.

r? @remi-stripe


brandur-stripe:
Thanks Remi.

PR #488, opened 2017-10-26
Add support for listing source transactions

ob-stripe:
r? @brandur-stripe
cc @stripe/api-libraries @stan-stripe

Adds support for the `/v1/sources/src_.../source_transactions` endpoint.

I wrote a test, but as stripe-mock doesn't support this endpoint yet the test is skipped for the time being. Not great, but it doesn't seem like we have an easy way to stubbing requests in stripe-go tests.

### DIFF
--- a/sourcetransaction.go
+++ b/sourcetransaction.go
@@ -1,0 +1,55 @@
+package stripe
+
+import "encoding/json"
+
+// SourceTransactionListParams is the set of parameters that can be used when listing SourceTransactions.
+type SourceTransactionListParams struct {
+	ListParams `form:"*"`
+	Source     string `form:"-"` // Sent in with the URL
+}
+
+// SourceTransactionList is a list object for SourceTransactions.
+type SourceTransactionList struct {
+	ListMeta
+	Values []*SourceTransaction `json:"data"`
+}
+
+// SourceTransaction is the resource representing a Stripe source transaction.
+type SourceTransaction struct {
+	Amount       int64    `json:"amount"`
+	Created      int64    `json:"created"`
+	Currency     Currency `json:"currency"`
+	CustomerData string   `json:"customer_data"`
+	ID           string   `json:"id"`
+	Live         bool     `json:"livemode"`
+	Source       string   `json:"source"`
+	Type         string   `json:"type"`
+	TypeData     map[string]interface{}
+}
+
+// UnmarshalJSON handles deserialization of a SourceTransaction. This custom
+// unmarshaling is needed to extract the type specific data (accessible under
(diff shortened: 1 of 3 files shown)


brandur-stripe:
@ob-stripe Can you double check the test code? Even if it's not run, it's still compiled, and I think the new test has a couple problems:

```
$ make
go test -race ./...
# github.com/stripe/stripe-go
./sourcetransaction.go:35:9: undefined: json
./sourcetransaction.go:39:3: undefined: t
./sourcetransaction.go:42:8: undefined: json
./sourcetransaction.go:46:18: undefined: t
./sourcetransaction.go:48:4: undefined: t
# github.com/stripe/stripe-go
./sourcetransaction.go:35:9: undefined: json
./sourcetransaction.go:39:3: undefined: t
./sourcetransaction.go:42:8: undefined: json
./sourcetransaction.go:46:18: undefined: t
./sourcetransaction.go:48:4: undefined: t
```

sourcetransaction.go
```diff
@@ -0,0 +1,53 @@
...
+type SourceTransaction struct {
+	Amount       int64    `json:"amount"`
+	Created      int64    `json:"created"`
+	Currency     Currency `json:"currency"`
+	CustomerData string   `json:"customer_data"`
+	ID           string   `json:"id"`
+	Live         bool     `json:"livemode"`
+	Source       string   `json:"source"`
+	Type         string   `json:"type"`
+	TypeData            map[string]string  `form:"-"`
```
brandur-stripe: We shouldn't be mixing `json` and `form` annotations (this might have been a bad copy + paste). It also looks like you'll have to run a `gofmt` on this.

sourcetransaction/client.go
```diff
@@ -0,0 +1,63 @@
...
+
+	if params != nil {
+		body = &form.Values{}
+		form.AppendTo(body, params)
+		lp = &params.ListParams
+		p = params.ToParams()
+	}
+
+	return &Iter{stripe.GetIter(lp, body, func(b *form.Values) ([]interface{}, stripe.ListMeta, error) {
+		list := &stripe.SourceTransactionList{}
```
brandur-stripe: It's probably not convention elsewhere, but it might be worth failing fast here by returning an error if `params.Source` wasn't set. Currently I believe it'll fail with kind of a confusing `couldn't find URL /sources//source_transactions` 404 (or something of the sort).

brandur-stripe:
@ob-stripe Left a couple comments above, but thanks!

ob-stripe:
Not sure how I managed to commit a file in such a borked state, but it should be all fixed now. I also added a check for `params.Source` as requested.

PR #492, opened 2017-11-02
Rename language fields in Stripe's user agent to start with `lang`

brandur:
I noticed this as I was looking at a dashboard that @ob-stripe produced today.
Other languages send language and language version as `lang` and `lang_version`
while we were using the fully expanded `language_*` here. I'm changing it here
for consistency with other libraries, but I'm also going to patch the server
because there are so many Go libraries out there that could potentially be
reporting if we supported both styles.

r? @ob-stripe

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -111,8 +111,8 @@ type Backends struct {
 type stripeClientUserAgent struct {
 	Application     *AppInfo `json:"application"`
 	BindingsVersion string   `json:"bindings_version"`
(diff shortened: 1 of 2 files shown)


ob-stripe:
Oh, nice find! I actually checked in Splunk that the library was sending the values, but failed to notice the `lang` vs `language` difference.

lgtm 👍

brandur-stripe:
Thanks!

Issue #499, opened 2017-12-12
Direct Charge on Platform Account Customer to Connected Account Not working


anshul35:
Stripe-go version : 2568e0f2a1ab55fa486a10e8d6797d2dce0c70b5

**Scenario**

Platform Account: A
Connected Account: B
Customer Registered to A: X

**Expectation**:

Charge Customer X on his default source and and put the money in account B.

**Error** 

```
{\"status\":400,\"message\":\"You provided a customer without specifying a source. The default source of the customer is a source and cannot be shared from existing customers.\\n\",\"request_id\":\"req_rqGRipNZaiDJbM\",\"type\":\"invalid_request_error\"}
```

**Code**
```
**CreateToken**:
        stripe.Key = config.STRIPE_CLIENT_SECRET
	params := &stripe.TokenParams{
		Customer: stripeCustomerId,
	}
	params.SetStripeAccount(connectAccountId)
	return token.New(params)

**CreateChargeParams**:
      params := stripe.ChargeParams{
		Currency: "usd",		
		Amount: uint64(userBill.Amount * 100), //Convert the amount in cents, lowest unit of USD
		Fee: 0, 
	}
	params.SetStripeAccount(merchantStripeInfo.AccountID)

	//Set source
	token, err := CreateToken(stripeCustomer.ID, connectedAccountId)
	if err != nil {
		return nil, err
	}
	params.SetSource(token.ID)

**Charge**:
       chargeParams, err := CreateChargeParams(userBill, stripeCustomer)
	if err != nil {
		logger.Errorf("Unable to create charge params for userbill for customer ID = %s with error = %s",
			stripeCustomer.ID, err.Error())
		return err
	}
	//Try with the default source
	ch, err := charge.New(chargeParams)
```

ob-stripe:
Hi @anshul35, thanks for reporting this!

In order to be able to share a card source, the API requires that the source ID be explicitly passed in the `card` parameter, even if the card source is the customer's default source.

I think the Go client library does not support specifying this parameter, but I could be mistaken. cc @brandur-stripe @remi-stripe, can one of you confirm this is the case?

remi-stripe:
So there are 2 issues at hand here. First, we don't support passing `card="card_XXXX"` when creating a Token which is what @ob-stripe is talking about. Second, we don't support cloning a Source which is the API described [here](https://stripe.com/docs/sources/connect#shared-card-sources) and what @anshul35 needs right now.

Going to open a PR in a few minutes.

[...]

PR #500, opened 2017-12-12
Support source sharing via Stripe Connect

remi-stripe:
When using Stripe Connect, you can clone a card/bank account/source form the platform to a connected account. This PR adds support for all of this:
- Creating a new token on a connected account passing the customer id in `customer` and the card id in `card`.
- Creating a new token on a connected account passing the customer id in `customer` and the bank account id in `bank_account`.
- Creating a new Source on a connected account passing the customer id in `customer` and the Source id in `original_source`.

This fixes https://github.com/stripe/stripe-go/issues/499

r? @brandur-stripe

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -37,6 +37,9 @@ type BankAccountParams struct {
 	// Token is a token referencing an external account like one returned from
 	// Stripe.js.
 	Token string `form:"-"`
+
+	// ID is used when tokenizing a bank account for shared customers
+	ID string `form:"*"`
 }
 
 // AppendToAsSourceOrExternalAccount appends the given BankAccountParams as
--- a/card.go
+++ b/card.go
@@ -53,6 +53,9 @@ type CardParams struct {
 	Token     string `form:"-"`
 	Year      string `form:"exp_year"`
 	Zip       string `form:"address_zip"`
+
+	// ID is used when tokenizing a card for shared customers
+	ID string `form:"*"`
 }
 
 // AppendToAsCardSourceOrExternalAccount appends the given CardParams as either a
--- a/source.go
+++ b/source.go
@@ -81,6 +81,7 @@ type SourceObjectParams struct {
 	Currency            Currency           `form:"currency"`
 	Customer            string             `form:"customer"`
 	Flow                SourceFlow         `form:"flow"`
+	OriginalSource      string             `form:"original_source"`
 	Owner               *SourceOwnerParams `form:"owner"`
 	Redirect            *RedirectParams    `form:"redirect"`
 	StatementDescriptor string             `form:"statement_descriptor"`
--- a/source/client_test.go
+++ b/source/client_test.go
@@ -45,3 +45,16 @@ func TestSourceDetach(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, source)
 }
+
+func TestSourceSharing(t *testing.T) {
+	params := &stripe.SourceObjectParams{
+		Type:           "card",
+		Customer:       "cus_123",
+		OriginalSource: "src_123",
+		Usage:          stripe.UsageReusable,
+	}
+	params.SetStripeAccount("acct_123")
+	source, err := New(params)
+	assert.Nil(t, err)
+	assert.NotNil(t, source)
+}
--- a/token/client_test.go
+++ b/token/client_test.go
@@ -47,3 +47,29 @@ func TestTokenNew_WithPII(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, token)
 }
+
+func TestTokenNew_SharedCustomerCard(t *testing.T) {
+	params := &stripe.TokenParams{
+		Card: &stripe.CardParams{
+			ID: "card_123",
+		},
+		Customer: "cus_123",
+	}
+	params.SetStripeAccount("acct_123")
+	token, err := New(params)
+	assert.Nil(t, err)
+	assert.NotNil(t, token)
+}
+
+func TestTokenNew_SharedCustomerBankAccount(t *testing.T) {
+	params := &stripe.TokenParams{
+		Bank: &stripe.BankAccountParams{
+			ID: "ba_123",
+		},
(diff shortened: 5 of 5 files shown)


token.go
```diff
@@ -7,10 +7,12 @@ type TokenType string
...
 type TokenParams struct {
-	Params   `form:"*"`
-	Bank     *BankAccountParams `form:"bank_account"`
-	Card     *CardParams        `form:"card"`
-	Customer string             `form:"customer"`
+	Params        `form:"*"`
+	Bank          *BankAccountParams `form:"bank_account"`
+	BankAccountId string             `form:"bank_account"`
+	Card          *CardParams        `form:"card"`
+	CardId        string             `form:"card"`
```
brandur-stripe: The addition of `BankAccountId` and `CardId` go a little off-road compared to our standard conventions. Generally when referencing an ID, you just do it with the name of the target resource (e.g. passing `Plan` when creating a subscription).

I wonder if we should add an `ID` field into the existing `BankAccountParams` and `CardParams` instead? Is the reason you didn't try that because the encoding implementation on these two types of params so messed up already?

remi-stripe:
@brandur-stripe okay I took your approach instead. PTAL

bankaccount.go
```diff
@@ -37,6 +37,9 @@ type BankAccountParams struct {
 	// Token is a token referencing an external account like one returned from
 	// Stripe.js.
 	Token string `form:"-"`
+
+	// ID is used when tokenizing a bank account for shared customers
+	ID string `form:"*"`
```
brandur-stripe: Do know if the `*` is required here? If we just need it for `AppendToAsSourceOrExternalAccount`, it might be worth changing to `-` to indicate that it's never needed elsewhere.

bankaccount.go
```diff
@@ -68,6 +71,8 @@ func (a *BankAccountParams) AppendToAsSourceOrExternalAccount(body *form.Values)
 		if a.Default {
 			body.Add("default_for_currency", strconv.FormatBool(a.Default))
 		}
+	} else if len(a.ID) > 0 {
+		body.Add("bank_account", a.ID)
```
brandur-stripe: Would you mind adding a quick test in `bankaccount_test.go` (there should a few for the params in there already) to double-check this behavior? One of the reasons that we had a few sources-related regressions during the `form` changeover was that their parameters tend to have pretty complicated encoding logic, but didn't really have any good testing around it.

token.go
```diff
@@ -7,10 +7,11 @@ type TokenType string
 // TokenParams is the set of parameters that can be used when creating a token.
 // For more details see https://stripe.com/docs/api#create_card_token and https://stripe.com/docs/api#create_bank_account_token.
 type TokenParams struct {
-	Params   `form:"*"`
-	Bank     *BankAccountParams `form:"bank_account"`
-	Card     *CardParams        `form:"card"`
-	Customer string             `form:"customer"`
+	Params        `form:"*"`
+	Bank          *BankAccountParams `form:"bank_account"`
+	BankAccountId string             `form:"bank_account"`
```
brandur-stripe: (I think the intention here was probably to remove `BankAccountId` as well as `CardId`.

brandur-stripe:
Thanks Remi! Left a couple more comments, but this is looking like it's getting to a very good place.

bankaccount.go
remi-stripe: Hmmm isn't the test in `token/client_test.go` enough? What do you want me to test?

bankaccount.go
remi-stripe: @brahn-stripe with `-` the field seems ignored. Not sure if I'm making another mistake but it's not sent to stripe-mock in that case

remi-stripe: sorry that was for @brandur-stripe

token.go
remi-stripe: correct, missed that one

bankaccount.go
brandur-stripe: Ah okay, this was more of a question I guess. If you look at `BankAccountParams`, it's basically serialized in two separate ways:

1. When it's used for `bankaccount` or `token` operations, the default `form` implementation is used.
2. In certain cases (e.g. recipients), we call the special `AppendToAsSourceOrExternalAccount` function.

I just want to make sure that we're implementing the minimal logic that we need so that we don't make this anymore complicated. If we need the new `ID` field for just (1) above, then we should remove the changes in `AppendToAsSourceOrExternalAccount`. If we need it for just (2) (I don't think this is the case because you implemented a new test for `token`), then we should have just `form:"-"` to indicate that this field is never needed by `form`. If we need `ID` to be encoded for both (1) and (2), then we should leave this as is.

Can you just confirm that the latter is the case?

bankaccount.go
brandur-stripe: When implementing custom encoders like `AppendToAsSourceOrExternalAccount`, it's just kind of nice to know that we're encoding as we expect. `token/client_test.go` is something, but stripe-mock's validations are nowhere near as strict as the real API's, so we can't guarantee that we didn't make an encoding mistake that was allowed through.

There are some encoding tests already in `bankaccount_test.go`, but I was just thinking something like this:

``` go
func TestBankAccountParams_AppendToAsSourceOrExternalAccount(t *testing.T) {
	// Sets ID
	{
		params := &BankAccountParams{ID: "ba_123"}
		body := &form.Values{}
		params.AppendToAsSourceOrExternalAccount(body)
		t.Logf("body = %+v", body)
		assert.Equal(t, []string{"ba_123"}, body.Get("bank_account"))
	}

...
}
```

bankaccount.go
remi-stripe: @brandur-stripe My understanding was that we call `AppendToAsSourceOrExternalAccount` even for Token but I see now that this is wrong and I did not need to add the if inside the custom append. I will fix

remi-stripe:
@brandur-stripe  Okay I think I did this one properly this time! PTAL

brandur-stripe:
LGTM. Thanks!

PR #501, opened 2017-12-13
Add support for coveralls.io

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Adds support for coveralls.io to the Go library.

This requires testing each package separately, because Go cannot create coverage profiles when testing multiple packages (cf. https://github.com/golang/go/issues/6909). We then use [`gover`](https://github.com/sozorogami/gover) to merge the multiple coverage profiles in a single file and [`goveralls`](https://github.com/mattn/goveralls) to send the data to coveralls.io.

I'm annoyed that this requires running the test suite a second time, but can't find a way around it. It looks like the Go issue is fixed in 1.10, so hopefully at some point we will be able to just run the test suite once with `go test -race -covermode=atomic -coverprofile=cover.out`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .env
 *.test
+*.coverprofile
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
 before_install:
   - go get -u github.com/stretchr/testify/require
 
+  # Install code coverage / coveralls tooling
+  - go get -u golang.org/x/tools/cmd/cover
+  - go get -u github.com/modocache/gover
+  - go get -u github.com/mattn/goveralls
+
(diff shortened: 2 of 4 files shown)


brandur-stripe:
Nice! Would you mind just putting a TODO-like (maybe not a literal TODO, but a note that thes could be improved) comment somewhere (maybe the `Makefile`?) to the same effect as you've put in this pull request description? I'm a little afraid that it's going to be easy to forget that we're running the test suite twice and by the time Go 1.10 is our minimum supported version, we'll have long forgotten that we can fix this.

ob-stripe:
Good point. Done!

brandur-stripe:
Thanks OB. Looks great!

PR #506, opened 2018-01-05
Support slices in `event.GetObjValue` and add tests

### LINKED ISSUE #503: Event.GetObjValue cannot handle arrays
Hi, I've spoken with Stripe support and the IRC channel over the course of a couple days, and we've come to the conclusion that it's not a bug, just not implemented yet. If that's not the case please let me know otherwise!

On the invoice created event, I'm trying to access the plan ID, but am having trouble because I'm not sure how to drill down into an array with `GetObjValue`. `planID := event.GetObjValue("lines", "data[0]", "plan", "id")` would be great, or `"data", "0"`. Upon more research I've found that not every object in `data` will have a `plan` key, but I think this is a useful feature regardless.

Let me know if you have any plans to implement this, or if there is a better way to go about dealing with arrays located within events.

Thanks!

brandur:
Modifies `event.GetObjValue` so that when specifying additional keys to
descend into, it's now possible to specify integer keys if the target
object is a slice of type `[]interface{}` (which should be what anything
coming out of event data deserializes to).

I initially expressed some concern about expanding this interface, but
given that the previous behavior would have been an outright panic, I
changed my mind and don't think it's too bad to add support for this
given that it's been requested by a user.

Fixes #503.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/event_test.go
+++ b/event_test.go
@@ -1,0 +1,55 @@
+package stripe
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestGetObjValue(t *testing.T) {
+	event := &Event{
+		Data: &EventData{
+			Obj: map[string]interface{}{
+				"top_level_key": "top_level",
+				"integer_key":   123,
+				"map": map[string]interface{}{
+					"nested_key": "nested",
+				},
+				"slice": []interface{}{
+					"index-0",
+					"index-1",
+					"index-2",
+				},
+				"slice_of_maps": []interface{}{
+					map[string]interface{}{
+						"slice_nested_key": "slice_nested",
+					},
+				},
+			},
+		},
+	}
+
+	assert.Equal(t, "top_level", event.GetObjValue("top_level_key"))
+
+	// Check that it coerces non-string values into strings (this behavior is
+	// somewhat questionable, but I'm going with how it already works)
+	assert.Equal(t, "123", event.GetObjValue("integer_key"))
+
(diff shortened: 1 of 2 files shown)


event_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// By design a `nil` just returns an empty string
+	assert.Equal(t, "", event.GetObjValue("bad_key"))
+
+	// Panic conditions. Usually the function tries to just return a value is
+	// fairly forgiving, but it does panic under certain obviously impossible
+	// cases.
+	assert.Panicsf(t, func() {
+		event.GetObjValue("slice", "string_key")
+	}, "Cannot access nested slice element with non-integer key: %s", "bad_key")
```
ob-stripe: I'm likely missing something since the tests are passing, but shouldn't the assert expect `"string_key"`, not `"bad_key"`?

ob-stripe:
I left a question in a comment, but it's likely just me missing something. lgtm!

(On an unrelated note, it looks like tests are failing on go-tip.)

brandur-stripe:
Thanks OB. I'm glad you noticed that testing problem — it turns out that what I really wanted was `PanicsWithValue` instead of `Panicsf`. In `Panicsf` the message and args are just for display when the assertion fails and will have effect on the expected value. I just pushed a fix.

brandur-stripe:
And good catch on go-tip. I think the patch in https://github.com/stripe/stripe-go/pull/508 will fix it.

brandur-stripe:
Released as 28.9.0.

PR #507, opened 2018-01-07
Move all booleans to boolean pointers

remi-stripe:
This change will allow us to remove all the custom work we have around boolean as parameters in the library. Instead of having `Prorate` and `NoProrate` we will simply have `Prorate` as a boolean pointer that is explicitly set as needed. 

This will be merged as the same time as https://github.com/stripe/stripe-go/pull/459 to ensure that we only have one large breaking change for developers to work around.

cc @stripe/api-libraries

### DIFF
--- a/account.go
+++ b/account.go
@@ -62,25 +62,107 @@ const (
 // AccountParams are the parameters allowed during account creation/updates.
 type AccountParams struct {
 	Params                    `form:"*"`
-	BusinessName              string                        `form:"business_name"`
-	BusinessPrimaryColor      string                        `form:"business_primary_color"`
-	BusinessURL               string                        `form:"business_url"`
-	Country                   string                        `form:"country"`
+	BusinessName              *string                       `form:"business_name"`
+	BusinessPrimaryColor      *string                       `form:"business_primary_color"`
+	BusinessURL               *string                       `form:"business_url"`
+	Country                   *string                       `form:"country"`
 	DebitNegativeBalances     *bool                         `form:"debit_negative_balances"`
-	DefaultCurrency           string                        `form:"default_currency"`
-	Email                     string                        `form:"email"`
+	DefaultCurrency           *string                       `form:"default_currency"`
+	Email                     *string                       `form:"email"`
 	ExternalAccount           *AccountExternalAccountParams `form:"external_account"`
-	FromRecipient             string                        `form:"from_recipient"`
-	LegalEntity               *LegalEntity                  `form:"legal_entity"`
+	FromRecipient             *string                       `form:"from_recipient"`
+	LegalEntity               *LegalEntityParams            `form:"legal_entity"`
 	PayoutSchedule            *PayoutScheduleParams         `form:"payout_schedule"`
-	PayoutStatementDescriptor string                        `form:"payout_statement_descriptor"`
-	ProductDescription        string                        `form:"product_description"`
-	StatementDescriptor       string                        `form:"statement_descriptor"`
-	SupportEmail              string                        `form:"support_email"`
-	SupportPhone              string                        `form:"support_phone"`
-	SupportURL                string                        `form:"support_url"`
+	PayoutStatementDescriptor *string                       `form:"payout_statement_descriptor"`
+	ProductDescription        *string                       `form:"product_description"`
+	StatementDescriptor       *string                       `form:"statement_descriptor"`
+	SupportEmail              *string                       `form:"support_email"`
+	SupportPhone              *string                       `form:"support_phone"`
+	SupportURL                *string                       `form:"support_url"`
 	TOSAcceptance             *TOSAcceptanceParams          `form:"tos_acceptance"`
-	Type                      AccountType                   `form:"type"`
+	Type                      *string                       `form:"type"`
+}
+
+// LegalEntityParams represents a legal_entity during account creation/updates.
+type LegalEntityParams struct {
+	AdditionalOwners []AdditionalOwnerParams `form:"additional_owners,indexed"`
+
+	// AdditionalOwnersEmpty can be set to clear a legal entity's additional
+	// owners.
+	AdditionalOwnersEmpty bool `form:"additional_owners,empty"`
+
+	Address                  *AccountAddressParams       `form:"address"`
+	AddressKana              *AccountAddressParams       `form:"address_kana"`
+	AddressKanji             *AccountAddressParams       `form:"address_kanji"`
+	BusinessName             *string                     `form:"business_name"`
+	BusinessNameKana         *string                     `form:"business_name_kana"`
+	BusinessNameKanji        *string                     `form:"business_name_kanji"`
+	BusinessTaxID            *string                     `form:"business_tax_id"`
+	BusinessTaxIDProvided    *bool                       `form:"-"`
+	BusinessVatID            *string                     `form:"business_vat_id"`
+	BusinessVatIDProvided    *bool                       `form:"-"`
+	DOB                      *DOBParams                  `form:"dob"`
+	FirstName                *string                     `form:"first_name"`
+	FirstNameKana            *string                     `form:"first_name_kana"`
+	FirstNameKanji           *string                     `form:"first_name_kanji"`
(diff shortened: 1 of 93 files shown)


remi-stripe:
@brandur-stripe I'm not sure how to handle this one:
```
// LegalEntity is the structure for properties related to an account's legal state.
type LegalEntity struct {
	AdditionalOwners []Owner `json:"additional_owners" form:"additional_owners,indexed"`

	// AdditionalOwnersEmpty can be set to clear a legal entity's additional
	// owners.
	AdditionalOwnersEmpty bool `form:"additional_owners,empty"`
```

This allows us to explicitly unset `additional_owners` in the library but it should only be done when you really want it. Is this an exception to the Empty logic and we should keep it?

account.go
```diff
@@ -118,13 +117,13 @@ func (p *AccountExternalAccountParams) AppendTo(body *form.Values, keyParts []st
 type PayoutScheduleParams struct {
 	DelayDays     uint64   `form:"delay_days"`
 	Interval      Interval `form:"interval"`
-	MinimumDelay  bool     `form:"-"` // See custom AppendTo
+	MinimumDelay  *bool    `form:"-"` // See custom AppendTo
```
brandur-stripe: Off topic, but should we rename this field? It might be better called `DelayDaysMinimum` or such so that it gets alphabetized with `DelayDays`.

stripe.go
```diff
@@ -476,3 +476,31 @@ func initUserAgent() {
...
 }
+
+// Bool returns a pointer to the bool value passed in.
+func Bool(v bool) *bool {
+	return &v
+}
+
+// BoolValue returns the value of the bool pointer passed in or
+// false if the pointer is nil.
+func BoolValue(v *bool) bool {
```
brandur-stripe: Just OOC, do you think that encapsulating this in a function is a good optimization? I wonder if handling `nil` a little more explicitly might make the code a little easier to understand. (And for clarity, I'm not particularly against it or anything.)

account.go
remi-stripe: ah yes good idea, I can take that one

stripe.go
remi-stripe: Hmmm I just copy/pasted what the AWS SDK does (you referred me to it) and it seemed easier for most cases: https://github.com/aws/aws-sdk-go/blob/71a68668c287f9ad58d0c58c78e27f08eb39fa49/aws/convert_types.go#L64-L76

brandur-stripe:
Cool! LGTM.

> This allows us to explicitly unset additional_owners in the library but it should only be done when you really want it. Is this an exception to the Empty logic and we should keep it?

Hm, good question. Thinking about this broadly, I think that it would be nice if we had a facility that could at least in theory allow you to unset _any_ array (even if the API might not allow it). This would be nicely implemented with something like an `Option<>` in languages that support it, but I'm not sure how off hand how to do this cleanly in Go.

For now, given that `AdditionalOwnersEmpty` already exists and we should probably have a light preference for not changing things for the sake of change, I'd say it's fine to just leave it.

stripe.go
brandur-stripe: Ah, k. I didn't know about the `*Value` functions in AWS. It's fine to leave them.

account.go
remi-stripe: fixed!

remi-stripe:
@brandur-stripe I'm starting to pass all parameters to pointers as agreed. I'm hitting a snag that I'm not sure how to tackle yet though.

We have a lot of types based on String such as `Currency` or `SubscriptionBilling`. Those are passed as parameters. I'm not entirely sure how to handle those as pointers though.

Do I have to add a getter/setter for each one the same way we did for bool/string/int/etc.? It seems weird to use in that case. Or is there a way to "cast" like we would in C automatically?

remi-stripe:
@brandur-stripe friendly bump on this one!

brandur-stripe:
> We have a lot of types based on String such as Currency or SubscriptionBilling. Those are passed as parameters. I'm not entirely sure how to handle those as pointers though.
>
> Do I have to add a getter/setter for each one the same way we did for bool/string/int/etc.? It seems weird to use in that case. Or is there a way to "cast" like we would in C automatically?

Hey @remi-stripe, sorry to keep you waiting!

Hm, so yeah indeed this one is pretty difficult. I'm not really sure if there's a way of fixing this that will fit inside our existing framework here. I believe the answer to your question is "no" in that Go won't be happy to do an implicit type coercion like you might see in C.

I played around with a few possibilities. Here's one that adds a helper for creating a `Currency` pointer:

``` go
type Currency string

const (
	USD Currency = "usd"
)

type MyParams struct {
	MyCurrency *Currency
}

func MakeCurrency() {
	p := MyParams{MyCurrency: CurrencyPointer(USD)}
	fmt.Printf("p = %+v", p)
}

func CurrencyPointer(c Currency) *Currency {
	return &c
}
```

Personally, I think this is a bad idea in that every single new type will need its own function to convert it to a pointer, which isn't great.

I'd suggest that for now we change any instances of special types on individual structs back to their basic types, and then allow an instance of the specialized struct to be type-converted to before being used in one of the standard helpers like `String`:

``` go
type Currency string

const (
	USD Currency = "usd"
)

type MyParams struct {
	MyCurrency *string
}

func MakeCurrency() {
	p := MyParams{MyCurrency: String(string(USD))}
	fmt.Printf("p = %+v", p)
}

func String(v string) *string {
	return &v
}
```

In the long run though, we may just want to drop these special types. The idea of type safety is nice, but TBH I'm not sure that in these types of cases it's buying us all that much. We should probably just go back to this:

``` go
type MyParams struct {
	MyCurrency *string
}

func MakeCurrency() {
	p := MyParams{MyCurrency: String("usd")}
	fmt.Printf("p = %+v", p)
}
```

remi-stripe:
Flagging something so that we don't forget. When passing everything to pointers, we need to triple-check that parameters used in the URL directly are properly passed. I just had a bug where it would hit /v1/0x88888/sources instead of /v1/cus_123/sources and we don't seem to have test that ensure we encode the right URL for various cases.

remi-stripe:
Almost all resources have been moved to pointers. Any type used for both resource and params has been split in 2 separate types instead.

The only remaining one that I'm confused about is in `Account` and has a custom `AppendTo` with a complex comment: https://github.com/stripe/stripe-go/blob/0b14bc3a8a457f89c07d1ccb6148c321b0790a9a/account.go#L440

Now we need to review all the changes (yay) and catch anything I likely missed. We also need to ensure that we add more tests to catch common things.

I will also try to replace all strings in tests by their corresponding const if it exists as we've been quite liberal on that over the years.

remi-stripe:
Also wondering if `Metadata` should be a map of string pointers? A bit unsure how to approach that one specifically.

remi-stripe:
Another question is around the way const are defined. Some list on the resource `xxx.go` while others live on the `xxx/client.go` which means you sometimes do `stripe.CONSTANT_NAME` and other times `CONSTANT_NAME` in the `xxx/client_test.go`. Unsure if it's because we did not do something right but want to flag as something we might want to fix.

account.go
```diff
@@ -252,7 +248,7 @@ type Account struct {
 		UserAgent string `json:"user_agent"`
 	} `json:"tos_acceptance"`
 
-	Type AccountType
+	Type AccountType `json:"type"`
```
remi-stripe: so it worked without the `json` part though felt inconsistent. Is the library doing some magic and de-serializing if property name matches the JSON?

remi-stripe:
@brandur-stripe friendly bump on this one!

remi-stripe:
@brandur-stripe friendly bump on this one!

brandur-stripe:
(Remi and I finally had a chance to go over this in person.)

@remi-stripe I think we made good progress on the decisions above, but let me know if I can help provide clarity anywhere else. And thanks again for biting this off :$

remi-stripe:
rebased on the most recent push to avoid the PR going stale!

remi-stripe:
Closing in favor of https://github.com/stripe/stripe-go/pull/544 which should be the final one and close to review

PR #508, opened 2018-01-08
Correct `Fprintf` format verb

brandur:
Go-tip is failing because it looks like `Fprintf` got a little more
picky about the use of inappropriate verbs.

`%t` is meant to be used for booleans (it prints the word "true" or
"false"). I think what we actually wanted here was `%v`.

cc @ob-stripe


brandur-stripe:
Go-tip is now passing! Pulling this in.

ob-stripe:
👍  Thanks for the quick fix!

PR #509, opened 2018-01-09
Handle all `json.Unmarshal` errors

brandur:
We had a user over in #381 wondering why the errors from a
`json.Unmarshal` in `payout.go` wasn't being handled.

I believe the answer is that we've had a convention in here forever
whereby certain unmarshals from JSON blobs that are "low risk" are
assumed to succeed often enough that the error isn't worth handling. In
some of these `Unmarshal` implementations in particular, the `data` has
already been run through a previous `Unmarshal` so it's known that it
contains valid JSON and is unlikely to error on a second run.

Given that we've never had a complain about this before, it's probably
true that these `Unmarshals` are low risk and probably fine as is, but
nonetheless, it's still good practice to make sure that every error is
handled just in case. In this patch I've gone back through to any
`Unmarshal` invocations that were missing a check and added one.

r? @remi-stripe

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -158,21 +158,25 @@ func (s *TransactionSource) UnmarshalJSON(data []byte) error {
 
 		switch s.Type {
 		case TransactionSourceCharge:
-			json.Unmarshal(data, &s.Charge)
+			err = json.Unmarshal(data, &s.Charge)
 		case TransactionSourceDispute:
-			json.Unmarshal(data, &s.Dispute)
+			err = json.Unmarshal(data, &s.Dispute)
 		case TransactionSourceFee:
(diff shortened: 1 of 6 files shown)


brandur-stripe:
> Wondering if it's possible that this raises new errors which could be a breaking change? Might be worth releasing as a new major version though deferring to you on this, no strong feelings.

Good point, and yes, technically possible. I think if that we're happening though, it would be a sign of an even more significant problem in the library (in that we couldn't even decode our own JSON structures) and I would hope that someone would've noticed by now.

I'll bump the patch release up to a minor release, but I think I'm going to leave it as non-breaking.

brandur-stripe:
Released as 28.10.0.

PR #512, opened 2018-01-18
Encode empty map values

### LINKED ISSUE #511: Cannot remove subscription metadata
I'm trying to remove some metadata on a subscription, and before I simply provided an empty string, but now that does not seem to do anything. 

My Code
```go
params := &stripe.SubParams{}
params.AddMeta("Test", "")
sub, _ := sc.Subs.Update("sub_Bz2wy74jBPoaUD", params)
log.Debugf("METADATA VALUES %v", sub.Meta)
```
I have looked at the request that is created and it does not add the metadata field if the value is empty. I cannot find anything to suggest that removing metadata has changed, so if it has could you kindly show me where that is.

brandur:
Because most parameter fields in stripe-go are non-pointers, we
generally require that a value be non-empty before we encode it, because
otherwise we can't tell whether or not it was explicitly set.

It recently came up in #511 that it's no longer possible to set an empty
value in a map, which is a direct result of this behavior.

This patch tweaks the implementation slightly so that we always encode
all values that we find in maps, even if they're zero values. This makes
sense because unlike a struct which has default fields, we can always
guarantee that anything present in a map *was* explicitly added. This
resolves the problem described above.

Fixes #511.

r? @remi-stripe
cc @ob-stripe

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -298,11 +298,13 @@ func intEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero b
 	values.Add(FormatKey(keyParts), strconv.FormatInt(val, 10))
 }
 
-func interfaceEncoder(values *Values, v reflect.Value, keyParts []string, _ bool, _ *formOptions) {
+func interfaceEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero bool, _ *formOptions) {
+	// interfaceEncoder never encodes a `nil`, but it will pass through an
+	// `encodeZero` value into its chained encoder
 	if v.IsNil() {
 		return
 	}
-	reflectValue(values, v.Elem(), false, keyParts)
+	reflectValue(values, v.Elem(), encodeZero, keyParts)
 }
 
 func isAppender(t reflect.Type) bool {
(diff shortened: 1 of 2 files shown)


form/form.go
```diff
@@ -335,12 +341,17 @@ func uintEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero
 	values.Add(FormatKey(keyParts), strconv.FormatUint(val, 10))
 }
 
-func reflectValue(values *Values, v reflect.Value, _ bool, keyParts []string) {
+// reflectValue is roughly the shared entry point of any AppendTo functions.
+// It's also called recursively in cases where a precise type isn't yet known
+// and its encoding needs to be deferred down the chain; for example, when
+// encoding interface{} or the values in an array or map containing
+// interface{}.
+func reflectValue(values *Values, v reflect.Value, encodeZero bool, keyParts []string) {
```
brandur-stripe: I found myself forgetting the details of this implementation already, so tried to add a few more explanatory comments.

form/form_test.go
```diff
@@ -229,6 +229,15 @@ func TestAppendTo(t *testing.T) {
 			"bar",
 		},
 
+		// Tests map with an empty value
+		{
+			"map[empty]",
+			&testStruct{Map: map[string]interface{}{
+				"empty": int64(0),
```
remi-stripe: Why are you not testing an empty string instead of `int64(0)` since you have a map of strings?

brandur-stripe: Oops, good callout. I meant to write a comment about that and have added one in 886f389.

The explanation is that we are encoding these to a `form.Values` struct, and then in the test using `Get` to check a value against an expected a result. The problem is that `Get` returns an empty string for a value that wasn't set at all, which means that if we were using a string we wouldn't be able to differentiate a zeroed value from an empty value. As a cheap workaround, I've just used an `int64` instead. The behavior around how empties are handled is identical, but we can easily differentiate an empty integer once it's string encoded.

PR #515, opened 2018-01-29
Revert pull request #514 (add `AmountZero`)

brandur:
Reverts pull request #514 which accidentally added `AmountZero` to the
wrong struct.

cc @remi-stripe

PR #514, opened 2018-01-29
Plan can now have zero amount via AmountZero

remi-stripe:
r? @brandur-stripe 

This fixes https://github.com/stripe/stripe-go/issues/513

PR #516, opened 2018-01-29
Add `AmountZero` to `PlanParams`

### LINKED ISSUE #513: Error creating plan with 0 amount
The code below causes the error from the REST endpoint:

`Error encountered from Stripe: {"status":400,"message":"Plans require an `amount` parameter to be set.","param":"amount","request_id":"-some-req-id-","type":"invalid_request_error"}`

looks like it omits the empty value of the `Amount` during form serialization...

```
func (s *StripeClientImpl) CreatePlan(planid string) (*stripe.Plan, error) {
    param := &stripe.PlanParams{
         ID:       planid,
         Amount:   0,
         Interval: "month",
         Name:     planid,
         Currency: "usd",
      }
     s.log.Info(param.Amount)
     p, err := plan.New(param)

     if err != nil {
         return nil, err
     }

     return p, nil
 }
```

brandur:
Follows up #514 and adds `AmountZero` to `PlanParams` so that it's
possible to set amount to a zero value when creating or updating a plan.

Fixes #513.

r? @remi-stripe

---

For now this targets the branch in #515 until that's merged to master.

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -40,6 +40,7 @@ type PlanListParams struct {
 type PlanParams struct {
 	Params        `form:"*"`
 	Amount        uint64       `form:"amount"`
+	AmountZero    bool         `form:"amount,zero"`
 	Currency      Currency     `form:"currency"`
 	ID            string       `form:"id"`
 	Interval      PlanInterval `form:"interval"`


brandur-stripe:
Released as 28.11.0.

korovaisdead:
Sorry for having me here, but just wondering: 
1) What about documentation http://prntscr.com/i7bvet ? 
2) It adds some complexity on the side of the consumer of the package. I mean I can't add different plans by same method call (I need to perform the check on my side wether it has 0 as an amount and then check or uncheck the AmountZero). It makes the public interface of the package more dirty. Would it be better to perfrom this check inside the package instead?

brandur-stripe:
> Sorry for having me here

No worries at all.

> What about documentation http://prntscr.com/i7bvet ?

Thanks for the heads up! I have an internal patch for that going through review now.

> It adds some complexity on the side of the consumer of the package. I mean I can't add different plans by same method call (I need to perform the check on my side wether it has 0 as an amount and then check or uncheck the AmountZero). It makes the public interface of the package more dirty. Would it be better to perfrom this check inside the package instead?

Yeah, we're not crazy about it either.

The trouble is that as written, we _can't_ currently perform the check inside the package because we can't distinguish a zero value in Go that's been explicitly set from one that wasn't (if you initialize a `PlanParams` without including the `Amount` field, it will automatically get a value of `0`).

Pointer-based fields would make things nicer because we wouldn't have to depend on zero values anymore, and we're looking into potentially changing the interface in PRs like #507.

korovaisdead:
Thanks a lot for the answer! 

I do understand your concerns.

PR #519, opened 2018-02-15
update PaymentSource.MarshalJSON to serialize bank accounts too

tylerkovacs:
This pull request fixes serialization for `PaymentSourceBankAccount`.  `PaymentSource` has a  `MarshalJSON` implementation with custom behavior based on type but there was no implementation for type  `PaymentSourceBankAccount` so bank accounts were being serialized as `null`.

### DIFF
--- a/paymentsource.go
+++ b/paymentsource.go
@@ -191,6 +191,21 @@ func (s *PaymentSource) MarshalJSON() ([]byte, error) {
 			ID:   s.ID,
 			Type: s.Type,
 		}
+	case PaymentSourceBankAccount:
+		var customerID *string
+		if s.BankAccount.Customer != nil {
+			customerID = &s.BankAccount.Customer.ID
+		}
+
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Makes sense. Thanks for adding that bonus test too!

Just out of curiosity, what exactly are you using this for?

I was a little surprised to find that it even existed, and had to dig back for context to #88 to see why.

brandur-stripe:
Released as 29.0.1.

tylerkovacs:
Thanks for the merge.

My situation seems similar to #88. I'm building a [Gmail Add-on](https://developers.google.com/gmail/add-ons/) for Stripe. In my case, the client is Gmail and requests are proxied through my servers rather than hitting Stripe directly. I noticed that the response from the Stripe API was correct, but became corrupt as soon as the objects were re-serialized on the way back to the client.

PR #522, opened 2018-02-17
Fix latest version to properly support Product in plans

remi-stripe:
This ensures that you can create plans with the latest API version. We had 2 separate bugs:
* re-using `ProductParams` meant that you could pass invalid parameters when creating a product inside a plan (such as Type which is unsupported).
* not properly enforcing the latest API version which means all requests are currently failing.

This is a breaking change **but** it fixes what we broke in 29.0 so it's likely a better fit for 29.0.2 but unsure.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -39,14 +39,24 @@ type PlanListParams struct {
 // For more details see https://stripe.com/docs/api#create_plan and https://stripe.com/docs/api#update_plan.
 type PlanParams struct {
 	Params        `form:"*"`
-	Amount        uint64         `form:"amount"`
-	AmountZero    bool           `form:"amount,zero"`
-	Currency      Currency       `form:"currency"`
-	ID            string         `form:"id"`
-	Interval      PlanInterval   `form:"interval"`
-	IntervalCount uint64         `form:"interval_count"`
-	Nickname      string         `form:"nickname"`
-	Product       *ProductParams `form:"product"`
-	ProductID     *string        `form:"product"`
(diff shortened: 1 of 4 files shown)


remi-stripe:
@brandur-stripe While on this PR, I think there's a bug in stripe-mock since it did not catch that `product[type]` on plan creation is forbidden.

brandur-stripe:
Oops, I kind of skipped over this one. Thanks for taking Remi!

> This is a breaking change but it fixes what we broke in 29.0 so it's likely a better fit for 29.0.2 but unsure.

Yeah .. I'm going to send it out as a patch release and make a note about 29.0. I doubt too many people were using the plan features of 29.0 yet, so hopefully it's okay.

> @brandur-stripe While on this PR, I think there's a bug in stripe-mock since it did not catch that product[type] on plan creation is forbidden.

I think the problem currently is that extra/unknown parameters are allowed. In this particular product subschema, `type` isn't even registered as a possibility:

``` yaml
product:
  anyOf:
  - description: The product whose pricing the created plan will represent.
      This can either be the ID of an existing product, or a dictionary
      containing fields used to create a [service product](/docs/api#product_object-type).
    properties:
      id:
        type: string
      metadata:
        type: object
      name:
        type: string
      statement_descriptor:
        type: string
    required:
    - name
    title: inline_product_params
    type: object
  - description: The ID of the product whose pricing the created plan
      will represent.
    type: string
```

brandur-stripe:
Released as 29.1.1.

PR #525, opened 2018-02-26
Support shipping carrier and tracking number in orders

tylerkovacs:
This PR addresses missing functionality related to shipping attributes of an order.

* The `Shipping` struct was missing attributes for `carrier` and `tracking_number`.

* The `OrderUpdateParams` struct did not allow the shipping carrier or tracking number to be updated. You can workaround it using `AddExtra` but we might as well have typed support for these values.

### DIFF
--- a/order.go
+++ b/order.go
@@ -46,9 +46,11 @@ type OrderReturnParams struct {
 }
 
 type Shipping struct {
-	Address Address `json:"address"`
-	Name    string  `json:"name"`
-	Phone   string  `json:"phone"`
+	Address        Address `json:"address"`
+	Name           string  `json:"name"`
+	Phone          string  `json:"phone"`
+	Carrier        string  `json:"carrier"`
+	TrackingNumber string  `json:"tracking_number"`
 }
 
 type ShippingMethod struct {
--- a/order.go
+++ b/order.go
@@ -35,8 +35,8 @@ type OrderUpdateParams struct {
 	Params                 `form:"*"`
 	Coupon                 string                     `form:"coupon"`
 	SelectedShippingMethod string                     `form:"selected_shipping_method"`
-	Status                 OrderStatus                `form:"status"`
 	Shipping               *OrderUpdateShippingParams `form:"shipping"`
+	Status                 OrderStatus                `form:"status"`
 }
 
 type OrderUpdateShippingParams struct {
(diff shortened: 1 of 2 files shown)


order.go
```diff
@@ -33,9 +33,15 @@ type ShippingParams struct {
 
 type OrderUpdateParams struct {
 	Params                 `form:"*"`
-	Coupon                 string      `form:"coupon"`
-	SelectedShippingMethod string      `form:"selected_shipping_method"`
-	Status                 OrderStatus `form:"status"`
+	Coupon                 string                     `form:"coupon"`
+	SelectedShippingMethod string                     `form:"selected_shipping_method"`
+	Status                 OrderStatus                `form:"status"`
+	Shipping               *OrderUpdateShippingParams `form:"shipping"`
```
brandur-stripe: This is a bit of a nitpick, but would you mind alphabetizing the fields here? (i.e., put `Shipping` above `Status`.)

order.go
```diff
@@ -46,9 +52,11 @@ type OrderReturnParams struct {
...
 
 type Shipping struct {
-	Address Address `json:"address"`
-	Name    string  `json:"name"`
-	Phone   string  `json:"phone"`
+	Address        Address `json:"address"`
+	Name           string  `json:"name"`
+	Phone          string  `json:"phone"`
+	Carrier        string  `json:"carrier"`
+	TrackingNumber string  `json:"tracking_number"`
```
brandur-stripe: And same here — just move `Carrier` into the right place.

brandur-stripe:
Thanks for fixing this, and with tests too! Excellent PR. Thanks for the contribution!

order.go
tylerkovacs: Done! Can't believe I missed that since I've left that same comment on countless PRs... :)

tylerkovacs:
BTW, I noticed a few documentation bugs, where should I report those?

e.g., The Go example for update order passes in a `&stripe.OrderParams` but the code expects a `&stripe.OrderUpdateParams`
https://stripe.com/docs/api/go#update_order

order.go
brandur-stripe: Haha! Thanks again.

brandur-stripe:
Released as 29.2.0.

brandur-stripe:
> BTW, I noticed a few documentation bugs, where should I report those?
>
> e.g., The Go example for update order passes in a &stripe.OrderParams but the code expects a &stripe.OrderUpdateParams
> https://stripe.com/docs/api/go#update_order

Oof, right again. If you report them here or through support, someone can open an internal ticket to get them fixed. Let me just fix this one now.

PR #526, opened 2018-02-26
returning an Order is not using common params

tylerkovacs:
The `Client` code is not currently using commonParams when returning an order. As a result, information (such as setting the StripeAccount) is lost.


tylerkovacs:
Tested by spinning up an ephemeral server (using `httptest.NewServer`) and verifying that a request header, which is derived from common params, is sent.

brandur-stripe:
Thanks a lot Tyler! (It's nice to have a test template now for this kind of thing now too, thanks for adding that.)

brandur-stripe:
Released as 29.2.0.

PR #530, opened 2018-03-07
Fix mixed-up types in CountrySpec.SupportedBankAccountCurrencies

sfriedel:
This PR fixes a wrong type declaration of the  `CountrySpec.SupportedBankAccountCurrencies` field.
The API exposes a `map[Currency][]Country` while the library defined it as a `map[Country][]Currency`.


ob-stripe:
Released as 29.3.1. Thanks for the fix @sfriedel!

PR #533, opened 2018-03-14
Change Transfer.DestPayment from string to *Charge

drtz:
This change allows an expanded `transfer.destination_payment` to be accessed via the `DestPayment` field.

It is a small breaking change to the API.


brandur-stripe:
Makes sense to me. Thanks!

brandur-stripe:
Released as 30.0.0.

PR #535, opened 2018-03-23
Add const for redirect flow status "not_required"

sfriedel:
This PR adds the missing const definition for the "not_required" redirect flow status that was added in API version 2017-08-15.


brandur-stripe:
Thanks!

brandur-stripe:
Released as 30.2.0.

PR #544, opened 2018-04-14
Rename all properties in the library to match the API reference

remi-stripe:
This should be the final PR to rework most of the library to:
- match the naming in the API
- split out resource and params structures
- remove magic such as `AmountZero` or `CouponEmpty` and move all parameters to pointers

We can't list all the changes one by one as those are extensive but here are the ones likely to impact most integrations:

* All properties with the word url in it now have the same case such as `BusinessURL`.
* All properties with the word four in it now use a digit such as `DynamicLast4` or `SSNLast4`.
* Renamed confusing properties such as `Account` on `BankAccount` becoming `AccountNumber` to avoid confusing with its Connect counterpart which is now named `Account` instead of `AccountID`.
* Properties such as `MonthAnchor` or constants such as `Month` are now renamed to match the corresponding parameter/value in the API `MonthlyAnchor` and `Monthly`.
* `Meta` and `Live` have been renamed to `Metadata` and `Livemode` on all classes
* `Values` on List objects has been renamed to `Data`
* Some classes have been renamed. 
  * `Owner` is now `AdditionalOwner`,
  * `Transaction` is now `BalanceTransaction` with all constants and corresponding classes renamed to match this
  * references to `Tx` were also replaced by `BalanceTransaction` in classes, properties and constants
  * `Fee` and `FeeRefund` have been renamed to `ApplicationFee` and `ApplicationFeeRefund` along with all related classes, properties and constants.
  * `Sub` has been renamed to `Subscription` along with all related classes, properties and constants.
* Card parameters have been renamed such as `City` becoming `AddressCity` and `Month` becoming `ExpMonth`.
* Errors codes have been prefixed with `ErrorCode` such as `ErrorCodeCardDeclined` or `ErrorCodeIncorrectZip`.
* Event properties are now named `Object` and `PreviousAttributes` and methods `GetObjectValue()` and `GetPreviousValue()`.
* Pagination properties have been renamed to `StartingAfter`, `EndingBefore` and `HasMore`.
* `AddMeta()` has been renamed to `AddMetadata()`, `Expand()` to `AddExpand()`

This replaces https://github.com/stripe/stripe-go/pull/459 and https://github.com/stripe/stripe-go/pull/507


fileupload.go
```diff
@@ -14,10 +13,6 @@ import (
 type FileUploadParams struct {
 	Params `form:"*"`
 
-	// File is a deprecated form of FileReader and Filename that will do the same thing, but
-	// allows referencing a file directly. Please prefer the use of FileReader and Filename instead.
-	File *os.File
-
 	// FileReader is a reader with the contents of the file that should be uploaded.
```
remi-stripe: Flagging that I seem to not be able to upload with `FileReader` if I don't explicitly set `Filename` too. Not sure if expected but I don't want to forget

event.go
```diff
@@ -54,14 +54,14 @@ type EventListParams struct {
 	Types        []string          `form:"types"`
```
remi-stripe: Should this be a pointer to the array?

card.go
```diff
@@ -35,24 +35,24 @@ type Verification string
...
+	DefaultForCurrency bool   `form:"default_for_currency"`
+	ExpMonth           string `form:"exp_month"`
+	ExpYear            string `form:"exp_year"`
+	Name               string `form:"name"`
+	Number             string `form:"number"`
+	Recipient          string `form:"-"`
+	Token              string `form:"-"`
 
 	// ID is used when tokenizing a card for shared customers
 	ID string `form:"*"`
```
remi-stripe: Should this be a pointer?

brandur-stripe:
Haha, holy cow this is a monster! Amazing work Remi!

Just to be explicit about it: is this ready for review?

remi-stripe:
@brandur-stripe I think it is. The only todo left is to move/rename all constants to the upper package. This is less likely to break something and I will do it as one unique commit separately hopefully this week but we should not block on it.

Let me know if I can help with the review in any way. The main gotcha I know of would be for parameters passed in the URL to be incorrectly sent as pointers address instead of the value. I haven't figured out a way to ensure this does not happen without just carefully reading the code though.

brandur-stripe:
> Let me know if I can help with the review in any way. The main gotcha I know of would be for parameters passed in the URL to be incorrectly sent as pointers address instead of the value. I haven't figured out a way to ensure this does not happen without just carefully reading the code though.

We should write an "encode parameters" function that takes a splat and checks each value probably. I think we can push that to a different PR though. Maybe one that I could do for a change :)

remi-stripe:
Ah yes we talked about this last week but that seemed like a lot of work and I was worried I might make things worse mixing this here. Do you think we should do this before merging? It does seem safer for sure.

If you were to start the PR/build the method/examples, I can happily port all the clients to it afterwards

balance.go
```diff
@@ -62,56 +62,56 @@ type BalanceParams struct {
...
-	Type       TransactionType   `json:"type"`
+type BalanceTransaction struct {
+	Amount      int64                    `json:"amount"`
+	AvailableOn int64                    `json:"available_on"`
+	Created     int64                    `json:"created"`
+	Currency    Currency                 `json:"currency"`
+	Description string                   `json:"description"`
+	ID          string                   `json:"id"`
+	Fee         int64                    `json:"fee"`
+	FeeDetails  []BalanceTransactionFee  `json:"fee_details"`
```
brandur-stripe: Maybe pointer-ify this.

balance/client.go
```diff
@@ -7,22 +7,22 @@ import (
...
+	BalanceTransactionAppFee                  stripe.BalanceTransactionType = "application_fee"
+	BalanceTransactionFeeRefund               stripe.BalanceTransactionType = "application_fee_refund"
+	BalanceTransactionRecipientTransfer       stripe.BalanceTransactionType = "recipient_transfer"
+	BalanceTransactionRecipientTransferCancel stripe.BalanceTransactionType = "recipient_transfer_cancel"
+	BalanceTransactionRecipientTransferFail   stripe.BalanceTransactionType = "recipient_transfer_failure"
+	BalanceTransactionPayout                  stripe.BalanceTransactionType = "payout"
+	BalanceTransactionPayoutCancel            stripe.BalanceTransactionType = "payout_cancel"
+	BalanceTransactionPayoutFail              stripe.BalanceTransactionType = "payout_failure"
+	BalanceTransactionTransfer                stripe.BalanceTransactionType = "transfer"
+	BalanceTransactionTransferCancel          stripe.BalanceTransactionType = "transfer_refund"
```
brandur-stripe: Did we decide to move these to the top-level `stripe` package?

bankaccount.go
```diff
@@ -103,36 +102,36 @@ type BankAccountListParams struct {
...
-	LastFour          string            `json:"last4"`
-	Meta              map[string]string `json:"metadata"`
-	Name              string            `json:"bank_name"`
-	Routing           string            `json:"routing_number"`
-	Status            BankAccountStatus `json:"status"`
+	AccountHolderName  string            `json:"account_holder_name"`
+	AccountHolderType  string            `json:"account_holder_type"`
+	BankName           string            `json:"bank_name"`
+	Country            string            `json:"country"`
+	Currency           Currency          `json:"currency"`
```
brandur-stripe: We might want to pull the `Currency` type as well.

bankaccount/client.go
```diff
@@ -16,10 +16,11 @@ type Client struct {
...
 const (
-	NewAccount       stripe.BankAccountStatus = "new"
-	VerifiedAccount  stripe.BankAccountStatus = "verified"
-	ValidatedAccount stripe.BankAccountStatus = "validated"
-	ErroredAccount   stripe.BankAccountStatus = "errored"
+	NewAccount                stripe.BankAccountStatus = "new"
+	VerificationFAiledAccount stripe.BankAccountStatus = "verification_failed"
+	VerifiedAccount           stripe.BankAccountStatus = "verified"
+	ValidatedAccount          stripe.BankAccountStatus = "validated"
+	ErroredAccount            stripe.BankAccountStatus = "errored"
```
brandur-stripe: We should probably prefix these with `BankAccountStatus` and move them to the `stripe` package.

bitcoinreceiver.go
```diff
@@ -8,49 +8,49 @@ import (
...
 // For more details see https://stripe.com/docs/api/#create_bitcoin_receiver.
 type BitcoinReceiverParams struct {
 	Params   `form:"*"`
-	Amount   uint64   `form:"amount"`
-	Currency Currency `form:"currency"`
-	Desc     string   `form:"description"`
-	Email    string   `form:"email"`
+	Amount   *int64  `form:"amount"`
+	Currency *string `form:"currency"`
+	Desc     *string `form:"description"`
```
brandur-stripe: This should be `Description`.

bitcoinreceiver.go
```diff
@@ -8,49 +8,49 @@ import (
...
 
 // BitcoinReceiverUpdateParams is the set of parameters that can be used when
 // updating a BitcoinReceiver. For more details see
 // https://stripe.com/docs/api/#update_bitcoin_receiver.
 type BitcoinReceiverUpdateParams struct {
 	Params     `form:"*"`
-	Desc       string `form:"description"`
-	Email      string `form:"email"`
-	RefundAddr string `form:"refund_address"`
+	Desc       *string `form:"description"`
```
brandur-stripe: `Description` here as well.

bitcoinreceiver.go
```diff
@@ -8,49 +8,49 @@ import (
...
+	Email      *string `form:"email"`
+	RefundAddr *string `form:"refund_address"`
```
brandur-stripe: `RefundAddress` here.

charge.go
```diff
@@ -158,11 +167,11 @@ type ChargeOutcome struct {
 
 // ShippingDetails is the structure containing shipping information.
 type ShippingDetails struct {
-	Address  Address `json:"address" form:"address"`
-	Carrier  string  `json:"carrier" form:"carrier"`
-	Name     string  `json:"name" form:"name"`
-	Phone    string  `json:"phone" form:"phone"`
-	Tracking string  `json:"tracking_number" form:"tracking_number"`
```
brandur-stripe: So nice to see the dual-purpose structs gone!

customer.go
```diff
@@ -8,22 +8,26 @@ import (
...
-	Plan           string                   `form:"plan"`
-	Quantity       uint64                   `form:"quantity"`
-	Shipping       *CustomerShippingDetails `form:"shipping"`
-	Source         *SourceParams            `form:"*"` // SourceParams has custom encoding so brought to top level with "*"
-	TaxPercent     float64                  `form:"tax_percent"`
-	TaxPercentZero bool                     `form:"tax_percent,zero"`
-	Token          string                   `form:"-"` // This doesn't seem to be used?
-	TrialEnd       int64                    `form:"trial_end"`
+	AccountBalance *int64                         `form:"account_balance"`
+	BusinessVatID  *string                        `form:"business_vat_id"`
```
brandur-stripe: Another instance of "VAT".

customer.go
```diff
@@ -38,42 +42,42 @@ func (cp *CustomerParams) SetSource(sp interface{}) error {
...
-	Values []*Customer `json:"data"`
+	Data []*Customer `json:"data"`
 }
 
 // CustomerShippingDetails is the structure containing shipping information.
 type CustomerShippingDetails struct {
-	Address Address `json:"address" form:"address"`
-	Name    string  `json:"name" form:"name"`
-	Phone   string  `json:"phone" form:"phone"`
+	Address Address `json:"address"`
```
brandur-stripe: Maybe make `Address` here a pointer for consistency.

dispute/client.go
```diff
@@ -8,23 +8,23 @@ import (
...
+	Unrecognized         stripe.DisputeReason = "unrecognized"
+
+	Lost                 stripe.DisputeStatus = "lost"
+	NeedsResponse        stripe.DisputeStatus = "needs_response"
+	ChargeRefunded       stripe.DisputeStatus = "charge_refunded"
+	UnderReview          stripe.DisputeStatus = "under_review"
+	WarningClosed        stripe.DisputeStatus = "warning_closed"
+	WarningNeedsResponse stripe.DisputeStatus = "warning_needs_response"
+	WarningUnderReview   stripe.DisputeStatus = "warning_under_review"
+	Won                  stripe.DisputeStatus = "won"
```
brandur-stripe: Maybe prefix these and bump them up to `stripe` as well.

error.go
```diff
@@ -17,26 +17,19 @@ const (
...
-	// These additional types are written purely for backward compatibility
-	// (the originals were given quite unsuitable names) and should be
-	// considered deprecated. Remove them on the next major version revision.
-
-	APIErr         ErrorType = ErrorTypeAPI
-	CardErr        ErrorType = ErrorTypeCard
-	InvalidRequest ErrorType = ErrorTypeInvalidRequest
+	ErrorCodeCardDeclined       ErrorCode = "card_declined"
+	ErrorCodeExpiredCard        ErrorCode = "expired_card"
+	ErrorCodeIncorrectCvc       ErrorCode = "incorrect_cvc"
```
brandur-stripe: Maybe change this to "CVC".

error.go
```diff
@@ -17,26 +17,19 @@ const (
...
+	ErrorCodeIncorrectZip       ErrorCode = "incorrect_zip"
+	ErrorCodeIncorrectNumber    ErrorCode = "incorrect_number"
+	ErrorCodeInvalidCvc         ErrorCode = "invalid_cvc"
```
brandur-stripe: And same here: "CVC".

fileupload.go
brandur-stripe: Hm, that shouldn't be the case, but we should look at that more closely.

payout.go
```diff
@@ -58,55 +58,55 @@ type PayoutDestination struct {
...
-	Type                      PayoutType        `json:"type"`
+	Amount                    int64               `json:"amount"`
+	ArrivalDate               int64               `json:"arrival_date"`
+	Automatic                 bool                `json:"automatic"`
+	BalanceTransaction        *BalanceTransaction `json:"balance_transaction"`
+	BankAccount               *BankAccount        `json:"bank_account"`
+	Card                      *Card               `json:"card"`
+	Created                   int64               `json:"created"`
+	Currency                  Currency            `json:"currency"`
+	Destination               PayoutDestination   `json:"destination"`
```
brandur-stripe: Make this a point for consistency.

sub/client.go
```diff
@@ -9,12 +9,12 @@ import (
...
-	PastDue  stripe.SubStatus = "past_due"
-	Canceled stripe.SubStatus = "canceled"
-	Unpaid   stripe.SubStatus = "unpaid"
-	All      stripe.SubStatus = "all"
+	Trialing stripe.SubscriptionStatus = "trialing"
+	Active   stripe.SubscriptionStatus = "active"
+	PastDue  stripe.SubscriptionStatus = "past_due"
+	Canceled stripe.SubscriptionStatus = "canceled"
+	Unpaid   stripe.SubscriptionStatus = "unpaid"
+	All      stripe.SubscriptionStatus = "all"
```
brandur-stripe: Maybe hoist these up to `stripe` as wel.

remi-stripe:
@brandur-stripe Thank you so much for the thorough review. I just fixed everything you raised that was an omission!

remi-stripe:
@brandur-stripe Okay I just started removing the "fake" types that we alias to string in the `Account` resource. I had to change related resource to use the new way or have their own constants. For example, even though the values are the same for bank account holder types and legal entity, they definitely should be different constants.

Before I go ahead and change the entire library again, can you confirm that this is the right approach and looks cleaner? Also, what do you think about not adding comments for those? That seems a bit much and most names are self explanatory.

remi-stripe:
Once we've done all constants, I'll then move to passing each embedded struct as a pointer too.

brandur-stripe:
> Edit: Okay I had broken the build so I fixed it quickly. There's a small one that annoys me a bit which is HTTPStatusCode here. If we turn it in int64 you apparently can't do this:

Hah, yeah I thought about calling out status code too, but decided not to. If you look throughout the `[net/http](https://godoc.org/net/http#StatusText)` package in the standard library, it's an `int` everywhere, and I think that we should keep it as is just to be consistent with that. It would be a bad interface to have to convert to `int` when comparing to any codes from there.

> Before I go ahead and change the entire library again, can you confirm that this is the right approach and looks cleaner? Also, what do you think about not adding comments for those? That seems a bit much and most names are self explanatory.

Yep, that looks right. And +1 removing the comments.

> Once we've done all constants, I'll then move to passing each embedded struct as a pointer too.

Sounds good! Thanks Remi.

remi-stripe:
@brandur-stripe My understanding is that we want to remove `Currency` entirely. Do we also want to start passing `"usd"` which is likely what most people do or do we want to add `CurrencyUSD` and `CurrencyEUR`, etc.?

brandur-stripe:
> @brandur-stripe My understanding is that we want to remove Currency entirely. Do we also want to start passing "usd" which is likely what most people do or do we want to add CurrencyUSD and CurrencyEUR, etc.?

Yeah, good question.

I think that since we sort of already have the list of constants, we should probably keep them around (but make them strings, so that the interface is compatible with just passing `"usd"` if you want) — there's a large-ish class of developer (myself include I think) who like to use constants when they're available just for good hygiene.

The counterargument is that some of these lists can be prone to becoming outdated (as new currencies are added and such), but my intuition would be that currencies (or more specifically, supported currencies) should be stable _enough_ that isn't isn't likely to be a huge problem.

brandur-stripe:
@remi-stripe Amazing! Sorry again for the snail's pace here :$ Let me know when all your changes are ready.

remi-stripe:
@brandur-stripe Okay I think we're ~almost there. I'm not sure it's the final version yet but I would like some quick 👍  or 👎  on all the last few changes:
* Stripe types for most enum-like string properties and string for all parameters.
* Removal of the `currency` package and whether we want `currency.go` as a top-level file or if it should go in `stripe.go`

Also whether we want to also address one of your earlier comments in this PR too: 
> In the response objects, some embedded structs are pointers (*AdditionalOwner) and some are just structs (AdditionalOwner). I think we should skew all the way in one direction by just making them all pointers.

brandur-stripe:
Nice @remi-stripe! See below:

> Stripe types for most enum-like string properties and string for all parameters.

Did you change anything already? I think it'd be nice to have types where possible, but I don't think it matters _that_ much. I'd suggest leaving the ones that exist in place, and probably just not adding anything new for now.

> Removal of the currency package and whether we want currency.go as a top-level file or if it should go in stripe.go

I'd just put it in top-level `currency.go`. We want to kill the package, but it's fine to keep it organized in its own file.

> Also whether we want to also address one of your earlier comments in this PR too:

Hm, did you get a feeling as to how much of a pain this is to fix? I still think this would be somewhat nice to have just so it's predictable (i.e., things area always pointers) when you're writing your integration code.

brandur-stripe:
> Yes I did all the changes we agreed upon yesterday for all the packages. This means all params structs have *string but all resources have custom types for each property that is really an enum. This is done in this specific commit: 35258f8
>
> This means that I undid all the changes I did last month and re-added each type (and manually added a few more that were missing) and made sure that all constants are now top-level and properly prefixed.
>
> Can you confirm this is what we expected?

Oh man, you are a machine! Yes — this is what I was thinking.

> That's what I did in ffbc8bd can you confirm that was correct?

Yep, this also looks good to me.

> I'm hopeful this is mostly a matter of reading each struct and replacing with a *Type instead of Type so it should be doable.

Okay then, yeah I think we may as well go for it given the number of other changes that we already have in here. Feel free to push back though if this will push you over the edge, haha.

remi-stripe:
r? @brandur-stripe 

Okay I think we're there now. This is definitely a huge PR though so it's hard to review, really sorry about that. I tried to make it in separate PRs originally but the rebase(s) were just impossible in the end after months of changes.

My main worries are:
* Renamed the wrong parameter which means something will not deserialize properly
* Encoding in the client not properly deserializing a pointer, or me not passing the right thing causing things like an ID to end up as `0x9876123` instead of the ID. We had discussed this and potentially wanted to rewrite the client to do the work.
* Deserialization on embedded structs not working properly and resources from stripe-mock not detailed enough to trigger the path

Let me know what you think and if I can help make the review easier.

remi-stripe:
@brandur-stripe Flagging that while reviewing the raw logs of stripe-mock during a test run I found a few example parameters that were not properly dereferenced. Hopefully those are the only ones.

brandur-stripe:
> @brandur-stripe Flagging that while reviewing the raw logs of stripe-mock during a test run I found a few example parameters that were not properly dereferenced. Hopefully those are the only ones.

Doh, okay thanks. Was there any pattern to the bug that caused that that we could look for more of?

(By the way, I'm reviewing now, but it will still take some time.)

remi-stripe:
> Doh, okay thanks. Was there any pattern to the bug that caused that that we could look for more of?

Yes they were all calls like this:

```
err = c.B.Call("GET", fmt.Sprintf("/sources/%v/source_transactions", params.Source), c.Key, b, p, list)
```

Basically that params is a pointer now and needs to be converted with `stripe.StringValue()`. So I grep-ed for all the similar calls with `\.Call\(.* params\.` and was able to find only one other occurrence (the one above, as it did not use stripe-mock and just skipped the test)

brandur-stripe:
> Basically that params is a pointer now and needs to be converted with stripe.StringValue(). So I grep-ed for all the similar calls with \.Call\(.* params\. and was able to find only one other occurrence (the one above, as it did not use stripe-mock and just skipped the test)

Okay, great.

And yeah, I know I said it before, but as a follow up (not here) we should change those to use a helper function of our design that rejects pointers and empty values. Good for finding bugs, but it'd also probably help users who end up passing bad values by accident.

order.go
```diff
@@ -8,40 +8,58 @@ import (
...
+)
+
+// OrderItemType represents the type of order item
+type OrderItemType string
+
+const (
+	OrderItemTeypeDiscount OrderItemType = "discount"
+	OrderItemTeypeShipping OrderItemType = "shipping"
+	OrderItemTeypeSKU      OrderItemType = "sku"
+	OrderItemTeypeTax      OrderItemType = "tax"
```
brandur-stripe: Looks like some typos for all these constants: "OrderItemTeype".

plan.go
```diff
@@ -7,134 +7,136 @@ import (
...
-	UpTo   uint64 `json:"up_to"`
+	Amount int64 `json:"amount"`
+	UpTo   int64 `json:"up_to"`
 }
 
 // PlanTransformUsage represents the bucket billing configuration.
 type PlanTransformUsage struct {
-	DivideBy int64  `json:"bucket_size"`
-	Round    string `json:"round"`
+	DivideBy int64                   `json:"bucket_size"`
```
brandur-stripe: What's the detail with this one again? Should we rename to `BucketSize`?

plan.go
```diff
@@ -7,134 +7,136 @@ import (
...
+	Round    PlanTransformUsageRound `json:"round"`
 }
 
 // PlanTransformUsageParams represents the bucket billing configuration.
 type PlanTransformUsageParams struct {
-	DivideBy int64  `form:"bucket_size"`
-	Round    string `form:"round"`
+	DivideBy *int64  `form:"bucket_size"`
```
brandur-stripe: (And same question here.)

recipient.go
```diff
@@ -7,34 +7,38 @@ import (
 )
 
 // RecipientType is the list of allowed values for the recipient's type.
-// Allowed values are "individual", "corporation".
 type RecipientType string
 
+const (
+	RecipientTypeIndividual RecipientType = "individual"
+	RecipientTypeCorp       RecipientType = "corporation"
```
brandur-stripe: Not a big deal because of recipient, but should we call this `RecipientTypeCorporation` just to kill all abbreviations?

recipienttransfer.go
```diff
@@ -5,42 +5,60 @@ import "encoding/json"
...
 
 const (
-	// RecipientTransferDestinationBankAccount is a constant representing a recipient_transfer destination
-	// which is a bank account.
-	RecipientTransferDestinationBankAccount RecipientTransferDestinationType = "bank_account"
-
-	// RecipientTransferDestinationCard is a constant representing a recipient_transfer destination
-	// which is a card.
-	RecipientTransferDestinationCard RecipientTransferDestinationType = "card"
+	RecipientTransferTypeBank RecipientTransferType = "bank_account"
```
brandur-stripe: Similarly, maybe expand this out to `RecipientTransferTypeBankAccount`.

stripe.go
```diff
@@ -485,3 +480,59 @@ func initUserAgent() {
...
+// false if the pointer is nil.
+func BoolValue(v *bool) bool {
+	if v != nil {
+		return *v
+	}
+	return false
+}
+
+// String returns a pointer to the string value passed in.
+func String(v string) *string {
```
brandur-stripe: Would you mind if we just alphabetized these new functions relative to each other for nicety?

usage_record.go
```diff
@@ -1,33 +1,26 @@
...
-	// UsageRecordParamsActionIncrement indicates that if two usage records conflict
-	// (i.E. are reported a the same timestamp), their Quantity will be summed
-	UsageRecordParamsActionIncrement string = "increment"
-
-	// UsageRecordParamsActionSet indicates that if two usage records conflict
-	// (i.E. are reported a the same timestamp), the Quantity of the most recent
-	// usage record will overwrite any other quantity.
-	UsageRecordParamsActionSet string = "set"
+	UsageRecordActionIncrement string = "increment"
+	UsageRecordActionSet       string = "set"
```
brandur-stripe: Should these be a new type like `type UsageRecordAction string`?

account.go
```diff
@@ -6,81 +6,174 @@ import (
...
+	LegalEntityTypeIndividual LegalEntityType = "individual"
+)
+
 // IdentityVerificationDetailsCode is a machine-readable code specifying the
-// verification state of a legal entity. Allowed values are
-// "failed_keyed_identity", "failed_other", "scan_corrupt",
-// "scan_failed_greyscale", "scan_failed_other",
-// "scan_id_country_not_supported", "scan_id_type_not_supported",
-// "scan_name_mismatch", "scan_not_readable", "scan_not_uploaded".
+// verification state of a legal entity
```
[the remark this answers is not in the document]
remi-stripe: I'm happy you say that. I was afraid this was a bad idea but I really did not want to have to update all of them manually

bankaccount.go
```diff
@@ -65,35 +79,35 @@ func (a *BankAccountParams) AppendToAsSourceOrExternalAccount(body *form.Values)
...
 	// Use token (if exists) or a dictionary containing a user’s bank account details.
-	if len(a.Token) > 0 {
-		body.Add(sourceType, a.Token)
+	if a.Token != nil {
+		body.Add(sourceType, StringValue(a.Token))
 
-		if a.Default {
-			body.Add("default_for_currency", strconv.FormatBool(a.Default))
+		if a.DefaultForCurrency != nil {
+			body.Add("default_for_currency", strconv.FormatBool(BoolValue(a.DefaultForCurrency)))
```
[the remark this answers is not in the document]
remi-stripe: Yeah I think it's better just for consistency. Because you end up grep-ing all of them the same way in that case.

plan.go
remi-stripe: damn yeah those were added after I renamed everything so I did not triple check. Also they were just completely broken because the constants had the wrong values lol

usage_record.go
remi-stripe: so the reason I did not put a type it's because it's only used as a parameter. Happy to add one though

remi-stripe:
@brandur-stripe Thank you for the thorough review! I fixed all the comments but the last one. Let me know if I should fix that one too.

Agreed on the Docs though if you look closely a lot of the examples are already wrong anyway for some of the big resources we renamed. But yeah that one will be a lot of work to land

account.go
```diff
@@ -6,81 +6,174 @@ import (
...
+// LegalEntityParams represents a legal_entity during account creation/updates.
+type LegalEntityParams struct {
+	AdditionalOwners []*AdditionalOwnerParams `form:"additional_owners,indexed"`
 
-	// IdentityVerificationUnverified is a constant value indicating that
-	// identity verification status is unverified.
-	IdentityVerificationUnverified IdentityVerificationStatus = "unverified"
+	// AdditionalOwnersEmpty can be set to clear a legal entity's additional
+	// owners.
+	AdditionalOwnersEmpty bool `form:"additional_owners,empty"`
```
remi-stripe: @brandur-stripe While updating the examples in the docs I remembered I forgot to ask you about this one! That is (I think) the only place left in the entire library where we have an `empty` logic because `AdditionalOwners` is a slice of pointers. Do you know of a good fix or do we leave this as an exception for now?

discount/client.go
```diff
@@ -38,7 +38,7 @@ func (c Client) Del(customerID string, params *stripe.DiscountParams) (*stripe.D
 
 // DelSub removes a discount from a customer's subscription.
 // For more details see https://stripe.com/docs/api#delete_subscription_discount.
-func DelSub(subscriptionID string, params *stripe.DiscountParams) (*stripe.Discount, error) {
+func DelSubscription(subscriptionID string, params *stripe.DiscountParams) (*stripe.Discount, error) {
```
remi-stripe: @brandur-stripe  while renaming this one, I realized we might want to rename all `Del` methods to `Delete`? Let me know what you think

sub.go
```diff
@@ -65,15 +65,15 @@ type SubscriptionCancelParams struct {
...
 		body.Add(form.FormatKey(append(keyParts, "billing_cycle_anchor")), "now")
 	}
 
-	if p.BillingCycleAnchorUnchanged != nil {
+	if BoolValue(p.BillingCycleAnchorUnchanged) {
 		body.Add(form.FormatKey(append(keyParts, "billing_cycle_anchor")), "unchanged")
 	}
 
-	if p.TrialEndNow != nil {
+	if BoolValue(p.TrialEndNow) {
```
remi-stripe: @brandur-stripe I think I missed this when I changed the AppendTo. Is this the correct way to do it? Those are custom values and we only send the custom string if the boolean is true I think. If so, we likely need to review all the AppendTo for similar mistakes.

invoice.go
```diff
@@ -6,125 +6,131 @@ import "encoding/json"
...
 // For more details see https://stripe.com/docs/api#invoice_lines.
 type InvoiceLineListParams struct {
 	ListParams `form:"*"`
 
-	Customer string `form:"customer"`
+	Customer *string `form:"customer"`
 
 	// ID is the invoice ID to list invoice lines for.
-	ID string `form:"-"` // Goes in the URL
+	ID *string `form:"-"` // Goes in the URL
```
remi-stripe: @brandur-stripe Should we remove this and change the prototype of the method to take the id as a parameter instead?

sub.go
brandur-stripe: I don't really think it's great, but I don't have a better way of doing it right now.

Because TrialEnd is normally an integer, the only way we could have it allow the value `now` is to make it something like an `interface{}` which isn't great.

invoice.go
brandur-stripe: Good call ... that would probably be the better interface. If we did it though, we'd probably want to change other places that get a URL parameter based a `Params` struct though — application fee refund's `New` for example.

Looking around, there are quite a few of these (I used `ag '/%v/'` as a heuristic to find them), and changing them all might be a project of its own. In the spirit of changing less, I could see just leaving them as is, but I don't feel super strongly either way. Thoughts?

account.go
brandur-stripe: One possibility is that we could have users explicitly set an empty slice:

``` go
AdditionalOwners: []*AdditionalOwnerParams{},
```

Then in either a custom `AppendTo` or even directly in `form` (given that emptying an array with an empty string is probably a reusable convention) we could detect the slice's zero length and encode it as an empty string. It's a little weird, but I think it might be slightly better than having the companion `AdditionalOwnersEmpty`. What do you think?

invoice.go
remi-stripe: I think there's a difference when the parameter is just the ID versus when the parameter is dependent on the type of resource like `/v1/customers/xxx/sources/yyy` insread of `/v1/recipients/xxx/sources/yyy`.
But I'm fine leaving as is for now

sub.go
remi-stripe: @brandur-stripe  would you be able to go through each AppendTo and make sure I did not miss any edge-case with those if?

account.go
remi-stripe: Sorry missed that one. Both would work. Since this is going away in X weeks though it might be easier to just leave as is. What do you think is better?

brandur-stripe: Yeah good point. Let's just leave it for now then.

invoice.go
brandur-stripe: Yeah, let's just leave it for the time being then.

sub.go
brandur-stripe: Sure!

I went through and these are the ones that I found that have a similar pattern:

* `UpToInf` on `PlanTierParams`.
* `BillingCycleAnchorNow` on `SubscriptionParams`.
* `BillingCycleAnchorUnchanged` on `SubscriptionParams`.
* `DelayDaysMinimum` on `PayoutScheduleParams`.

I guess one possibility is we could have a "special" field for these that would allow you to pass a special value for these types of fields:

``` go
PlanTierParams {
    UpToSpecial: stripe.String(stripe.UpToMinimum),
}
```

You could also call this `UpToString` or `UpToEnum`.

IMO though, this isn't too much better or worse than what's already implemented, so I'd probably just leave it. What do you think?

brandur-stripe:
ptal @remi-stripe

feerefund.go
```diff
@@ -4,48 +4,48 @@ import (
 	"encoding/json"
 )
 
-// FeeRefundParams is the set of parameters that can be used when refunding a fee.
+// ApplicationFeeRefundParams is the set of parameters that can be used when refunding an application fee.
```
remi-stripe: @brandur-stripe I realized tonight that I renamed to `ApplicationFeeRefund*` but the resource is really `fee_refund` and called FeeRefund elsewhere. Should I undo this one?

remi-stripe:
@brandur-stripe Okay as crazy as it seems, I think we're done. The PR internally to update all the Go examples is also ready. So if all goes well we can merge next week?

feerefund.go
brandur-stripe: I'd just leave it — it's called "application fee refund" in both its name in the reference documentation (i.e., on the left navbar) and in Stripe.NET, and "application fee refund" is the superior name in general. It groups well with "application fee" and describes exactly what it is.

brandur-stripe:
> Okay as crazy as it seems, I think we're done. The PR internally to update all the Go examples is also ready. So if all goes well we can merge next week?

Yep, let's do it!

feerefund.go
brandur-stripe: Actually, on second thought — I think it kind of sucks how the struct is `ApplicationFeeRefund` but the package is `feerefund`. Given the value of preferring to leave as is where possible, I'd be okay renaming it back to `FeeRefund`.

remi-stripe: @brandur-stripe fixed, PTAL

brandur-stripe:
LGTM.

brandur-stripe:
Hey @remi-stripe, just a heads up: I pushed a commit to your branch here to fix the tests for compatibility with stripe-mock 0.17. If you rebase again, just make sure to slurp it in as well.

brandur-stripe:
@remi-stripe And I just pushed one more commit that moves all fields in `Params` and `ListParams` over to pointers as well for consistency with other parameter structures. Can you take a look when you get a chance? Thanks!

brandur-stripe:
I also added `Params.SetIdempotencyKey` for consistency with `Params.SetStripeAccount`.

My one thought on that is that this API for setting a standard random idempotency key is a little awkward:

``` go
params.SetIdempotencyKey(stripe.NewIdempotencyKey())
```

I think it's okay to have both those functions (we should definitely leave the latter in place for backwards compatibility), but we might also want to consider a convenience function to do this very common operation like:

``` go
params.SetIdempotencyKeyRandom()
```

(Or something like that.) We could add that separately too.

params.go
```diff
@@ -75,24 +75,24 @@ type ListParams struct {
...
+	EndingBefore *string   `form:"ending_before"`
+	Expand       []*string `form:"expand"`
+	Filters      Filters   `form:"*"`
+	Limit        *int64    `form:"limit"`
 
 	// Single specifies whether this is a single page iterator. By default,
 	// listing through an iterator will automatically grab additional pages as
 	// the query progresses. To change this behavior and just load a single
 	// page, set this to true.
 	Single bool `form:"-"` // Not an API parameter
```
remi-stripe: shouldn't this one be `*bool` too?

params.go
```diff
@@ -162,14 +162,19 @@ func NewIdempotencyKey() string {
...
 
+// SetIdempotencyKey sets a value for the Idempotency-Key header.
+func (p *Params) SetIdempotencyKey(val string) {
+	p.IdempotencyKey = &val
+}
+
 // SetStripeAccount sets a value for the Stripe-Account header.
 func (p *Params) SetStripeAccount(val string) {
-	p.StripeAccount = val
+	p.StripeAccount = &val
```
remi-stripe: OOC why is this better than `String(val)`? I guess they are ultimately the same thing but since I've use `String(xxx)` everywhere I wanted to ask

params_test.go
```diff
@@ -216,20 +216,23 @@ func TestListParams_Expand(t *testing.T) {
 func TestListParams_ToParams(t *testing.T) {
 	listParams := &stripe.ListParams{
 		Context:       context.Background(),
-		StripeAccount: TestMerchantID,
+		StripeAccount: stripe.String(TestMerchantID),
```
remi-stripe: Shouldn't we use `params.SetStripeAccount()` instead here?

brandur-stripe:
> stripe-mock is being weird if it does not accept any of those parameters which work fine in the API :sadpanda:

Yeah, totally agree :/

Unfortunately, some of these API methods have been written so edge-casey that I basically need to go in and find special solutions to fix every edge case. We should continue to do that as we go, but this is the faster solution for now.

params.go
brandur-stripe: I was kind of split on this one ... the reason I didn't change it is that it's not really an API parameter, more like an internal pagination option for the library's iterator. The downside is that it may seem a little inconsistent, but it does have the advantage is that we will never care about the unset versus `false` states. What do you think?

params.go
brandur-stripe: So the reason I didn't bother with `String()` is just that those helpers are basically a workaround for the fact that there's no way to take an address of a literal value. In cases like this where that's not a problem, it seems fine to me to just take the address directly. Thoughts?

params_test.go
brandur-stripe: Yeah, I guess it can't hurt just to demonstrate our suggested usage. I'll change it.

brandur-stripe: Pushed in 4e88ccfc4ed567c8b0776e4c6200e5a667f5c44f.

brandur-stripe:
Alright Remi, I think we're getting pretty close to final release. Want to take one more look?

remi-stripe:
@brandur-stripe I fixed one last example but other than that I think we're ready to merge.

remi-stripe:
@brandur-stripe I just pushed a fix for https://github.com/stripe/stripe-go/issues/565 in https://github.com/stripe/stripe-go/pull/544/commits/0cf81bfe3d271a0bf95f98879389e71b5c7fc242
I tested it with stripe-mock but also with complete end-to-end examples locally that created a customer and a Custom account with 3 cards and 3 bank accounts and I made sure I could list the right ones.

remi-stripe:
@brandur-stripe  just rebased on the latest master.

brandur-stripe:
HERE WE GO!

ob-stripe:
itshappening

brandur-stripe:
LOL! Yep.

Released as 32.0.0!

remi-stripe:
@hiren-bacancy Please work with our support team for help on this instead moving forward, this is a close to 4 years old PR. 

To unblock you quickly the `keys` property was removed back in 2019 as documented [here](https://stripe.com/docs/upgrades#2019-02-19)
> The keys property on Account creation has been removed. Platforms should now authenticate as their connected accounts with their own key via [the Stripe-Account header](https://stripe.com/docs/connect/authentication#stripe-account-header).

Our support team can be contacted here https://support.stripe.com/contact for follow up questions

PR #549, opened 2018-05-02
Update README for Google App Engine

afilina:
Explain how to use the newly created `sc` variable, for those who are new to Go.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	sc := client.New("sk_live_key", stripe.NewBackends(httpClient))
 
 	fmt.Fprintf(w, "Ready to make calls to the Stripe API")
+	// Example: sc.Charges.New(params)
 }
 ```
 


brandur-stripe:
Hi @afilina, yes I think you're right in that this example could be much more clear.

Rather than relegating this to a comment in the code block though, would you mind fleshing this out to be full first-class code? Maybe something like this:

``` go
  Amount: 2000,
  Currency: "usd",
  Desc: "Charge from Google App Engine",
}
chargeParams.SetSource("tok_amex") // obtained with Stripe.js
charge, err := charge.New(chargeParams)
```

Also, I think the `fmt.Fprintf(w, "Ready to make calls to the Stripe API")` that's already in there really doesn't add any useful information, so maybe just get rid of it.

Thanks!

afilina:
@brandur-stripe Tested this code snippet locally with a valid API key. I think that this example should also be published on the official docs.

afilina:
Perhaps not the best place to report it, but the Go example on this page uses `Description` instead of `Desc`. https://stripe.com/docs/quickstart

PR #550, opened 2018-05-04
Append standard `Params` in `AppendToAsCardSourceOrExternalAccount`

brandur:
`CardParams` has a special `AppendTo` implementation that's used in
certain cases to properly massage itself into the calls of certain API
methods.

The default `AppendTo` implementation will recurse into substructs like
`Params`, but this custom one will not. This means that although it was
correctly encoding card data, it wouldn't encode common fields like
`Metadata`.

This patch adds an additional encoding call to make sure that standard
fields in `Params` are also encoded even when calling this custom
implementation.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/card.go
+++ b/card.go
@@ -71,6 +71,8 @@ type CardParams struct {
 // because the cards endpoint is a little unusual. There is one other resource
 // like it, which is bank account.
 func (c *CardParams) AppendToAsCardSourceOrExternalAccount(body *form.Values, keyParts []string) {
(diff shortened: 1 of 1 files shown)


remi-stripe:
LGTM but coveralls is unhappy with you!

brandur-stripe:
Thanks Remi! Coveralls seems to have changed its mind, and we're now green :)

brandur-stripe:
Released as 30.6.1.

PR #551, opened 2018-05-04
Rename EndPeriod to CancelAtPeriodEnd

fred-stripe:
Looks like this flag was also not implemented, but it works now!

r? @stripe/api-libraries

### DIFF
--- a/sub.go
+++ b/sub.go
@@ -22,6 +22,7 @@ type SubParams struct {
 	BillingCycleAnchor          int64             `form:"billing_cycle_anchor"`
 	BillingCycleAnchorNow       bool              `form:"-"` // See custom AppendTo
 	BillingCycleAnchorUnchanged bool              `form:"-"` // See custom AppendTo
+	CancelAtPeriodEnd           bool              `form:"cancel_at_period_end"`
 	Card                        *CardParams       `form:"card"`
 	Coupon                      string            `form:"coupon"`
 	CouponEmpty                 bool              `form:"coupon,empty"`
@@ -100,7 +101,7 @@ type Sub struct {
 	Customer           *Customer         `json:"customer"`
 	DaysUntilDue       uint64            `json:"days_until_due"`
 	Discount           *Discount         `json:"discount"`
-	EndCancel          bool              `json:"cancel_at_period_end"`
+	CancelAtPeriodEnd  bool              `json:"cancel_at_period_end"`
 	Ended              int64             `json:"ended_at"`
 	FeePercent         float64           `json:"application_fee_percent"`
 	ID                 string            `json:"id"`


sub.go
```diff
@@ -64,6 +64,10 @@ func (p *SubParams) AppendTo(body *form.Values, keyParts []string) {
 	if p.TrialEndNow {
 		body.Add(form.FormatKey(append(keyParts, "trial_end")), "now")
 	}
+
+	if p.CancelAtPeriodEnd {
+		body.Add(form.FormatKey(append(keyParts, "cancel_at_period_end")), "true")
+	}
```
remi-stripe: are you sure this is required? The whole form encoding logic should do this automatically and there should not be a need for this.

fred-stripe: Yep, the tests fail without it, so it appears to not get generated by default.

fred-stripe:
This is specifically to make the Billing upgrade/downgrade example from the docs work so I'd rather merge it now.

On the other hand this has been not implemented for a while and no one noticed… and it won't be released until we release a new version. If you'd rather pull this into your PR I'm peaceful with closing this.

r? @remi-stripe

remi-stripe:
As discussed out of band you are incorrectly renaming a parameter that is still valid. We have `at_period_end` on The Cancel Subscription API and `cancel_at_period_end` on update. So `SubParams` needs both parameters

fred-stripe:
Narrowed this change to only add support for updates & retrieval of `cancel_at_period_at` field on subscriptions. Left existing `EndCancel` in place so that deleting a subscription with `at_period_end` is also possible.

r? @remi-stripe

sub.go
```diff
@@ -100,7 +101,7 @@ type Sub struct {
 	Customer           *Customer         `json:"customer"`
 	DaysUntilDue       uint64            `json:"days_until_due"`
 	Discount           *Discount         `json:"discount"`
-	EndCancel          bool              `json:"cancel_at_period_end"`
+	CancelAtPeriodEnd  bool              `json:"cancel_at_period_end"`
```
remi-stripe: I would just leave this one as is so that we don't have ship it as a major revision

sub.go
```diff
@@ -22,6 +22,7 @@ type SubParams struct {
 	BillingCycleAnchor          int64             `form:"billing_cycle_anchor"`
 	BillingCycleAnchorNow       bool              `form:"-"` // See custom AppendTo
 	BillingCycleAnchorUnchanged bool              `form:"-"` // See custom AppendTo
+	CancelAtPeriodEnd           bool              `form:"cancel_at_period_end"`
```
remi-stripe: the issue here is that this only works for the value `true`. It does not work for false as it will be ignored. You have to do an ugly hack like `NoProrate` in that same struct.

remi-stripe:
I meant months, not minutes, obviously.

remi-stripe:
Flagging that I ended up doing this one in https://github.com/stripe/stripe-go/pull/544/commits/420725cc4b6a6dd85fd4cbddfa9b943529b8a729 because I was working on updating the docs and the library needed to support it.

remi-stripe:
closing as the fix was merged in a separate PR

PR #555, opened 2018-05-16
Fix some parameters being sent in tests

brandur:
I found a bug recently in stripe-mock which causes it not to actually be
validating that parameters not in the spec are not being sent (the
actual Stripe API does check for this).

After applying a fix, I found that stripe-go's test suite no longer
passes against it, and the reason is that there are some subtle mistakes
throughout. This patch corrects them to be in line with what the API
actually expects.

cc @stripe/api-libraries


remi-stripe:
LGTM

brandur-stripe:
Thanks Remi!

Issue #529, opened 2018-03-01
Implement a retry scheme for intermittent failures in `stripe-go`


brandur-stripe:
Libraries like `stripe-php` and `stripe-ruby` implement a scheme that allows them to automatically retry requests that look like they failed because of an intermittent problem. Requests occasionally fail with `stripe-go` too, and we've seen some evidence of this in issues like #528. We should have a retry scheme here too.

Like the aforementioned libraries, we should be smart about it by generating idempotency keys and implementing exponential backoff and jitter.

Issue #528, opened 2018-03-01
Requests randomly fail with EOF


yns01:
Hello,

Some of our requests fail with an EOF error:

**Logs**

I set the log level for the stripe package to the maximum value, 3. Here’s a timeline
```
"2018/02/28 21:06:24 Requesting POST api.stripe.com/v1/customers"
"2018/02/28 21:06:24 Completed in 65.481468ms"
"2018/02/28 21:06:24 Request to Stripe failed: Post https://api.stripe.com/v1/customers: EOF"
```

**go version go1.9.4 linux/amd64**
**stripe-go version = "v28.8.0"**

65ms seems way too fast to make an actual request to your servers (from EU, POST /customers takes usually more than 1s); also I could not find any request around that time on the dashboard. So I’m assuming that the request did not reach Stripe.

```
res, err := s.HTTPClient.Do(req)
if err != nil {
    if LogLevel > 0 {
        Logger.Printf("Request to Stripe failed: %v\n", err)
    }
    return err
}
```
Thanks!

brandur-stripe:
Hi @yns01, thanks for including the extra logging information!

Given a network that's not perfectly reliable, some level of request failure is to be expected. Unless you're noticing issues on a problematically high number of requests, there probably isn't a bug here.

Some of our client libraries are built to retry requests automatically when they detect a failure that looks like it's intermittent like a timeout, socket error, or EOF, but `stripe-go` doesn't yet have this functionality. In the meantime, you might want to consider implementing your own basic wrapper to do the job. (It's relatively easy, and you can use [existing code in `stripe-ruby` as a reference](https://github.com/stripe/stripe-ruby/blob/3bc4256e2532d75e93ca744edc3dc7913aa79a86/lib/stripe/stripe_client.rb#L219-L223).

The one key thing to remember when implementing a retry scheme is that for any mutating operation, [an idempotency key must be used](https://stripe.com/docs/api#idempotent_requests). This guarantees that the operation won't be duplicated even if our server's received and processed the original (even if it looked like an error to the client).

Having a retry strategy is usually good practice for any external services that you're integrated with, and Stripe's no different. I hope we'll get to implementing one in `stripe-go` soon, and I've just opened #529 so that you'll be able to tell when we do.

yns01:
Hey @brandur-stripe, thank you for your prompt reply.

We'll probably have some retry in place. We already use idempotency keys so we're covered there!

brandur-stripe:
No worries!

[...]

PR #557, opened 2018-05-21
add MaxNetworkRetries parameter to backend configuration

Compaurum:
implement mechanism for retrying failure request caused by network errors.
Solves problem described here https://github.com/stripe/stripe-go/issues/528 and here https://github.com/stripe/stripe-go/issues/529


stripe.go
```diff
@@ -283,6 +296,8 @@ func (s *BackendConfiguration) NewRequest(method, path, key, contentType string,
 			}
 
 			req.Header.Add("Idempotency-Key", idempotency)
+		} else if strings.ToUpper(req.Method) == http.MethodPost || strings.ToUpper(req.Method) == http.MethodPut || strings.ToUpper(req.Method) == http.MethodDelete {
```
brandur-stripe: Given that the use of `PUT` is somewhat speculative already, I wonder if we should just add `PATCH` in here as well for good measure.

stripe.go
```diff
@@ -315,7 +330,24 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
...
 	start := time.Now()
 
-	res, err := s.HTTPClient.Do(req)
+	var res *http.Response
+	var err error
+	for retry := 0; ; {
+		res, err = s.HTTPClient.Do(req)
+		if s.shouldRetry(err, res, retry) {
+			sleep := s.sleepTime(retry)
+			if LogLevel > 2 {
```
brandur-stripe: I'm kind of thinking that maybe this should be `LogLevel > 0`. Errors that occur here are still errors and should probably appear in the logs.

stripe.go
```diff
@@ -485,3 +517,44 @@ func initUserAgent() {
...
+	return false
+}
+
+//sleepTime calculates sleeping/delay time in milliseconds between failure and a new one request.
+func (s *BackendConfiguration) sleepTime(numRetries int) (delay int) {
+	// Apply exponential backoff with minNetworkRetryDelay on the
+	// number of num_retries so far as inputs.
+	delay = minNetworkRetryDelay + minNetworkRetryDelay*numRetries*numRetries
+
+	//Do not allow the number to exceed maxNetworkRetryDelay.
```
brandur-stripe: Mind just editing this to have a space after the `//`?

stripe.go
```diff
@@ -485,3 +517,44 @@ func initUserAgent() {
...
+	if delay > maxNetworkRetryDelay {
+		delay = maxNetworkRetryDelay
+	}
+
+	//Apply some jitter by randomizing the value in the range of 75%-100%.
```
brandur-stripe: And same here.

stripe.go
```diff
@@ -44,6 +45,11 @@ const TotalBackends = 2
 // one from `uname`.
 const UnknownPlatform = "unknown platform"
 
+// maxNetworkRetryDelay and minNetworkRetryDelay defines sleep time in milliseconds between
+// tries to send HTTP request again after network failure.
+const maxNetworkRetryDelay = 5000
+const minNetworkRetryDelay = 500
```
brandur-stripe: So I think just for maximum correctness and because Go supports this natively, we should use `time.Duration` for both of these. So something like `5000 * time.Millisecond`.

stripe.go
```diff
@@ -72,13 +78,15 @@ func (a *AppInfo) formatUserAgent() string {
 type Backend interface {
 	Call(method, path, key string, body *form.Values, params *Params, v interface{}) error
 	CallMultipart(method, path, key, boundary string, body io.Reader, params *Params, v interface{}) error
+	SetMaxNetworkRetry(maxNetworkRetry int)
```
brandur-stripe: It seems that using "retries" instead of "retry" would make this more fluid, and also be consistent with `MaxNetworkRetries` in `BackendConfiguration` below. Any reason you went with the singular version of the word here?

stripe.go
```diff
@@ -485,3 +517,44 @@ func initUserAgent() {
...
+		return true
+	}
+	return false
+}
+
+//sleepTime calculates sleeping/delay time in milliseconds between failure and a new one request.
+func (s *BackendConfiguration) sleepTime(numRetries int) (delay int) {
+	// Apply exponential backoff with minNetworkRetryDelay on the
+	// number of num_retries so far as inputs.
+	delay = minNetworkRetryDelay + minNetworkRetryDelay*numRetries*numRetries
```
brandur-stripe: It looks like you were using stripe-ruby's implementation as a template to build this, but that you deviated in a few ways from it  — would you mind just explaining the rationale for why? It's not too big of a deal, but of course maintenance will be slightly easier if all of these are using the same algorithm.

Also, it might be somewhat nice just to make a note in here that we expect the delay in here to be measured in milliseconds and how given that, it's probably okay to calculate randomness with an integer.

stripe.go
```diff
@@ -485,3 +517,44 @@ func initUserAgent() {
...
+	}
+
+	if resp.StatusCode == http.StatusConflict {
+		return true
+	}
+	return false
+}
+
+//sleepTime calculates sleeping/delay time in milliseconds between failure and a new one request.
+func (s *BackendConfiguration) sleepTime(numRetries int) (delay int) {
```
brandur-stripe: Can you fix the spacing after the `//` here? And similar to the constants above, we should have this return a `time.Duration` instead of an integer. Lastly, is there a good reason to make this a named return value? I'd prefer to avoid those if possible because they're somewhat confusing unless you already know how they work.

stripe.go
```diff
@@ -485,3 +517,44 @@ func initUserAgent() {
...
+	// number of num_retries so far as inputs.
+	delay = minNetworkRetryDelay + minNetworkRetryDelay*numRetries*numRetries
+
+	//Do not allow the number to exceed maxNetworkRetryDelay.
+	if delay > maxNetworkRetryDelay {
+		delay = maxNetworkRetryDelay
+	}
+
+	//Apply some jitter by randomizing the value in the range of 75%-100%.
+	jitter := rand.Intn(delay / 4)
```
brandur-stripe: Also looks like a bit of deviation from the stripe-ruby implementation. It's not the end of the world, but is there any particular reason to choose `4` here?

brandur-stripe:
(Just adding a comment here because we're currently seeing a false positive on our response checking script.)

stripe.go
Compaurum: In my opinion stripe-ruby value (`2`) is too big. Delay should increase smoothlyIn ruby implementation previous
 value could be greater then next. For example possible answer in ruby is:  `6939
15189
18939
26439
24439`.  So on 4-th iteration value is bigger then on 5-th.   Anyway if you want, I will change it - it's not the end of the world :)

stripe.go
Compaurum: I think this logic is more clear for go programmer. Also I ruby version delay time could be bigger then maxDelay (because of jitter). I don't think that it's very good. Just say if it's not clear and I should do the identical logic as in ruby.

stripe.go
Compaurum: ok

stripe.go
Compaurum: @brandur-stripe Add `PATCH` or replace `PUT` with `PATCH`?

brandur-stripe: Add `PATCH`.

stripe.go
brandur-stripe: Yeah agreed that it shouldn't go past max delay. Okay, it's fine to leave as is.

stripe.go
brandur-stripe: Yeah, it's not a big deal. I was just curious where the `4` in particular came from.

brandur-stripe:
@Compaurum Thanks for taking a look at that feedback. Let me know when this is ready for re-review.

stripe.go
```diff
@@ -250,6 +263,7 @@ func (s *BackendConfiguration) CallMultipart(method, path, key, boundary string,
 // NewRequest is used by Call to generate an http.Request. It handles encoding
 // parameters and attaching the appropriate headers.
 func (s *BackendConfiguration) NewRequest(method, path, key, contentType string, body io.Reader, params *Params) (*http.Request, error) {
+	method = strings.ToUpper(method)
```
brandur-stripe: Did you end up putting in the call to `ToUpper` for any particular reason? I just grepped the library and it seems like all verbs are specified as uppercase already, so I don't think we need this.

It also occurred to me that we should really just be using the `http` constants like `http.MethodPut` instead of directly passing verb names. I'll put together a patch that does that separately.

brandur-stripe: And just to be clear, I'd have a preference for dropping `ToUpper` for now if it's not needed.

stripe.go
```diff
@@ -283,6 +297,8 @@ func (s *BackendConfiguration) NewRequest(method, path, key, contentType string,
 			}
 
 			req.Header.Add("Idempotency-Key", idempotency)
+		} else if method == http.MethodPost || method == http.MethodPut || method == http.MethodPatch || method == http.MethodDelete {
```
brandur-stripe: Thanks for making this change!

Now that I look at this again, would you mind creating a new function called `isHTTPWriteMethod` (or such) and abstract this line into it? I think it'll be cleaner code-wise and more suggestive of why this check is being made.

stripe.go
brandur-stripe: I'm moving us to `net/http` constants in #587.

Compaurum:
@brandur-stripe Sorry, I forget to mention you after those changes. I've added function `isHTTPWriteMethod`, removed redundant `ToUpper` and fixed some conflicts with master branch.
Waiting for your call

brandur-stripe:
> @brandur-stripe Sorry, I forget to mention you after those changes. I've added function isHTTPWriteMethod, removed redundant ToUpper and fixed some conflicts with master branch.

No worries! I just feel bad because we ended up leaving you hanging for so long.

Anyway, thanks for the quick changes! This looks great.

brandur-stripe:
Released as 35.0.0.

PR #561, opened 2018-05-23
Remi rename properties use pointers and strings on params types

remi-stripe:
Yet another attempt to make sure the way to use constants is the right direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ Below are a few simple examples:
 
 ```go
 params := &stripe.CustomerParams{
-	Balance: -123,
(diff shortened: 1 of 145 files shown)


remi-stripe:
@brandur-stripe Can you look at the commit https://github.com/stripe/stripe-go/pull/561/commits/1b898841360f9c5a05c16cc826f6f33ef3742892 and confirm this is the direction we want to take? I want to be 100% sure before I re-do every single file, just in case I misunderstood.

remi-stripe:
closing as we agreed on the right method.

PR #564, opened 2018-05-25
Add a V32 migration guide

brandur:
Adds a migration guide for the prospective version 32 of stripe-go which
will include the changes in #544. Some of the breaking changes there are
so sizable that I think we should special-case this with a particular
upgrade guide to help ease user pain somewhat while they're upgrading.

Here we explain some of the background context and make a list of all
the included changes and put in suggestions for remediation.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/v32_migration_guide.md
+++ b/v32_migration_guide.md
@@ -1,0 +1,146 @@
+# V32 Migration Guide
+
+Version 32 of stripe-go contains some very sizable breaking changes.
+
+The major reason that we moved forward on them is that it was previously
+impossible when encoding a parameter struct for an API call to make a
+distinction between a field that a user had left unset versus a field that had
+been set explicitly, but to an empty value. So if we had a parameter struct
+like this:
+
+``` go
+type UsageRecordParams struct {
+	Quantity uint64 `form:"quantity"`
+}
+```
+
+We were unable to differentiate these two cases:
+
+``` go
+// Initialized with no quantity
+UsageRecord {}
+
+// Initialized with an explicitly zero
+UsageRecord {
+    Quantity: 0,
+}
+```
+
+This is because any uninitialized fields on a struct in Go are set to their
+type's "zero value", which for an integer is `0`.
+
+Working around the problem required a secondary field to help explicitly state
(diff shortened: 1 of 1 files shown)


brandur-stripe:
[Here's a rendered version of the guide](https://github.com/stripe/stripe-go/blob/brandur-v31-migration-guide/v31_migration_guide.md).

v31_migration_guide.md
```diff
@@ -0,0 +1,135 @@
...
+  fully expanded form. For example:
+    * `Desc` becomes `Description`.
+    * `Live` becomes `Livemode`.
+* A few names of API resources (and their corresponding parameter and list
+  classes) have changed:
+    * `Fee` becomes `ApplicationFee`.
+    * `FeeRefund` becomes `ApplicationFeeRefund`.
+    * `Sub` becomes `Subscription`.
+    * `SubItem` becomes `SubscriptionItem`.
+    * `Transaction` becomes `BalanceTransaction`.
```
brandur-stripe: Did I miss any here?

remi-stripe: `Owner` is now `AdditionalOwner` though it's not really an API resource so it might be fine to ignore

v31_migration_guide.md
```diff
@@ -0,0 +1,139 @@
...
+    * `TxFee` becomes `BalanceTransactionFee`.
+* Some sets of constants have been renamed and migrated to the top-level
+  `stripe` package. All constants now have a prefix according to what they (for
```
remi-stripe: missing a word after `they`

brandur-stripe:
Thanks Remi.

> Owner is now AdditionalOwner though it's not really an API resource so it might be fine to ignore

Good call. I ended up just putting this in because it doesn't hurt and might help someone.

> missing a word after they

Nice catch. Fixed.

brandur-stripe:
Updated all references (except the branch name) to read V32. (We just released the much more minor #566 breaking change as V31).

brandur-stripe:
Alright, #544 is coming in. Merging this.

PR #566, opened 2018-06-06
Support stripe.DisputeParams for Close dispute API

AlexYaroshenko:
We should accept and pass forward stripe.DisputeParams that will include `StripeAccount` as for any other calls for Connected Accounts
Right now `Close` dispute doesn't work for Connected accounts

```
{
    "error": {
        "code": "resource_missing",
        "doc_url": "https://stripe.com/docs/error-codes/resource-missing",
        "message": "No such dispute: dp_***",
        "param": "dispute",
        "type": "invalid_request_error"
    }
}
```

Works fine if i am doing manual call with one more header field `Stripe-Account`

### DIFF
--- a/dispute/client.go
+++ b/dispute/client.go
@@ -123,13 +123,22 @@ func (c Client) Update(id string, params *stripe.DisputeParams) (*stripe.Dispute
 
 // Close dismisses a dispute in the customer's favor.
 // For more details see https://stripe.com/docs/api#close_dispute.
-func Close(id string) (*stripe.Dispute, error) {
-	return getC().Close(id)
+func Close(id string, params *stripe.DisputeParams) (*stripe.Dispute, error) {
+	return getC().Close(id, params)
 }
 
-func (c Client) Close(id string) (*stripe.Dispute, error) {
+func (c Client) Close(id string, params *stripe.DisputeParams) (*stripe.Dispute, error) {
+	var body *form.Values
+	var commonParams *stripe.Params
+
+	if params != nil {
+		commonParams = &params.Params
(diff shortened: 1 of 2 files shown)


AlexYaroshenko:
@brandur  @brandur-stripe could u take a look?

remi-stripe:
@AlexYaroshenko That change looks good to me! Thanks for making the PR.

Brandur: not sure about the build failure on Go 1.10 exactly and why it happens now.

Do you think we should merge before my big PR? Happy to rebase on that one again.

AlexYaroshenko:
Trivial change that would help me a lot

brandur-stripe:
@remi-stripe I think the 1.10 failure is just intermittent:

> go get -u golang.org/x/tools/cmd/cover
> package golang.org/x/tools/cmd/cover: unrecognized import path "golang.org/x/tools/cmd/cover" (https fetch: Get https://golang.org/x/tools/cmd/cover?go-get=1: net/http: TLS handshake timeout)

I just restarted it.

brandur-stripe:
Thanks Alex.

LGTM.

brandur-stripe:
Released as 31.0.0.

Heads up Alex: we're about to merge a large breaking change from #544 quite soon which will likely be cut as 32.0.0. We released a separate one here for you so that you can get this fix without having to go straight to that, but if you're just pulling latest, you might need to lock your update to a specific version (31.0.0).

AlexYaroshenko:
We are using GLOCK, so i just connected to your 31.0.0 commit ID - should be fine.

PR #568, opened 2018-06-06
doc(webhook): cleanup example

powerman:
There is no need in closing request body.

From https://golang.org/pkg/net/http/#Request:
// The Server will close the request body. The ServeHTTP
// Handler does not need to.


brandur-stripe:
Nice find. Thanks!

PR #569, opened 2018-06-06
Fix unmarshalling an expanded balance transaction source into a balan…

prestonmorgan:
…ce transaction

Currently it's trying to unmarshal into a string, which is causing an error.

Relatedly, in `(*BalanceTransaction) UnmarshalJSON()`, there should be some handling of the unmarshal error beyond assuming that the error is for a particular reason, since that masked this issue, though not sure how y'all want to handle that.

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -100,19 +100,19 @@ type Balance struct {
 // BalanceTransaction is the resource representing the balance transaction.
 // For more details see https://stripe.com/docs/api/#balance.
 type BalanceTransaction struct {
-	Amount      int64                    `json:"amount"`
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks! Pulling this in.

> Relatedly, in (*BalanceTransaction) UnmarshalJSON(), there should be some handling of the unmarshal error beyond assuming that the error is for a particular reason, since that masked this issue, though not sure how y'all want to handle that.

Yeah, agreed. Unfortunately this is a fairly pervasive problem right now, so let's tackle that separately.

brandur-stripe:
Released as 32.0.1.

PR #576, opened 2018-06-07
Fix bucket_size -> divide_by

### LINKED ISSUE #575: PlanTransformUsageParams: DivideBy != BucketSize
Hi, just in the process of upgrading to the latest version with all the big changes and it's super nice to see a lot of alignment between the REST API and the Go structs, great work!

I did notice one discrepency. It may be a non-issue, but I thought I'd ask anyway. [PlanTransformUsageParams](https://godoc.org/github.com/stripe/stripe-go#PlanTransformUsageParams) has a field called `BucketSize`(`bucket_size`) but the API docs call it [`divide_by`](https://stripe.com/docs/api#create_plan-transform_usage-divide_by).

I think they are probably the same thing, but can you verify -- and should it be fixed in this library?

mholt:
Fixes #575. Not much to explain here, methinks. :)

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -127,14 +127,14 @@ type PlanTier struct {
 
 // PlanTransformUsage represents the bucket billing configuration.
 type PlanTransformUsage struct {
-	BucketSize int64                   `json:"bucket_size"`
-	Round      PlanTransformUsageRound `json:"round"`
+	DivideBy int64                   `json:"divide_by"`
+	Round    PlanTransformUsageRound `json:"round"`
 }
 
 // PlanTransformUsageParams represents the bucket billing configuration.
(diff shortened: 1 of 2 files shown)


remi-stripe:
Thanks of making the change. This one is weird, I could have sworn it was called bucket_size instead of divide_by a week ago.

@brandur-stripe while it's a breaking change, it also did not work so we likely can release it as a minor version.

brandur-stripe:
Thanks @mholt!

Remi — agree on not cutting a major for this change. I'll put a note about it in the changelog.

brandur-stripe:
Released as 32.3.0.

mholt:
Great! Thanks guys!

PR #578, opened 2018-06-07
Improve expansion parsing by not discarding unmarshal errors

### LINKED ISSUE #570: Handling errors from UnmarshalJSON
From the conversation here: https://github.com/stripe/stripe-go/pull/569

The UnmarshalJSON methods of some types, like BalanceTransaction, do not handle unexpected errors, which can mask issues like the one linked above.

brandur:
As described in #570, the current pattern in custom `UnmarshalJSON`
implementations used to perform object expansion is quite poor because
it assumes that *any* unmarshal error on an object indicates that the
object was actually a string. This is not true, and these custom
functions can have the effect of hiding actual problems and making
debugging difficult (see #569).

Here I introduce a new pattern for custom `UnmarshalJSON`
implementations. Instead of trying for an object right away, it tries to
detect whether the value is a string, and if it is, sets that as the
object's ID. If it isn't, it then performs unmarshaling on the full
object. If that errors, we return the error.

This also has the benefit of making unmarshaling on values that are
strings faster -- we return almost immediately instead of allocating an
object, trying to decode it, failing, and falling back.

I was hoping to be able to cut down the length of the implementation by
quite a bit, and although it is shortened some, it's not too different.
I'll also try to use this opportunity to standardize on conventions
(name values `v`) and add some tests.

For now, I just redid two implementations to demonstrate the pattern. One easy
one in the form of "charge outcome rule", and one harder one in the form of
"balance transaction source". If this looks good, we can redo them all.

Fixes #570.

r? @remi-stripe Mind taking a look?
cc @stripe/api-libraries

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -140,16 +140,18 @@ type BalanceTransactionFee struct {
 // This custom unmarshaling is needed because the resulting
 // property may be an id or the full struct if it was expanded.
 func (t *BalanceTransaction) UnmarshalJSON(data []byte) error {
-	type bt BalanceTransaction
-	var tt bt
-	err := json.Unmarshal(data, &tt)
-	if err == nil {
-		*t = BalanceTransaction(tt)
-	} else {
-		// the id is surrounded by "\" characters, so strip them
-		t.ID = string(data[1 : len(data)-1])
+	if id, ok := ParseID(data); ok {
+		t.ID = id
+		return nil
 	}
 
+	type balanceTransaction BalanceTransaction
+	var v balanceTransaction
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	*t = BalanceTransaction(v)
 	return nil
 }
 
--- a/balance.go
+++ b/balance.go
@@ -157,40 +157,40 @@ func (t *BalanceTransaction) UnmarshalJSON(data []byte) error {
 // This custom unmarshaling is needed because the specific
 // type of transaction source it refers to is specified in the JSON
 func (s *BalanceTransactionSource) UnmarshalJSON(data []byte) error {
-	type source BalanceTransactionSource
-	var ss source
-	err := json.Unmarshal(data, &ss)
-	if err == nil {
-		*s = BalanceTransactionSource(ss)
-
-		switch s.Type {
-		case BalanceTransactionSourceTypeApplicationFee:
-			err = json.Unmarshal(data, &s.ApplicationFee)
-		case BalanceTransactionSourceTypeCharge:
-			err = json.Unmarshal(data, &s.Charge)
-		case BalanceTransactionSourceTypeDispute:
-			err = json.Unmarshal(data, &s.Dispute)
-		case BalanceTransactionSourceTypePayout:
-			err = json.Unmarshal(data, &s.Payout)
-		case BalanceTransactionSourceTypeRecipientTransfer:
-			err = json.Unmarshal(data, &s.RecipientTransfer)
-		case BalanceTransactionSourceTypeRefund:
-			err = json.Unmarshal(data, &s.Refund)
-		case BalanceTransactionSourceTypeReversal:
-			err = json.Unmarshal(data, &s.Reversal)
-		case BalanceTransactionSourceTypeTransfer:
-			err = json.Unmarshal(data, &s.Transfer)
-		}
-
-		if err != nil {
-			return err
-		}
-	} else {
-		// the id is surrounded by "\" characters, so strip them
-		s.ID = string(data[1 : len(data)-1])
+	if id, ok := ParseID(data); ok {
+		s.ID = id
+		return nil
 	}
 
-	return nil
+	type balanceTransactionSource BalanceTransactionSource
+	var v balanceTransactionSource
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	var err error
+	*s = BalanceTransactionSource(v)
+
+	switch s.Type {
+	case BalanceTransactionSourceTypeApplicationFee:
+		err = json.Unmarshal(data, &s.ApplicationFee)
+	case BalanceTransactionSourceTypeCharge:
+		err = json.Unmarshal(data, &s.Charge)
+	case BalanceTransactionSourceTypeDispute:
+		err = json.Unmarshal(data, &s.Dispute)
+	case BalanceTransactionSourceTypePayout:
+		err = json.Unmarshal(data, &s.Payout)
+	case BalanceTransactionSourceTypeRecipientTransfer:
+		err = json.Unmarshal(data, &s.RecipientTransfer)
+	case BalanceTransactionSourceTypeRefund:
+		err = json.Unmarshal(data, &s.Refund)
+	case BalanceTransactionSourceTypeReversal:
+		err = json.Unmarshal(data, &s.Reversal)
+	case BalanceTransactionSourceTypeTransfer:
+		err = json.Unmarshal(data, &s.Transfer)
+	}
+
+	return err
 }
 
 // MarshalJSON handles serialization of a BalanceTransactionSource.
--- a/dispute.go
+++ b/dispute.go
@@ -160,17 +160,19 @@ type File struct {
 // UnmarshalJSON handles deserialization of a Dispute.
 // This custom unmarshaling is needed because the resulting
 // property may be an id or the full struct if it was expanded.
-func (t *Dispute) UnmarshalJSON(data []byte) error {
-	type dispute Dispute
-	var dd dispute
-	err := json.Unmarshal(data, &dd)
-	if err == nil {
-		*t = Dispute(dd)
-	} else {
-		// the id is surrounded by "\" characters, so strip them
-		t.ID = string(data[1 : len(data)-1])
+func (d *Dispute) UnmarshalJSON(data []byte) error {
+	if id, ok := ParseID(data); ok {
+		d.ID = id
+		return nil
 	}
 
+	type dispute Dispute
+	var v dispute
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	*d = Dispute(v)
 	return nil
 }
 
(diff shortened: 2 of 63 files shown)


balance.go
```diff
@@ -157,40 +157,40 @@ func (t *BalanceTransaction) UnmarshalJSON(data []byte) error {
...
-			err = json.Unmarshal(data, &s.Transfer)
-		}
-
-		if err != nil {
-			return err
-		}
-	} else {
-		// the id is surrounded by "\" characters, so strip them
-		s.ID = string(data[1 : len(data)-1])
+	if id, ok := ParseID(data); ok {
```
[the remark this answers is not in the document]
brandur-stripe: Yeah, it's idiomatic, but you'll see some division in the community as to whether people like it or not :$

I used it here because it makes the implementation a little shorter (and most people are at least relatively used to it), but I'm okay with going the other direction as well.

balance.go
```diff
@@ -157,40 +157,40 @@ func (t *BalanceTransaction) UnmarshalJSON(data []byte) error {
...
+		s.ID = id
+		return nil
 	}
 
-	return nil
+	type balanceTransactionSource BalanceTransactionSource
+	var v balanceTransactionSource
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
```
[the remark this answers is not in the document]
brandur-stripe: Yep :$

The good news is that balance transaction source is a bit of a special case. The bad news is that I'm not sure if there's a way around it with the way we do things.

balance.go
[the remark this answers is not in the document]
remi-stripe: don't change this, I was really asking for my own education. I had never seen code like this, and saw it twice in the same week so I wanted to ask.

balance.go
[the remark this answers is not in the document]
remi-stripe: yeah I think it's fine really, it's likely taking a few milliseconds out of an API request that takes 4/10 seconds itself

remi-stripe:
@brandur-stripe Can you look at the few I just did? I wanted to see if we could make the naming consistent so `v` for the temp variable and `out` for the output of the function.

Also why do we need to do
```
	type balanceTransactionSource BalanceTransactionSource
	var v balanceTransactionSource
```
and can't do this:
```
	var v BalanceTransactionSource
```
I tried and it crashes for some reason.

brandur-stripe:
Thanks Remi!

> @brandur-stripe Can you look at the few I just did? I wanted to see if we could make the naming consistent so v for the temp variable and out for the output of the function.

Yep, looks good!

In terms out `out`, I'd be mildly -1 on this. That variable is actually just the alias of the "instance" struct that this function is going to be operating on is usually not something that you'd set directly like we're doing. It's basically exactly the same as something like this elsewhere in the codebase:

``` go
func (s *BackendConfiguration) Call(method, path, key string, form *form.Values, params *Params, v interface{}) error {
```

It's nominally conventional (I believe) to assign these one or two letter aliases that are suggestive as to the name of the type they're operating on (account = a, balance = b, charge = c, etc.).

> I tried and it crashes for some reason.

Haha, so this is kind of a crazy trick that I kept around from the old implementation.

Basically `UnmarshalJSON` is a special function that you can implement on your type (e.g. `BalanceTransactionSource`) which the `encoding/json` will call automatically when it finds an instance of that type that needs to be unmarshaled.

What's happening here in our case is that if you just called back into `json.Unmarshal(...)` to try and get a `BalanceTransactionSource` again, you'd end up calling back into the same custom `UnmarshalJSON` function that you're already in! You'd thus cause infinite recursion and a stack overflow.

So the trick that's used here:

``` go
	type balanceTransactionSource BalanceTransactionSource
	var v balanceTransactionSource
```

Works around this by creating a new type called `balanceTransactionSource` based off of `BalanceTransactionSource`. This new type _doesn't_ inherit the `UnmarshalJSON` implementation from its parent, and therefore when we call `json.Unmarshal` on it, we _don't_ come back into our custom implementation. Pretty clever, eh?

BTW, thanks a lot for the assist, but I had some of the grunt work done and kind of find this sort of thing soothing every once in a while, haha. Why don't you let me take the rest of this and we can save your sanity for a different repetitive project? :)

remi-stripe:
Wow thanks a lot for the detailed explanation. I never knew there was so much magic in that method. Every time I read the original I was confused and assumed there was some unnecessary boilerplate we never removed but after all it is just some cool workaround to make it behave as needed.

I went ahead and removed my commit so you should be all set and I think I got a handle on the logic now!

brandur-stripe:
No worries, and thanks! Hopefully I can push the rest through sometime today.

brandur-stripe:
@remi-stripe Alright, I converted the rest over and added them all as a second commit.

There are a few minor deviations compared to the patterns established in the first commit that you reviewed, but by and large the vast majority fit into one of those two buckets.

I also tried running a simple test program against the new version of stripe-go to make sure that expanded and non-expanded fields still worked (in case there was a problem with my testing methodology), and it seems to behave exactly like it did before.

Sorry for the line churn, but would you mind taking a quick pass over that second commit? Thanks!

ptal @remi-stripe

dispute.go
```diff
@@ -160,33 +160,37 @@ type File struct {
...
 	}
 
+	*d = Dispute(v)
 	return nil
 }
 
 // UnmarshalJSON handles deserialization of a File.
 // This custom unmarshaling is needed because the resulting
 // property may be an id or the full struct if it was expanded.
 func (f *File) UnmarshalJSON(data []byte) error {
```
remi-stripe: ooc why doesn't this live in `fileupload.go` instead?
Or maybe why isn't Dispute using FileUpload instead of File?

payout_test.go
```diff
@@ -0,0 +1,56 @@
...
+		// `object` field is included so that the source knows what type to
+		// decode
+		data := []byte(`{"id":"ba_123", "object":"bank_account"}`)
+
+		var v PayoutDestination
+		err := json.Unmarshal(data, &v)
+		assert.NoError(t, err)
+		assert.Equal(t, PayoutDestinationTypeBankAccount, v.Type)
+
+		// The source has a field for each possible type, so the charge is
```
remi-stripe: "charge" should be "bank account" and "source" should be "destination"

paymentsource_test.go
```diff
@@ -80,3 +80,30 @@ func TestPaymentSource_MarshalJSON(t *testing.T) {
...
+		// `object` field is included so that the source knows what type to
+		// decode
+		data := []byte(`{"id":"ba_123", "object":"bank_account"}`)
+
+		var v PaymentSource
+		err := json.Unmarshal(data, &v)
+		assert.NoError(t, err)
+		assert.Equal(t, PaymentSourceTypeBankAccount, v.Type)
+
+		// The source has a field for each possible type, so the charge is
```
remi-stripe: minor but "charge" should be "bank account", not sure what "source" should be

payout_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestPayout_UnmarshalJSON(t *testing.T) {
+	// Unmarshals from a JSON string
+	{
+		var v Payout
+		err := json.Unmarshal([]byte(`"tr_123"`), &v)
```
remi-stripe: minor but payout have `po_XXX` ids

refund.go
```diff
@@ -63,15 +63,17 @@ type RefundList struct {
...
-	} else {
-		// the id is surrounded by "\" characters, so strip them
-		r.ID = string(data[1 : len(data)-1])
+	if id, ok := ParseID(data); ok {
+		r.ID = id
+		return nil
 	}
 
+	type sku Refund
+	var v sku
```
remi-stripe: should be refund not sku

refund_test.go
```diff
@@ -3,69 +3,27 @@ package stripe
...
-	if refund.Charge == nil {
-		t.Errorf("Problem deserializing refund, didn't get a Charge")
-	}
 
-	if refund.Charge.ID != "ch_1234" {
-		t.Errorf("Problem deserializing refund.charge, wrong value for ID")
-	}
-}
-
-func TestRefundUnmarshalExpandedCharge(t *testing.T) {
```
remi-stripe: I'm guessing you removed those on purpose because we assume that this is tested by the charge's `UnmarshalJSON`?

source.go
```diff
@@ -209,18 +209,17 @@ func (p *SourceObjectParams) AppendTo(body *form.Values, keyParts []string) {
...
+	*s = Source(v)
 
 	var raw map[string]interface{}
-	err = json.Unmarshal(data, &raw)
-	if err != nil {
+	if err := json.Unmarshal(data, &raw); err != nil {
 		return err
 	}
+
 	if d, ok := raw[s.Type]; ok {
```
remi-stripe: the code at this line still kind of bothers me and feel not really Go agnostic I wonder if we should revisit at some point to list every single Source type explicitly with their own class/struct and such.

sourcetransaction.go
```diff
@@ -33,12 +33,12 @@ type SourceTransaction struct {
 // source transaction.
 func (t *SourceTransaction) UnmarshalJSON(data []byte) error {
 	type sourceTransaction SourceTransaction
-	var st sourceTransaction
-	err := json.Unmarshal(data, &st)
+	var v sourceTransaction
```
remi-stripe: guessing it's expected that this can not just be an ID right?

payout_test.go
brandur-stripe: Wow, can't believe you caught that! Thanks.

refund.go
brandur-stripe: Amazing catch! Thanks.

refund_test.go
brandur-stripe: Yeah, I think what's in the `UnmarshalJSON` is "good enough". It tests the important parts of unmarshaling to an object, but more succinctly.

source.go
brandur-stripe: Man, I was totally thinking exactly the same thing ...

Especially when we have things like balance transaction sources, etc. modeled to have one field per possible expansion type, this should really be the same thing.

I think the idea at the time is that there was still too many changes happening in sources to be able to model them accurately, but the API should be stable by now.

sourcetransaction.go
brandur-stripe: Yeah exactly. It was like this before, so I just left it.

dispute.go
brandur-stripe: Haha, I don't know! That's where I found it.

I think probably what happened is that originally files were only needed on disputes, so they were just implemented in the same file. When `FileUpload` was added, there was just no attempt to reconcile them.

For now I've moved `File` and associated tests into `fileupload.go` to make it a little easier to find in bb5e87a. I don't want to reconcile the two PRs here because `MIMEType` changing to `Type` is technically a breaking change, but that should be easy enough to do separately (too bad we didn't catch it before V33 this morning!).

Sound good?

brandur-stripe:
Thanks for the quick review Remi! All feedback addressed. Want to take one more quick glance before shipping? Thanks!

brandur-stripe:
Excellent. Thanks Remi!

brandur-stripe:
Released as 33.1.0.

PR #579, opened 2018-06-08
format the example README code

harlow:
Looks like the copy+paste of example code lost it's formatting

* Indent example to code to match gofmt


remi-stripe:
@harlow Thank you for the fixes!

PR #581, opened 2018-06-09
Push parameter encoding into `BackendConfiguration.Call`

brandur:
Now that we have a little more consistency throughout the library, here
we modify `BackendConfiguration.Call` (temporarily renamed to `Call2`
while we refactor) so that it can take parameter structs directly, then
encode them. This allows us to remove this common boilerplate from every
API call throughout the entire library:

``` go
if params != nil {
    commonParams = &params.Params
    body = &form.Values{}
    form.AppendTo(body, params)
}
```

Temporarily though, only the `charge/` package has been converted over
to show what it looks like before we convert everything.

There is a little bit of type trickiness that allows this to happen, and
which requires a somewhat advanced understanding of some Go language
semantics. We define a `ParamsContainer` interface as this:

``` go
type ParamsContainer interface {
    GetParams() *Params
}
```

Then on the `Params` struct itself, we define an implementation:

``` go
func (p *Params) GetParams() *Params {
    return p
}
```

In Go, whenever a struct is embedded in another struct, the host
inherits the function implementations of the struct which was embedded.
Because every parameter struct in the library embeds `Params`, they all
get a `GetParams` implementation and thus implement `ParamsContainer`
automatically:

``` go
type ChargeParams struct {
    Params `form:"*"`
    ...
}
```

This allows us to take a `ParamsContainer` from `Call`.

Most of this applies similarly for `ListParams`, and similarly `iter.go`
also gets a `Query2` and `GetIter2` to keep everything compiling while
we refactor everything. I did a little more refactoring in `iter.go` as
well just because there was quite a few very bad names in there. If this
patch lands, I'll probably refactor even more of it in a follow up.

r? @remi-stripe I got the idea to do this from a chat we had the other day. I
think it'll tighten up the implementation even a little more and cut quite a
few LOCs.

cc @stripe/api-libraries

### DIFF
--- a/account/client.go
+++ b/account/client.go
@@ -17,19 +17,8 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
 }
 
 func (c Client) New(params *stripe.AccountParams) (*stripe.Account, error) {
-	body := &form.Values{}
-
-	// Type is now required on creation and not allowed on update
-	// It can't be passed if you pass `from_recipient` though
-	if params.FromRecipient != nil {
-		body.Add("type", stripe.StringValue(params.Type))
-	}
-
-	form.AppendTo(body, params)
-
 	acct := &stripe.Account{}
-	err := c.B.Call("POST", "/accounts", c.Key, body, &params.Params, acct)
-
+	err := c.B.Call2("POST", "/accounts", c.Key, params, acct)
 	return acct, err
 }
 
@@ -40,8 +29,7 @@ func Get() (*stripe.Account, error) {
 
 func (c Client) Get() (*stripe.Account, error) {
 	account := &stripe.Account{}
-	err := c.B.Call("GET", "/account", c.Key, nil, nil, account)
-
+	err := c.B.Call2("GET", "/account", c.Key, nil, account)
 	return account, err
 }
 
@@ -51,18 +39,9 @@ func GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 }
 
 func (c Client) GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
-	var body *form.Values
-	var commonParams *stripe.Params
-
-	if params != nil {
-		commonParams = &params.Params
-		body = &form.Values{}
-		form.AppendTo(body, params)
-	}
-
+	path := stripe.FormatURLPath("/accounts/%s", id)
 	account := &stripe.Account{}
-	err := c.B.Call("GET", stripe.FormatURLPath("/accounts/%s", id), c.Key, body, commonParams, account)
-
+	err := c.B.Call2("GET", path, c.Key, params, account)
 	return account, err
 }
 
@@ -72,18 +51,9 @@ func Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 }
 
 func (c Client) Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
-	var body *form.Values
-	var commonParams *stripe.Params
-
-	if params != nil {
-		commonParams = &params.Params
-		body = &form.Values{}
-		form.AppendTo(body, params)
-	}
-
+	path := stripe.FormatURLPath("/accounts/%s", id)
 	acct := &stripe.Account{}
-	err := c.B.Call("POST", stripe.FormatURLPath("/accounts/%s", id), c.Key, body, commonParams, acct)
-
+	err := c.B.Call2("POST", path, c.Key, params, acct)
 	return acct, err
 }
 
@@ -93,18 +63,9 @@ func Del(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 }
 
 func (c Client) Del(id string, params *stripe.AccountParams) (*stripe.Account, error) {
-	var body *form.Values
-	var commonParams *stripe.Params
-
-	if params != nil {
-		body = &form.Values{}
-		form.AppendTo(body, params)
-		commonParams = &params.Params
-	}
-
+	path := stripe.FormatURLPath("/accounts/%s", id)
 	acct := &stripe.Account{}
-	err := c.B.Call("DELETE", stripe.FormatURLPath("/accounts/%s", id), c.Key, body, commonParams, acct)
-
+	err := c.B.Call2("DELETE", path, c.Key, params, acct)
 	return acct, err
 }
 
@@ -114,13 +75,9 @@ func Reject(id string, params *stripe.AccountRejectParams) (*stripe.Account, err
 }
 
 func (c Client) Reject(id string, params *stripe.AccountRejectParams) (*stripe.Account, error) {
-	body := &form.Values{}
-	if params.Reason != nil {
-		body.Add("reason", stripe.StringValue(params.Reason))
-	}
+	path := stripe.FormatURLPath("/accounts/%s/reject", id)
 	acct := &stripe.Account{}
-	err := c.B.Call("POST", stripe.FormatURLPath("/accounts/%s/reject", id), c.Key, body, nil, acct)
-
+	err := c.B.Call2("POST", path, c.Key, params, acct)
 	return acct, err
 }
 
@@ -129,22 +86,10 @@ func List(params *stripe.AccountListParams) *Iter {
 	return getC().List(params)
 }
 
-func (c Client) List(params *stripe.AccountListParams) *Iter {
-	var body *form.Values
-	var lp *stripe.ListParams
-	var p *stripe.Params
-
-	if params != nil {
-		body = &form.Values{}
-
-		form.AppendTo(body, params)
-		lp = &params.ListParams
-		p = params.ToParams()
-	}
-
-	return &Iter{stripe.GetIter(lp, body, func(b *form.Values) ([]interface{}, stripe.ListMeta, error) {
+func (c Client) List(listParams *stripe.AccountListParams) *Iter {
+	return &Iter{stripe.GetIter2(listParams, func(p *stripe.Params, b *form.Values) ([]interface{}, stripe.ListMeta, error) {
 		list := &stripe.AccountList{}
-		err := c.B.Call("GET", "/accounts", c.Key, b, p, list)
+		err := c.B.CallRaw("GET", "/accounts", c.Key, b, p, list)
 
 		ret := make([]interface{}, len(list.Data))
 		for i, v := range list.Data {
--- a/account/client.go
+++ b/account/client.go
@@ -18,7 +18,7 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
 
 func (c Client) New(params *stripe.AccountParams) (*stripe.Account, error) {
 	acct := &stripe.Account{}
-	err := c.B.Call2("POST", "/accounts", c.Key, params, acct)
+	err := c.B.Call("POST", "/accounts", c.Key, params, acct)
 	return acct, err
 }
 
@@ -29,7 +29,7 @@ func Get() (*stripe.Account, error) {
 
 func (c Client) Get() (*stripe.Account, error) {
 	account := &stripe.Account{}
-	err := c.B.Call2("GET", "/account", c.Key, nil, account)
+	err := c.B.Call("GET", "/account", c.Key, nil, account)
 	return account, err
 }
 
@@ -41,7 +41,7 @@ func GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 func (c Client) GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	account := &stripe.Account{}
-	err := c.B.Call2("GET", path, c.Key, params, account)
+	err := c.B.Call("GET", path, c.Key, params, account)
 	return account, err
 }
 
@@ -53,7 +53,7 @@ func Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 func (c Client) Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	acct := &stripe.Account{}
-	err := c.B.Call2("POST", path, c.Key, params, acct)
+	err := c.B.Call("POST", path, c.Key, params, acct)
 	return acct, err
 }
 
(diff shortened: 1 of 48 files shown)


charge/client.go
```diff
@@ -35,18 +31,9 @@ func Get(id string, params *stripe.ChargeParams) (*stripe.Charge, error) {
...
-		body = &form.Values{}
-		form.AppendTo(body, params)
-	}
-
+	path := stripe.FormatURLPath("/charges/%s", id)
 	charge := &stripe.Charge{}
-	err := c.B.Call("GET", stripe.FormatURLPath("/charges/%s", id), c.Key, body, commonParams, charge)
-
+	err := c.B.Call2("GET", path, c.Key, params, charge)
 	return charge, err
```
brandur-stripe: I wish this was all a little more readable, but this is essentially a demonstration of the golden sample: almost every new/get/update/other (non-list) call becomes four lines of code.

iter.go
```diff
@@ -9,6 +9,9 @@ import (
 // Query is the function used to get a page listing.
 type Query func(*form.Values) ([]interface{}, ListMeta, error)
 
+// Query2 is the function used to get a page listing.
+type Query2 func(*Params, *form.Values) ([]interface{}, ListMeta, error)
```
brandur-stripe: As with `Call2` and `GetIter2`, `Query2` becomes `Query` as soon as we finish refactoring everything. It'll never hit `master`.

iter.go
```diff
@@ -18,39 +21,76 @@ type Query func(*form.Values) ([]interface{}, ListMeta, error)
...
-	cur    interface{}
-	err    error
-	meta   ListMeta
-	params ListParams
-	qs     *form.Values
-	query  Query
-	values []interface{}
+	cur        interface{}
+	err        error
+	formValues *form.Values
```
brandur-stripe: Naming like `qs` was terrible, so renamed a few of these to be more obvious.

stripe.go
```diff
@@ -72,6 +73,8 @@ func (a *AppInfo) formatUserAgent() string {
 // This interface exists to enable mocking for during testing if needed.
 type Backend interface {
 	Call(method, path, key string, body *form.Values, params *Params, v interface{}) error
+	Call2(method, path, key string, params ParamsContainer, v interface{}) error
+	CallRaw(method, path, key string, body *form.Values, params *Params, v interface{}) error
```
brandur-stripe: `Call2` will become `Call`. `CallRaw` needs to stick around for now so that we can still implement the iterator closures in our list functions (we may be able to simplify those in the future too, but I haven't looked into it yet).

charge/client.go
```diff
@@ -101,21 +69,10 @@ func List(params *stripe.ChargeListParams) *Iter {
...
-
-	return &Iter{stripe.GetIter(lp, body, func(b *form.Values) ([]interface{}, stripe.ListMeta, error) {
+func (c Client) List(listParams *stripe.ChargeListParams) *Iter {
+	return &Iter{stripe.GetIter2(listParams, func(p *stripe.Params, b *form.Values) ([]interface{}, stripe.ListMeta, error) {
 		list := &stripe.ChargeList{}
-		err := c.B.Call("GET", "/charges", c.Key, b, p, list)
+		err := c.B.CallRaw("GET", "/charges", c.Key, b, p, list)
 
 		ret := make([]interface{}, len(list.Data))
 		for i, v := range list.Data {
```
brandur-stripe: List functions are still shorter than before (lots of the frontmatter is removed), but they keep their closures for now, so they're not down to ~4 lines like some of the other API calls (not yet at least :).

charge/client.go
```diff
@@ -188,8 +145,7 @@ func CloseDispute(id string) (*stripe.Dispute, error) {
 
 func (c Client) CloseDispute(id string) (*stripe.Dispute, error) {
 	dispute := &stripe.Dispute{}
-	err := c.B.Call("POST", stripe.FormatURLPath("/charges/%s/dispute/close", id), c.Key, nil, nil, dispute)
-
+	err := c.B.Call2("POST", stripe.FormatURLPath("/charges/%s/dispute/close", id), c.Key, nil, dispute)
```
remi-stripe: This is an example where we should fix `CloseDispute()` to take a real param instead of us passing a fake one internally. This change will force us to fix all of those.

stripe.go
```diff
@@ -210,6 +213,38 @@ func SetBackend(backend SupportedBackend, b Backend) {
...
+}
+
+// TODO: After every API call uses Call2, remove Call, then rename all
+// instances of Call2 to Call. This only exists as a separate method to keep
+// the build/tests working while we refactor.
+func (s *BackendConfiguration) Call2(method, path, key string, params ParamsContainer, v interface{}) error {
+	var body *form.Values
+	var commonParams *Params
+
+	if params != nil {
```
remi-stripe: we likely could/should enforce that params never be `nil` in the first place right?

charge/client.go
brandur-stripe: Good call. Hm, we mighttt want to consider doing that change separately though given that I think we could push this one through without it being a breaking change.

stripe.go
brandur-stripe: Yeah, we could go that direction too.

I was thinking though for calls like `*.Get` and even `CloseDispute` above, it might be kind of nice to be able to pass a `nil` because while you sometimes care about a parameter and want to pass a structure, you often don't. What do you think?

remi-stripe: That works for me!

brandur-stripe:
> This change is amazing, and it relies on such a cool feature of Go. You've explained that one to me a few times but it had not sinked in until this change the the UnmarshallJSON one.
>
> I like how much cleaner the calls are now and it will make it a lot easier to enforce that params is never nil lower in the stack which will require that all methods take a params (which some do not yet).

Thanks! Glad you like it :) It was my proudest moment as a semi-literate Go developer when I realized that I kind of understood all the ramifications around struct embedding, haha.

stripe.go
remi-stripe: Ah yes I see how passing `nil` is easier/better in that case. I meant was that the method itself should not pass `nil` but ultimately it boils down to the same thing so agreed after all

brandur-stripe: Excellent. I'll continue work on this after #578 is in!

brandur-stripe:
@remi-stripe Sorry for the big patch, but I ended up pushing this one the rest of the way through. Would you mind taking a review pass?

If you like, you can look at each step in a separate commit:

* 7c389d7 converted the rest of the API calls over to the new technique.
* 7130a47 renamed `Call2` back to `Call`, `GetIter2` to `GetIter`, etc.

Thanks!

bankaccount/client.go
```diff
@@ -28,13 +35,7 @@ func (c Client) New(params *stripe.BankAccountParams) (*stripe.BankAccount, erro
...
 
 	ba := &stripe.BankAccount{}
-	var err error
-	if params.Customer != nil {
-		err = c.B.Call("POST", stripe.FormatURLPath("/customers/%s/sources", stripe.StringValue(params.Customer)), c.Key, body, &params.Params, ba)
-	} else {
-		err = c.B.Call("POST", stripe.FormatURLPath("/accounts/%s/external_accounts", stripe.StringValue(params.Account)), c.Key, body, &params.Params, ba)
-	}
-
+	err := c.B.CallRaw("POST", path, c.Key, body, &params.Params, ba)
```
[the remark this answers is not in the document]
brandur-stripe: > ack that sounds fine for now though it's likely worth trying to revisit in the future to make it easier to reason with. It's easy to miss that there are some edge-cases when you build a new module or revamp this one.

Yeah, totally agree. I think with a little more effort we could get this reconciled into the standard framework.

> It might be worth a comment in the code though wdyt?

Yep, good call. I'll add that here and to the implementation in card as well (which uses a similar pattern).

account/client.go
```diff
@@ -51,18 +39,9 @@ func GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
...
-		commonParams = &params.Params
-		body = &form.Values{}
-		form.AppendTo(body, params)
-	}
-
+	path := stripe.FormatURLPath("/accounts/%s", id)
 	account := &stripe.Account{}
-	err := c.B.Call("GET", stripe.FormatURLPath("/accounts/%s", id), c.Key, body, commonParams, account)
-
+	err := c.B.Call("GET", path, c.Key, params, account)
```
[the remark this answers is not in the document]
brandur-stripe: Yeah. Every client package defines its own `Client` which is where a `Key` is configured:

``` go
type Client struct {
	B   stripe.Backend
	Key string
}
```

`Client` has a reference to a `Backend`, but not vice versa, so that backend can't reach back into the client to get at the `Key` field. That's why it's injected with every call.

I don't think I would have designed it this way, and there might be something we can do to refactor to remove even more boilerplate, but I don't want to mess around with this too much for now.

bankaccount/client.go
```diff
@@ -20,6 +20,13 @@ func New(params *stripe.BankAccountParams) (*stripe.BankAccount, error) {
 }
 
 func (c Client) New(params *stripe.BankAccountParams) (*stripe.BankAccount, error) {
+	var path string
+	if params.Customer != nil {
+		path = stripe.FormatURLPath("/customers/%s/sources", stripe.StringValue(params.Customer))
+	} else {
```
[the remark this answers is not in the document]
brandur-stripe: Yeah. I'd basically copied this over from the old implementation, but agreed that it's better and more consistent so I'll change it.

bankaccount/client.go
```diff
@@ -20,6 +20,13 @@ func New(params *stripe.BankAccountParams) (*stripe.BankAccount, error) {
 }
 
 func (c Client) New(params *stripe.BankAccountParams) (*stripe.BankAccount, error) {
+	var path string
+	if params.Customer != nil {
```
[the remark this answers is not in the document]
brandur-stripe: I'll do something a little different and copy over this implementation which is used by `card/`:

``` go
	if params == nil {
		return nil, errors.New("params should not be nil")
	}
```

bankaccount/client.go
```diff
@@ -73,26 +65,17 @@ func Update(id string, params *stripe.BankAccountParams) (*stripe.BankAccount, e
...
+	var path string
 	if params.Customer != nil {
-		err = c.B.Call("POST", stripe.FormatURLPath("/customers/%s/sources/%s", stripe.StringValue(params.Customer), id), c.Key, body, commonParams, ba)
+		path = stripe.FormatURLPath("/customers/%s/sources/%s", stripe.StringValue(params.Customer), id)
 	} else if params.Account != nil {
-		err = c.B.Call("POST", stripe.FormatURLPath("/accounts/%s/external_accounts/%s", stripe.StringValue(params.Account), id), c.Key, body, commonParams, ba)
+		path = stripe.FormatURLPath("/accounts/%s/external_accounts/%s", stripe.StringValue(params.Account), id)
 	} else {
-		err = errors.New("Invalid bank account params: either Customer or Account need to be set")
+		return nil, errors.New("Invalid bank account params: either Customer or Account need to be set")
```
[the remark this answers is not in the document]
brandur-stripe: There's always some possibility of slight bugs on the implementation side being introduced by this, but it's fairly well known in Go that you should check the error result before trying to use another return.

bankaccount/client.go
```diff
@@ -130,27 +104,29 @@ func List(params *stripe.BankAccountListParams) *Iter {
...
-
-		if params.Customer != nil {
-			err = c.B.Call("GET", stripe.FormatURLPath("/customers/%s/sources?object=bank_account", stripe.StringValue(params.Customer)), c.Key, b, p, list)
-		} else if params.Account != nil {
-			err = c.B.Call("GET", stripe.FormatURLPath("/accounts/%s/external_accounts?object=bank_account", stripe.StringValue(params.Account)), c.Key, b, p, list)
-		} else {
-			err = errors.New("Invalid bank account params: either Customer or Account need to be set")
+
+		if outerErr != nil {
+			return nil, list.ListMeta, outerErr
```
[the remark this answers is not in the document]
brandur-stripe: Looks like you answered your question below, but yeah it could probably stand to have a slightly better name.

card/client.go
```diff
@@ -95,22 +92,22 @@ func (c Client) Update(id string, params *stripe.CardParams) (*stripe.Card, erro
...
-		err = c.B.Call("POST", stripe.FormatURLPath("/recipients/%s/cards/%s", stripe.StringValue(params.Recipient), id), c.Key, body, &params.Params, card)
+		path = stripe.FormatURLPath("/recipients/%s/cards/%s",
+			stripe.StringValue(params.Recipient), id)
 	} else {
-		err = errors.New("Invalid card params: either account, customer or recipient need to be set")
+		return nil, errors.New("Invalid card params: either account, customer or recipient need to be set")
 	}
 
+	card := &stripe.Card{}
+	err := c.B.Call("POST", path, c.Key, params, card)
```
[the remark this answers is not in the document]
brandur-stripe: Yeah — it's probably best to [refer to the implementation before my patch to fully convince yourself](https://github.com/stripe/stripe-go/blob/master/card/client.go), but basically I just took whatever used to be there and carried it over. It was only in the `New` functions of bank account and card where the custom append functions were being used.

order/client.go
```diff
@@ -70,23 +51,13 @@ func Pay(id string, params *stripe.OrderPayParams) (*stripe.Order, error) {
...
-	if params != nil {
-		if params.Source == nil && params.Customer == nil {
-			err := errors.New("Invalid order pay params: either customer or a source must be set")
-			return nil, err
-		}
-
-		body = &form.Values{}
-		commonParams = &params.Params
-		form.AppendTo(body, params)
+	if params != nil && params.Source == nil && params.Customer == nil {
```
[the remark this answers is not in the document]
brandur-stripe: Yeah, I kept it because it's already there, but I'm totally peaceful with letting this be an API failure. Removed.

paymentsource/client.go
```diff
@@ -43,24 +38,13 @@ func Get(id string, params *stripe.CustomerSourceParams) (*stripe.PaymentSource,
...
 
 func (s Client) Get(id string, params *stripe.CustomerSourceParams) (*stripe.PaymentSource, error) {
-	var body *form.Values
-	var commonParams *stripe.Params
-
-	if params != nil {
-		commonParams = &params.Params
-		body = &form.Values{}
-		form.AppendTo(body, params)
+	if params.Customer == nil {
```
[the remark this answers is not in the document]
brandur-stripe: Yeah. To be fair though, I just copied over what was already there in all cases, but yes, we should be consistent about this.

I've added checks on params being `nil` to `paymentsource/` as well now.

brandur-stripe:
> My only "worry" is how do we make sure that any method with a custom AppendToXXXX knows to call the CallRaw approach?

So I think it'll be relatively obvious in most cases because the new `Call` won't take a `form.Values` (only a `ParamContainer`). So you could instantiate one, do your custom append, but it'd be pretty obvious that you weren't doing anything with. Also, these are currently only needed in a few corner cases, so I'm very much hoping that we never again introduce a new API call that follows this sort of model.

But anyway, thanks as usual for the fast review! I've addressed all feedback. Let me know if you happen to see anything else, but otherwise I'll bring this in tomorrow.

brandur-stripe:
No worries Remi and thank you. Pulling this in.

brandur-stripe:
Released as 34.2.0.

Issue #580, opened 2018-06-08
account.GetByID: LegalEntity not set


nd2s:
In the latest version (`32.4.0` / a90cccc8afb97cfae9becca752adb4bed955ba6d) Account.LegalEntity is not set when getting an account.

    acc, err := account.GetByID(accountID, nil)
    if err != nil {
    	panic(err)
    }
    fmt.Println(acc.LegalEntity)

=> `(*stripe.LegalEntity)(nil)`

It's there when fetching with `curl`.

remi-stripe:
@nd2s That code works for me. Do you have an example account id where it's not working?

nd2s:
@remi-stripe Full example without key:
```go
package main

import (
	"fmt"
	"github.com/stripe/stripe-go"
	"github.com/stripe/stripe-go/account"
)

func main() {
	stripe.Key = "sk_test_xxx"

	accountID := "acct_1CKmBeBlVKHXDelz"
	acc, err := account.GetByID(accountID, nil)
	if err != nil {
		panic(err)
	}
	fmt.Println(acc.LegalEntity)
}
```

Output:
```
$ go run test.go
2018/06/08 21:49:47 Requesting GET api.stripe.com/v1/accounts/acct_1CKmBeBlVKHXDelz
<nil>
```

Is that enough info?

remi-stripe:
@nd2s I can't seem to reproduce on my end unfortunately. Can you modify the library's `stripe.go` file [here](https://github.com/stripe/stripe-go/blob/master/stripe.go#L345) to add an extra line just before this that logs the raw response?

```
fmt.Printf("raw response body = %+v\n", string(resBody))
```

nd2s:
@remi-stripe 

```
2018/06/09 10:48:35 Requesting GET api.stripe.com/v1/accounts/acct_1CKmBeBlVKHXDelz
raw response body = {
  "id": "acct_1CKmBeBlVKHXDelz",
  "object": "account",
  "business_name": "TestComp",
  "business_url": null,
  "charges_enabled": true,
  "country": "AT",
  "created": 1524657206,
  "debit_negative_balances": false,
  "decline_charge_on": {
    "avs_failure": false,
    "cvc_failure": false
  },
  "default_currency": "eur",
  "details_submitted": true,
  "display_name": null,
  "email": null,
  "external_accounts": {
    "object": "list",
    "data": [
      {
        "id": "ba_1CanQLBlVKHXDelzCEmQzP7H",
[...]

PR #582, opened 2018-06-09
Fix legal_entity[additional_owners][][verification] to be a struct

remi-stripe:
When retrieving a Custom account with additional owners, the legal entity was not deserialized property and was empty. The issue was that `legal_entity[additional_owners][][verification]` is not a string but is instead a hash with various properties. Because of this, the entire deserialization of `LegalEntity` failed and it was just `nil`.

This fixes https://github.com/stripe/stripe-go/issues/580

r? @brandur-stripe 
cc @stripe/api-libraries 

I tried to add a test for this but stripe-mock returns no `legal_entity` at all so that did not work.

### DIFF
--- a/account.go
+++ b/account.go
@@ -396,13 +396,13 @@ type DOB struct {
 
 // AdditionalOwner is the structure for an account owner.
 type AdditionalOwner struct {
-	Address                  *AccountAddress `json:"address"`
-	DOB                      DOB             `json:"dob"`
-	FirstName                string          `json:"first_name"`
-	LastName                 string          `json:"last_name"`
-	MaidenName               string          `json:"maiden_name"`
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Wow, I'm surprised that the result came out so degenerately ... we should remember this bug in the future because it seems reasonably likely we'll see it again ...

Thanks for the investigation!

remi-stripe:
@brandur-stripe Yeah! Fun fact, while debugging this I found another API bug not related to Go, which explained why I could not reproduce in the first place.

I'm surprised that Go did not just fail on that sub-property but that everything else was not deserialized properly. I wonder if there's a way to fix the library to always deserialize what matches even if other things fail.

PR #583, opened 2018-06-09
Add new constants, rename one and fix DueBy

remi-stripe:
While working on fixing https://github.com/stripe/stripe-go/issues/580, I started improving `TestAccountUnmarshal` to ensure we tried to deserialize a complex account object to catch any issue.

Doing that, I caught a few mistakes that I wanted to fix at the same time:
* `AccountRejectReason` is now a type even though it's only used as a parameter. I should have fixed it when you asked me to, it's cleaner.
* `verification[disabled_reason]` now has its own type: `IdentityVerificationDisabledReason` and related constants
* `verification[due_by]` should not be a pointer so I fixed it back to a normal `int64`
* `PayoutIntervalDay` has been renamed to `PayoutIntervalDaily`.

I think we should merge this one as a major version as it makes enough breaking changes.

If we want to unblock the other issue without a major version though we could merge https://github.com/stripe/stripe-go/pull/582 first.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/account_test.go
+++ b/account_test.go
@@ -47,7 +47,104 @@ func TestAccountUnmarshal(t *testing.T) {
 				},
 			},
 		},
+		"legal_entity": map[string]interface{}{
+			"additional_owners": []map[string]interface{}{
+				{
+					"address": map[string]interface{}{
+						"city":        "city1",
+						"country":     "US",
+						"line1":       "line1",
+						"line2":       "line2",
+						"postal_code": "90210",
+						"state":       "CA",
+					},
+					"dob": map[string]interface{}{
+						"day":   1,
+						"month": 1,
+						"year":  1970,
+					},
+					"first_name":                  "First",
+					"last_name":                   "Last",
+					"maiden_name":                 "Maiden",
+					"personal_id_number_provided": true,
+					"verification": map[string]interface{}{
+						"details":       "details",
+						"details_code":  "failed_other",
+						"document":      "file_123",
+						"document_back": "file_234",
+						"status":        "pending",
+					},
+				},
+				{
+					"address": map[string]interface{}{
+						"city": "city2",
+					},
+					"dob": map[string]interface{}{
+						"day":   1,
+						"month": 1,
+						"year":  1960,
+					},
+					"first_name": "First2",
+				},
+			},
+			"address": map[string]interface{}{
+				"city":        "city-3",
+				"country":     "US",
+				"line1":       "line1-3",
+				"line2":       "line2-3",
+				"postal_code": "91111",
+				"state":       "CA",
+			},
+			"business_tax_id_provided": true,
+			"dob": map[string]interface{}{
+				"day":   1,
+				"month": 1,
+				"year":  1980,
+			},
+			"first_name":  "John",
+			"last_name":   "Doe",
+			"maiden_name": "Maiden",
+			"personal_address": map[string]interface{}{
+				"city":        "personal_city1",
+				"country":     "US",
+				"line1":       "personal_line1",
+				"line2":       "personal_line2",
+				"postal_code": "90210",
+				"state":       "CA",
+			},
+			"personal_id_number_provided": true,
+			"ssn_last_4_provided":         true,
+			"type":                        "company",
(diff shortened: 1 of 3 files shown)


account_test.go
```diff
@@ -47,7 +47,104 @@ func TestAccountUnmarshal(t *testing.T) {
...
+		"tos_acceptance": map[string]interface{}{
+			"date":       1528573382,
+			"ip":         "127.0.0.1",
+			"user_agent": "user agent",
+		},
 		"type": "custom",
+		"verification": map[string]interface{}{
+			"disabled_reason": "fields_needed",
+			"due_by":          1528573382,
+			"fields_needed": []interface{}{
```
remi-stripe: You have no idea how long I spent on figuring out the write combination of `[]` `string` `map` `interface` and `{}` to make that one work 😓

account_test.go
```diff
@@ -57,13 +154,97 @@ func TestAccountUnmarshal(t *testing.T) {
...
+	assert.Equal(t, "legal_entity.business_name", account.Verification.FieldsNeeded[1])
+
+	// Assert ExternalAccounts are fully deserialized
 	assert.Equal(t, true, account.ExternalAccounts.HasMore)
 
 	assert.Equal(t, 2, len(account.ExternalAccounts.Data))
 	assert.Equal(t, "ba_123", account.ExternalAccounts.Data[0].ID)
 	assert.Equal(t, "card_123", account.ExternalAccounts.Data[1].ID)
+
+	// Ensure LegalEntity is fully deserialized
```
remi-stripe: I made thorough tests of almost every property, not sure if too much or not, please let me know if I should cut those in half.

nd2s:
@remi-stripe JFYI: There are more struct fields of type `*string` that should be a custom type instead. Like `PayoutScheduleParams.Interval` or `.WeeklyAnchor`.

Would be nice if these changes would find their way into the library soon.

remi-stripe:
> JFYI: There are more struct fields of type *string that should be a custom type instead. Like PayoutScheduleParams.Interval or .WeeklyAnchor.

@nd2s That is expected. You can see that all parameters are `*string` and not `*CustomType`. The reason is that if we have used the custom types, we would have needed two custom functions for every single type which would have been unmanageable so we stuck with `*string` for all of those.

Instead your code should be:

```
    account, err := account.New(&stripe.AccountParams{
      Type:      stripe.String(string(stripe.AccountTypeCustom)),
      Country: stripe.String("US"),
      PayoutSchedule: &stripe.PayoutScheduleParams {
        Interval: stripe.String(string(stripe.PayoutIntervalManual)),
      },
    })
```

account_test.go
brandur-stripe: Oh god, haha.

account_test.go
brandur-stripe: Probably not necessary, but I think it's okay now that you've already written it.

remi-stripe: Out of curiosity what do you think would have been a good test? I think this large structure is fairly easy to mess up as there are so many edge-cases. You could end up deserializing the wrong property in the wrong place like maiden name as first_name or vice versa, or identification documents like I did.

brandur-stripe:
> If we want to unblock the other issue without a major version though we could merge #582 first.

+1.

This LGTM.

account_test.go
brandur-stripe: Probably one check per "level" (i.e., level of structures which you descend to) is probably a good idea. Agree that it's easy to misname a JSON field, but I think to some degree you need to rely on the fact that these are likely to trend towards correctness over time. Adding tests gives you a little more guarantee, but the tests themselves are likely to be butchered during any sizable refactor, and so not end up testing for regressions as well as you might hope anyway.

remi-stripe: Gotcha! I see where you're coming from. I think I had similar discussions with @ob-stripe on the stripe-java refactor so it is starting to make sense to me.

brandur-stripe: Yeah, it's definitely a little subjective :$

brandur-stripe:
Released as 33.0.0.

PR #585, opened 2018-06-12
Remove File in favor of FileUpload

brandur:
As we discussed in #578, here we unify the `File` and `FileUpload`
models into a single type. We do this by removing `File`, and changing
all instances where it was referenced to `FileUpload`.

These two structs were almost identical already, with the exception of
the `MIMEType` field on `File`, which is called `Type` on `FileUpload`.
Aside from that, the rename should cause very minimal breakage because
most users would have been accessing members without referencing the
struct's name directly like `dispute.Evidence.CustomerSignature.URL`.

Also, just examining the server-side resources, I'm pretty sure that
`MIMEType` didn't even work because it looks like the field is actually
called `Type` for both files and file uploads. So even if a user has
referenced the property, they should fix their use anyway.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/dispute.go
+++ b/dispute.go
@@ -118,33 +118,33 @@ type EvidenceDetails struct {
 // Almost all fields are strings since there structures (i.e. address)
 // do not typically get parsed by anyone and are thus presented as-received.
 type DisputeEvidence struct {
-	AccessActivityLog            string `json:"access_activity_log"`
-	BillingAddress               string `json:"billing_address"`
-	CancellationPolicy           *File  `json:"cancellation_policy"`
-	CancellationPolicyDisclosure string `json:"cancellation_policy_disclosure"`
-	CancellationRebuttal         string `json:"cancellation_rebuttal"`
-	CustomerCommunication        *File  `json:"customer_communication"`
-	CustomerEmailAddress         string `json:"customer_email_address"`
-	CustomerName                 string `json:"customer_name"`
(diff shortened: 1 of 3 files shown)


remi-stripe:
@brandur-stripe Before I dig, do you think we should punt and wait for the new File resource instead that @ob-stripe is working on instead? Since it's shipping soon-ish and would be a good way to unify everything only once.

brandur-stripe:
@remi-stripe Yeah, good point. I was just re-reading Olivier's documentation on the subject, and while most of it will stay the same, the intent is for `file` to replace `file_upload` as the resource name, so I guess long term `File` will be the better name for this struct.
 
This isn't a critical refactor or anything, so I'm fine with waiting on that project to finish.

brandur-stripe:
(Although on the other hand, this is once again a fairly minor breaking change, so it might be plausible just do it now, and then fix the naming again later once that happens. I think both directions are plausible.)

remi-stripe:
Sounds good. I think we're getting closer so it might be worth putting this one as a future and shipping it in a few weeks.

@ob-stripe if you think it will take longer, let us know and we could ship this for now.

remi-stripe:
@brandur-stripe  Okay that sounds good after all. Let me review the PR!

brandur-stripe:
Excellent. Thanks Remi! I'm happy to write up the next patch as Olivier's work continues to progress too.

brandur-stripe:
Released as 34.0.0.

PR #586, opened 2018-06-13
Add support for new properties on the Invoice object

remi-stripe:
Add support for new properties on the Invoice object:
- `amount_paid`, `amount_remaining` and `subscription_proration_date`
- add support for `billing_reason` with the new type `InvoiceBillingReason` and the potential values for that property as documented [here](https://stripe.com/docs/api#invoice_object-billing_reason)
- remove comments that listed specific enum values.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -3,7 +3,6 @@ package stripe
 import "encoding/json"
 
 // InvoiceLineType is the list of allowed values for the invoice line's type.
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Awesome. Thanks Remi!

PR #587, opened 2018-06-14
Use `net/http` constants instead of string literals for HTTP verbs

brandur:
Switches from string literals like `"GET"` over to their corresponding
constants in `net/http` like `http.MethodGet`. This isn't a huge win,
but seems like slightly better hygiene and slightly reduces the risk of
typos because a mistyped verb will now not compile instead of crashing
on a test or at runtime.

For review: I basically did a project-wide find and replace, so me
accidentally screwing up a verb change is relatively unlikely.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/account/client.go
+++ b/account/client.go
@@ -1,6 +1,8 @@
 package account
 
 import (
+	"net/http"
+
 	stripe "github.com/stripe/stripe-go"
 	"github.com/stripe/stripe-go/form"
 )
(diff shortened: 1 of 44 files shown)


account/client.go
```diff
@@ -1,6 +1,8 @@
 package account
 
 import (
+	"net/http"
```
brandur-stripe: Just to possibly pre-empt some feedback: `net/http` is new in all client packages, but they were already depending on it transitively through the top-level `stripe` package, so I don't think including it in all of these will have any meaningful effect on final binary size.

stripe.go
```diff
@@ -240,7 +240,7 @@ func (s *BackendConfiguration) CallRaw(method, path, key string, form *form.Valu
 	var body io.Reader
 	if form != nil && !form.Empty() {
 		data := form.Encode()
-		if strings.ToUpper(method) == "GET" {
+		if method == http.MethodGet {
```
brandur-stripe: I stripped the `ToUpper` here — all calls now use `net/http` constants so it should be perfectly safe to just compare it directly against `http.MethodGet` now.

brandur-stripe:
Thank you as usual Remi!

brandur-stripe:
Released as 34.3.0.

PR #591, opened 2018-06-15
Organize core files alphabetically and into sections

brandur:
This patch organizes some of library's core files (`stripe.go`,
`params.go`, `iter.go`) alphabetically and into sections. Here are the
key changes:

* Introduces "sections" of codes, these are (in order how they actually
  appear in the file too):

    ```
    // Public constants
    // Public variables
    // Public types
    // Public functions
    // Private constants
    // Private variables
    // Private types
    // Private functions
    ```

    The premise is that symbols which are intended to be more widely
    used (because they're exported out of the package) are at the top of
    the file for easier access, while unexported symbols are organized
    closer to the bottom.

* Within each section, all symbols are organized alphabetically.
* All type-specific functions (e.g. `func (p *Params) GetParams`) are
  arranged below the type they're attached to, and organized
  alphabetically within their type. Unexported functions are arranged
  below exported functions because that's better, also because lowercase
  letters lexically sort after uppercase letters (`A` > `a`).

And a few other minor opportunistic changes that are somewhat unrelated:

* I dropped the `TotalBackends` constant because it wasn't referenced
  anywhere (this could technically be considered a breaking change, but
  because I don't think there was any sane reason to use it, I don't
  think I'll bump the major version).
* I changed `sleepTime` from a type-specific field to one that's not
  (i.e., `func sleepTime`) because it doesn't depend on any state within
  the struct it was attached to.

I kind of want to get this in because some of these files are turning
into a huge mess and it's a good time because we have no other
outstanding pull requests.

r? @remi-stripe Sorry this diff is a giant mess, but I listed everything that
changed above so it might be better/easier just to read that and let me know if
you agree with the decisions or not. I think there's even more organization 
that we can get out too, but I want to break this up a little bit.

### DIFF
--- a/iter.go
+++ b/iter.go
@@ -6,8 +6,9 @@ import (
 	"github.com/stripe/stripe-go/form"
 )
 
-// Query is the function used to get a page listing.
-type Query func(*Params, *form.Values) ([]interface{}, ListMeta, error)
+//
+// Public types
+//
 
 // Iter provides a convenient interface
 // for iterating over the elements
@@ -26,6 +27,67 @@ type Iter struct {
 	query      Query
 	values     []interface{}
 }
+
+// Current returns the most recent item
+// visited by a call to Next.
+func (it *Iter) Current() interface{} {
+	return it.cur
+}
+
+// Err returns the error, if any,
+// that caused the Iter to stop.
+// It must be inspected
+// after Next returns false.
+func (it *Iter) Err() error {
+	return it.err
+}
+
+// Meta returns the list metadata.
+func (it *Iter) Meta() *ListMeta {
+	return &it.meta
+}
+
+// Next advances the Iter to the next item in the list,
+// which will then be available
+// through the Current method.
+// It returns false when the iterator stops
+// at the end of the list.
+func (it *Iter) Next() bool {
+	if len(it.values) == 0 && it.meta.HasMore && !it.listParams.Single {
+		// determine if we're moving forward or backwards in paging
+		if it.listParams.EndingBefore != nil {
+			it.listParams.EndingBefore = String(listItemID(it.cur))
+			it.formValues.Set(EndingBefore, *it.listParams.EndingBefore)
+		} else {
+			it.listParams.StartingAfter = String(listItemID(it.cur))
+			it.formValues.Set(StartingAfter, *it.listParams.StartingAfter)
+		}
+		it.getPage()
+	}
+	if len(it.values) == 0 {
+		return false
+	}
+	it.cur = it.values[0]
+	it.values = it.values[1:]
+	return true
+}
+
+func (it *Iter) getPage() {
+	it.values, it.meta, it.err = it.query(it.listParams.GetParams(), it.formValues)
+
+	if it.listParams.EndingBefore != nil {
+		// We are moving backward,
+		// but items arrive in forward order.
+		reverse(it.values)
+	}
+}
+
+// Query is the function used to get a page listing.
+type Query func(*Params, *form.Values) ([]interface{}, ListMeta, error)
+
+//
+// Public functions
+//
 
 // GetIter returns a new Iter for a given query and its options.
 func GetIter(container ListParamsContainer, query Query) *Iter {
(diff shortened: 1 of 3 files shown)


iter.go
```diff
@@ -27,43 +28,23 @@ type Iter struct {
...
-		formValues: formValues,
-		listParams: *listParams,
-		query:      query,
-	}
-
-	iter.getPage()
-
-	return iter
+// Current returns the most recent item
+// visited by a call to Next.
```
remi-stripe: before I review everything, is this going to the next line too early? Is it because go expects your comment length to be aligned to the function's prototype or something?

brandur-stripe: Yep, definitely —

I left all code + comments exactly the same for now though (they've just been moved around). I want to fix this, but I think I'm going to push it to another PR where it'll be more clear what's changing and a lot easier to read.

remi-stripe: ah gotcha. I've always been confused at our comment lines length so felt like the right time to ask. Okay going to review the organization for real now.

params.go
```diff
@@ -116,97 +107,82 @@ type ListParams struct {
...
-
-	// LesserThan specifies that values should be lesser than this timetamp.
-	LesserThan int64 `form:"lt"`
-
-	// LesserThanOrEqual specifies that values should be lesser than or
-	// equalthis timetamp.
-	LesserThanOrEqual int64 `form:"lte"`
+// GetListParams returns a ListParams struct (itself). It exists because any
+// structs that embed ListParams will inherit it, and thus implement the
+// ListParamsContainer interface.
```
remi-stripe: Do you consider this type public? 
I guess I have some confusion around what is "public for other packages to use" and what is "public for users to use". And this seems to be the latter right?

stripe.go
```diff
@@ -485,6 +370,107 @@ func (s *BackendConfiguration) ResponseToError(res *http.Response, resBody []byt
...
 
+// SetMaxNetworkRetries sets max number of retries on failed requests
+func (s *BackendConfiguration) SetMaxNetworkRetries(maxNetworkRetries int) {
+	s.MaxNetworkRetries = maxNetworkRetries
+}
+
+// Checks if an error is a problem that we should retry on. This includes both
+// socket errors that may represent an intermittent problem and some special
+// HTTP statuses.
+func (s *BackendConfiguration) shouldRetry(err error, resp *http.Response, numRetries int) bool {
```
remi-stripe: Minor but should it be `ShouldRetry`? (sorry I keep mixing those reviews with trying to learn from you, lmk if I should try to separate both)

stripe.go
```diff
@@ -485,6 +370,107 @@ func (s *BackendConfiguration) ResponseToError(res *http.Response, resBody []byt
...
+		backends.mu.Lock()
+		defer backends.mu.Unlock()
+		backends.Uploads = &BackendConfiguration{Type: backend, URL: uploadsURL, HTTPClient: httpClient}
+		return backends.Uploads
+	}
+
+	return nil
+}
+
 // FormatURLPath takes a format string (of the kind used in the fmt package)
```
remi-stripe: minor but it should be before `GetBackend`

stripe.go
```diff
@@ -539,6 +550,91 @@ func SetAppInfo(info *AppInfo) {
...
+}
+
+// String returns a pointer to the string value passed in.
+func String(v string) *string {
+	return &v
+}
+
+// StringValue returns the value of the string pointer passed in or
+// "" if the pointer is nil.
+func StringValue(v *string) string {
```
remi-stripe: I wonder if it hurts discovery that this is not grouped with the other ones. But I do see the value of alphabetical ordering.

params.go
brandur-stripe: (I'll cover this in a comment below that responds to your top-level points.)

stripe.go
brandur-stripe: I think we want to keep it as `shouldRetry` so that it's not exported beyond the `stripe` package (it's just used a helper function by other functions on `BackendConfiguration`).

And just for clarity in case you haven't seen it before: whether a symbol is exported or not in Go depends on the case of its first letter. An uppercase letter means it's exported, and a lowercase letter mean it's not.

stripe.go
brandur-stripe: Great catch! Fixed.

stripe.go
brandur-stripe: Yeah, I thought about that as well :/

I think the best argument for keeping it sorted like this is that if you ever use any kind of Go-related tooling (say view the package in Godoc or use vim-go to do autocompletion), they'll all end up sorted like we see here because those tools will make no distinction about where a symbol is located in the source file.

IMO our best bet will be to keep it as it is here, but then introduce an explicit section of documentation in the source/Godoc that talks just about these helpers and lists each of them explicitly.

stripe.go
remi-stripe: oh I did not know that. Shouldn't it be in the private function list at the bottom in that case though?

brandur-stripe:
Excellent. Thanks Remi!

Anyway, let's keep the ball rolling here and try to apply smaller cleanups where we spot them. I'd say that all in all between `form`, stripe-mock, working tests, alphabetization, `stripe.*` helpers, sane field names, etc. the library is 10x better today than two years ago, but still plenty to do :)

PR #593, opened 2018-06-18
Activate linting for a few packages

brandur:
This patch brings golint [1] into the Travis build and adds a make
target that will start running it on a few packages that we've expunged
of all linting errors.

I've just added a few packages to this initially because there are
hundreds of linting errors across the entire system, and cleaning them
all up would be an incredibly noisy diff. My recommendation is that we
issue smaller patches with ~1-6 packages converted over at a time to
make reviews easier (and more reliable), and add them to the whitelist
in `Makefile` as we go. After all packages are converted, we change the
call to `golint ./...`, and it'll always just run for everything.

Golint's major advancement will be to help us vet that Godoc comments
are formatted correctly, e.g. they all start with a function name like
`Get retrieves a charge`, and that exported functions _have_ Godoc
comments. This will hopefully get us to a place where we can generate
and publish a complete set of clean Godocs.

[1] https://github.com/golang/lint

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/account/client.go
+++ b/account/client.go
@@ -1,3 +1,6 @@
+// Package account provides API functions related to accounts.
+//
+// For more details, see: https://stripe.com/docs/api/go#accounts.
 package account
 
 import (
@@ -7,7 +10,7 @@ import (
 	"github.com/stripe/stripe-go/form"
 )
 
-// Client is used to invoke /account APIs.
+// Client is used to invoke APIs related to accounts.
 type Client struct {
 	B   stripe.Backend
 	Key string
@@ -18,28 +21,31 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().New(params)
 }
 
+// New creates a new account.
 func (c Client) New(params *stripe.AccountParams) (*stripe.Account, error) {
 	acct := &stripe.Account{}
 	err := c.B.Call(http.MethodPost, "/accounts", c.Key, params, acct)
 	return acct, err
 }
 
-// Get returns the details of an account.
+// Get retrieves the authenticating account.
 func Get() (*stripe.Account, error) {
 	return getC().Get()
 }
 
+// Get retrieves the authenticating account.
 func (c Client) Get() (*stripe.Account, error) {
 	account := &stripe.Account{}
 	err := c.B.Call(http.MethodGet, "/account", c.Key, nil, account)
 	return account, err
 }
 
-// GetByID returns the details of your account.
+// GetByID retrieves an account.
 func GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().GetByID(id, params)
 }
 
+// GetByID retrieves an account.
 func (c Client) GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	account := &stripe.Account{}
@@ -47,11 +53,12 @@ func (c Client) GetByID(id string, params *stripe.AccountParams) (*stripe.Accoun
 	return account, err
 }
 
-// Update updates the details of an account.
+// Update updates an account.
 func Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().Update(id, params)
 }
 
+// Update updates an account.
 func (c Client) Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	acct := &stripe.Account{}
@@ -59,11 +66,12 @@ func (c Client) Update(id string, params *stripe.AccountParams) (*stripe.Account
 	return acct, err
 }
 
-// Del deletes an account
+// Del deletes an account.
 func Del(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().Del(id, params)
 }
 
+// Del deletes an account.
 func (c Client) Del(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	acct := &stripe.Account{}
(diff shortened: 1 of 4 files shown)


account/client.go
```diff
@@ -7,7 +10,7 @@ import (
 	"github.com/stripe/stripe-go/form"
 )
 
-// Client is used to invoke /account APIs.
+// Client is used to invoke APIs related to accounts.
```
remi-stripe: is the change before /account is not a valid comment?

account/client.go
```diff
@@ -18,76 +21,83 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().New(params)
 }
 
+// New creates a new account.
```
remi-stripe: I understand adding comments to a library but I'm not a fan of really short/terse comments like this. I don't know if it's a Go thing, but I wanted to flag just once.
Also, does it do like javadoc where you comment each parameter and returned values and errors?

account/client.go
```diff
@@ -18,76 +21,83 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
...
+// GetByID retrieves an account.
 func (c Client) GetByID(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	path := stripe.FormatURLPath("/accounts/%s", id)
 	account := &stripe.Account{}
 	err := c.B.Call(http.MethodGet, path, c.Key, params, account)
 	return account, err
 }
 
-// Update updates the details of an account.
+// Update updates of an account.
```
remi-stripe: Should we remove the `of`?

account/client.go
```diff
@@ -18,76 +21,83 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
...
 func Update(id string, params *stripe.AccountParams) (*stripe.Account, error) {
 	return getC().Update(id, params)
 }
 
+// Update updates of an account.
```
remi-stripe: same as above

account/client.go
```diff
@@ -18,76 +21,83 @@ func New(params *stripe.AccountParams) (*stripe.Account, error) {
...
 	return acct, err
 }
 
-// Get returns the details of an account.
+// Get retrieves the authenticating account.
 func Get() (*stripe.Account, error) {
 	return getC().Get()
 }
 
+// Get retrieves the authenticating account.
```
remi-stripe: Can anything about the duplicate comments between both calls? Also aren't you missing one `New` since there should be two of those in the file?

account/client.go
brandur-stripe: Yeah, I made this mostly because I feel a little like the particular URLs being accessed by the package are an implementation detail — not necessarily something that needs to be included in the documentation. The docs are still overly terse, but I was intending to basically replace a URL with a resource name.

account/client.go
brandur-stripe: Yeah, I kind of agree, but it's a bit of a compromise — we could go with something much more long-winded, but it has a higher chance of falling out of date more quickly. I went with this instead, and admittedly it's not great documentation, but it will have the advantage of at least forcing documentation for all exported types/functions later on if we're able to successfully activate Golint.

I'm actually up for trying an alternative if you have a good suggestion. One approach is that we could try to copy what's in the reference documentation, but again, I feel like it might be in slight danger of becoming outdated:

> To charge a credit card or other payment source, you create a Charge object. If your API key is in test mode, the supplied payment source (e.g., card) won't actually be charged, although everything else will occur as if in live mode. (Stripe assumes that the charge would have completed successfully).

Regarding commenting on each parameter and returned value: no this isn't really a thing in Go. Instead, important parameters and return values are expected to be mentioned in the normal documentation string. You can find some examples of this in the Godoc for the standard library:

https://godoc.org/net/http#Redirect

account/client.go
brandur-stripe: Oops, yep. Doing that now.

account/client.go
brandur-stripe: Unfortunately no ... Golint expects all exported types/functions to be documented. One thing we could do is to possibly just reference the first invocation from the second so that at least we don't have to duplicate the text (might be less change churn when updates are made in the future):

``` go
// New is an alias for Client.New that uses a default client.
```

Maybe that would better?

> Also aren't you missing one New since there should be two of those in the file?

Believe it or not, one of them already had this docstring on it. You can see that the final result looks right if you open the branch's file.

brandur-stripe:
Great questions on the review Remi. I left some answers if you could take a look whenever you get a chance.

ptal @remi-stripe

account/client.go
remi-stripe: > I'm actually up for trying an alternative if you have a good suggestion. One approach is that we could try to copy what's in the reference documentation, but again, I feel like it might be in slight danger of becoming outdated

I fully agree. That's why I mostly removed all of those when I did the merge. I feel like the methods are self explanatory and devs should look at our official API ref instead though I don't know how Go devs work. For me it's the same as the ongoing PRs to tag things in stripe-php.

But I think your explanation makes sense and I'm happy to move forward with this.

account/client.go
remi-stripe: gotcha. In that case I think it's better to document than reference the other one because it would display the right string to the user who needs it.

remi-stripe:
@brandur-stripe Thanks for the details answers! LGTM

account/client.go
brandur-stripe: Thanks that sounds good. I think if we eventually go to auto-generation, all of this could possibly be less big of a deal because we'll be able to easily change the source and distribute updates out to everywhere (after updating all those sources to be a little more language-agnostic of course).

One more possibility: there were direct links before in a couple of these to the associated action in the API reference documentation like:

https://stripe.com/docs/api/go#create_charge

I could put those back in, and at least that might the docstrings a little more helpful. What do you think?

remi-stripe: I would punt. More than once we ended up "guessing" the URLs while adding something gated that did not have docs already and it's been not maintained in general.

brandur-stripe: True — that has definitely been a problem in the past. Okay, thanks! I'm going to merge this and we can see how the rest of it goes.

Issue #497, opened 2017-12-05
function creating Stripe-Signature


jmhodges:
It would be nice if this library had a means to create a Stripe-Signature for use in tests of code that use `webhook.ConstructEvent` and related functions. I think I'll have one to contribute soon if that's of interest.

ob-stripe:
Hey @jmhodges, good news! The library already has this feature via the `computeSignature` function: https://github.com/stripe/stripe-go/blob/0f6277583a1514c96f7141d8aa1365eea6b382f8/webhook/client.go#L32-L38.

You can take a look at the code [here](https://github.com/stripe/stripe-go/blob/0f6277583a1514c96f7141d8aa1365eea6b382f8/webhook/client_test.go#L25-L41) to see how we generate signed payloads in our tests.

I'm closing the issue as this feature already exists, but feel free to reply if I misunderstood your request!

brandur-stripe:
The function signature should be pretty stable, so it might be okay to just export this. One thing we might want to clean up is the types: it's a little weird that `payload` comes in as `[]byte` and `secret` comes in as `string`. Ideally, they'd both be one or the other.

`secret` gets changed to a `[]byte` inside the function anyway, so should be an easy change at least.

brandur-stripe:
Exported in 40.0.0.

PR #602, opened 2018-06-24
webhook: export ComputeSignature

enzadesai:
Fixes https://github.com/stripe/stripe-go/issues/497.

### DIFF
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	// Signatures older than this will be rejected by ConstructEvent
+	// DefaultTolerance signatures older than this will be rejected by ConstructEvent
 	DefaultTolerance time.Duration = 300 * time.Second
 	signingVersion   string        = "v1"
 )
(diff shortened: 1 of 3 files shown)


webhook/client.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	// Signatures older than this will be rejected by ConstructEvent
+	// DefaultTolerance signatures older than this will be rejected by ConstructEvent
```
brandur-stripe: Thanks for the fix here, but could you make this a sensible sentence? Like `DefaultTolerance is the amount of drift in time that's allowed by ConstructEven as it's validating a signature.`

webhook/client.go
```diff
@@ -27,10 +27,10 @@ var (
 	ErrNoValidSignature error = errors.New("Webhook had no valid signature")
 )
 
-// Computes a webhook signature using Stripe's v1 signing method. See
+// ComputeSignature computes a webhook signature using Stripe's v1 signing method. See
 // https://stripe.com/docs/webhooks#signatures
-func computeSignature(t time.Time, payload []byte, secret string) []byte {
-	mac := hmac.New(sha256.New, []byte(secret))
+func ComputeSignature(t time.Time, payload, secret []byte) []byte {
```
brandur-stripe: Thanks a lot for trying to follow the advice I was giving in #497!

Contrary to what I said there though, I now see that unfortunately changing the signature to `[]byte` ends up breaking compatibility on a lot of these already-exported functions. It's also still not perfectly consistent because `header` in some of the functions below is still a string.

Sorry for the reversal, but for those reasons, I think it actually might be better to leave `secret` here as a `string`, at least for the time being. Would you mind changing it back?

akamensky:
Would this still be merged? I am trying to use hooks and surprise surprise -- I need to manually write code for signature verification (and what if the algo later will change?).

brandur-stripe:
This pull request seems to have stalled out, so I'm going to close it in favor of #650.

PR #603, opened 2018-06-24
Add support for Level III data on charge creation

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 
Related to https://github.com/stripe/stripe-dotnet/pull/1216 which is the same for stripe-dotnet.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -19,6 +19,24 @@ const (
 	ChargeFraudStripeReportFraudulent ChargeFraudStripeReport = "fraudulent"
 )
 
+// ChargeLevel3LineItemsParams is the set of parameters that represent a line item on level III data.
+type ChargeLevel3LineItemsParams struct {
+	DiscountAmount     *int64  `form:"discount_amount"`
+	ProductCode        *string `form:"product_code"`
+	ProductDescription *string `form:"product_description"`
+	Quantity           *int64  `form:"quantity"`
+	TaxAmount          *int64  `form:"tax_amount"`
+	UnitCost           *int64  `form:"unit_cost"`
+}
+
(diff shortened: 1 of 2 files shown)


remi-stripe:
changed to [WIP] as this requires changes to the Charge ressource too but those are behind a gate

remi-stripe:
Okay pushed a new version that supports the `level3` hash on the resource too. Since it's gated, I could not add real tests as stripe-mock won't return it.

charge.go
```diff
@@ -19,6 +19,24 @@ const (
...
+	TaxAmount          *int64  `form:"tax_amount"`
+	UnitCost           *int64  `form:"unit_cost"`
+}
+
+// ChargeLevel3Params is the set of parameters that can be used for the Level III data.
+type ChargeLevel3Params struct {
+	LineItems          []*ChargeLevel3LineItemsParams `form:"line_items"`
+	MerchantReference  *string                        `form:"merchant_reference"`
+	ShippingAddressZip *string                        `form:"shipping_address_zip"`
+	ShippingAmount     *int64                         `form:"shipping_amount"`
```
brandur-stripe: I was just looking at the implementation and noticed that there also seems to be parameters for `customer_reference` and `shipping_from_zip` in there. Do we need those?

remi-stripe: @brandur-stripe hmmm potentially though those are not documented even in our internal documentation so I think we likely should punt and add those later once the feature is out of beta or asked in one of the languages. What do you think?

brandur-stripe: Yeah, sounds good to me. Thanks.

brandur-stripe:
Released as 35.6.0.

PR #604, opened 2018-06-26
Add extra params/properties for Level III on Charge

remi-stripe:
@brandur-stripe turns out you were right and those are supported, and the docs are getting updated.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM. Thanks for tracking that down.

PR #605, opened 2018-06-26
Use `travis_retry` helper when using `go get` for dependencies

brandur:
Our resolved dependencies aren't cached and we occasionally see
intermittent failures as `go get` runs and temporarily fails to get one
of our dependencies. This fails the build, and someone has to go in and
restart it.

It turns out that Travis provides a handy `travis_retry` helper [1]
which will retry a command up to three times if it returns a non-zero
exit code. I've added it in for all dependency resolution, and we'll see
if it helps.

[1] https://docs.travis-ci.com/user/common-build-problems/#travis_retry

r? @remi-stripe

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,15 @@
 before_install:
+  # Install various build dependencies. We use `travis_retry` because `go get`
+  # will occasionally fail intermittently.
+
   # The testify require framework is used for assertions in the test suite
-  - go get -u github.com/stretchr/testify/require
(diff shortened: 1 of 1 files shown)


brandur-stripe:
@remi-stripe Okay this seems to work. Mind taking r? Thanks!

remi-stripe:
Thank you for fixing so fast as usual!

brandur-stripe:
No worries! Thanks for the quick review.

PR #606, opened 2018-06-27
Add support for the PaymentIntent resource

remi-stripe:
This resource is gated but stripe-mock has been updated to support the APIs so tests run as expected.

r? @brandur-stripe 
cc @stripe/api-libraries 
cc @michelle-stripe @jenan-stripe

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -28,6 +28,7 @@ import (
 	"github.com/stripe/stripe-go/loginlink"
 	"github.com/stripe/stripe-go/order"
 	"github.com/stripe/stripe-go/orderreturn"
+	"github.com/stripe/stripe-go/paymentintent"
 	"github.com/stripe/stripe-go/paymentsource"
 	"github.com/stripe/stripe-go/payout"
 	"github.com/stripe/stripe-go/plan"
@@ -98,6 +99,8 @@ type API struct {
 	Orders *order.Client
 	// OrderReturns is the client used to invoke /order_returns APIs.
 	OrderReturns *orderreturn.Client
+	// PaymentIntents is the client used to invoke /payment_intents APIs.
+	PaymentIntents *paymentintent.Client
 	// PaymentSource is used to invoke customer sources related APIs.
 	PaymentSource *paymentsource.Client
 	// Payouts is the client used to invoke /payouts APIs.
@@ -164,6 +167,7 @@ func (a *API) Init(key string, backends *Backends) {
 	a.LoginLinks = &loginlink.Client{B: backends.API, Key: key}
 	a.Orders = &order.Client{B: backends.API, Key: key}
 	a.OrderReturns = &orderreturn.Client{B: backends.API, Key: key}
+	a.PaymentIntents = &paymentintent.Client{B: backends.API, Key: key}
 	a.PaymentSource = &paymentsource.Client{B: backends.API, Key: key}
 	a.Payouts = &payout.Client{B: backends.API, Key: key}
 	a.Plans = &plan.Client{B: backends.API, Key: key}
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -1,0 +1,175 @@
+package stripe
+
+import (
+	"encoding/json"
+)
+
+type PaymentIntentCaptureMethod string
+
+const (
+	PaymentIntentCaptureMethodAutomatic PaymentIntentCaptureMethod = "automatic"
+	PaymentIntentCaptureMethodManual    PaymentIntentCaptureMethod = "manual"
+)
+
+type PaymentIntentConfirmationMethod string
+
+const (
+	PaymentIntentConfirmationMethodPublishable PaymentIntentConfirmationMethod = "publishable"
+	PaymentIntentConfirmationMethodSecret      PaymentIntentConfirmationMethod = "secret"
+)
+
+type PaymentIntentNextActionType string
+
+const (
+	PaymentIntentNextActionAuthorizeWithURL PaymentIntentNextActionType = "authorize_with_url"
+)
+
+type PaymentIntentStatus string
+
+const (
+	PaymentIntentStatusCanceled             PaymentIntentStatus = "canceled"
+	PaymentIntentStatusProcessing           PaymentIntentStatus = "processing"
+	PaymentIntentStatusRequiresCapture      PaymentIntentStatus = "requires_capture"
+	PaymentIntentStatusRequiresConfirmation PaymentIntentStatus = "requires_confirmation"
+	PaymentIntentStatusRequiresSource       PaymentIntentStatus = "requires_source"
+	PaymentIntentStatusRequiresSourceAction PaymentIntentStatus = "requires_source_action"
+	PaymentIntentStatusSucceeded            PaymentIntentStatus = "succeeded"
+)
+
+type PaymentIntentTransferDataParams struct {
+	Amount *int64 `form:"amount"`
+}
+
+// PaymentIntentParams is the set of parameters that can be used when handling a payment intent.
+type PaymentIntentParams struct {
+	Params               `form:"*"`
+	AllowedSourceTypes   []*string                        `form:"allowed_source_types"`
+	Amount               *int64                           `form:"amount"`
+	ApplicationFee       *int64                           `form:"application_fee"`
+	AttemptConfirmation  *bool                            `form:"attempt_confirmation"`
+	CaptureMethod        *string                          `form:"capture_method"`
+	Currency             *string                          `form:"currency"`
+	Customer             *string                          `form:"customer"`
+	Description          *string                          `form:"description"`
+	OnBehalfOf           *string                          `form:"on_behalf_of"`
+	ReceiptEmail         *string                          `form:"receipt_email"`
+	ReturnURL            *string                          `form:"return_url"`
+	SaveSourceToCustomer *bool                            `form:"save_source_to_customer"`
+	Shipping             *ShippingDetailsParams           `form:"shipping"`
+	Source               *string                          `form:"source"`
+	StatementDescriptor  *string                          `form:"statement_descriptor"`
+	TransferData         *PaymentIntentTransferDataParams `form:"transfer_data"`
+	TransferGroup        *string                          `form:"transfer_group"`
+}
+
+// PaymentIntentListParams is the set of parameters that can be used when listing payment intents.
+// For more details see https://stripe.com/docs/api#list_payouts.
+type PaymentIntentListParams struct {
+	ListParams `form:"*"`
+}
+
+type PaymentIntentSourceActionAuthorizeWithURL struct {
+	URL string `json:"url"`
+}
+
+// PaymentIntentSourceActionValue describes the `value` property in `next_source_action`
+// The `type` in the parent should indicate which object is fleshed out.
+type PaymentIntentSourceActionValue struct {
+	AuthorizeWithURL *PaymentIntentSourceActionAuthorizeWithURL `json:"-"`
+}
+
+type PaymentIntentSourceAction struct {
+	Type  PaymentIntentNextActionType     `json:"type"`
+	Value *PaymentIntentSourceActionValue `json:"-"`
+}
+
+// UnmarshalJSON handles deserialization of a PaymentIntentSourceAction.
+// This custom unmarshaling is needed because the specific
+// type of for `value` it refers to is specified in the `type` property
+func (s *PaymentIntentSourceAction) UnmarshalJSON(data []byte) error {
+	type paymentIntentSourceAction PaymentIntentSourceAction
+	var v paymentIntentSourceAction
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	var err error
+	*s = PaymentIntentSourceAction(v)
+	s.Value = &PaymentIntentSourceActionValue{}
+
+	// Unmarshal data a second time so that we can get the raw bytes for the
+	// `value` field
+	var rawObject map[string]*json.RawMessage
+	if err := json.Unmarshal(data, &rawObject); err != nil {
+		return err
+	}
+
+	switch s.Type {
+	case PaymentIntentNextActionAuthorizeWithURL:
+		err = json.Unmarshal(*rawObject["value"], &s.Value.AuthorizeWithURL)
+	}
+
+	return err
+}
+
+type PaymentIntentTransferData struct {
+	Amount int64 `json:"amount"`
+}
+
+// Payout is the resource representing a Stripe payout.
+// For more details see https://stripe.com/docs/api#payouts.
+type PaymentIntent struct {
+	AllowedSourceTypes  []string                        `json:"allowed_source_types"`
+	Amount              int64                           `json:"amount"`
+	AmountCapturable    int64                           `json:"amount_capturable"`
+	AmountReceived      int64                           `json:"amount_received"`
+	Application         *Application                    `json:"application"`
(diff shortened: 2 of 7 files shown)


paymentintent.go
```diff
@@ -0,0 +1,134 @@
...
+	TransferGroup        *string                          `form:"transfer_group"`
+}
+
+// PaymentIntentListParams is the set of parameters that can be used when listing payment intents.
+// For more details see https://stripe.com/docs/api#list_payouts.
+type PaymentIntentListParams struct {
+	ListParams `form:"*"`
+}
+
+type PaymentIntentNextSourceActionValue struct {
```
remi-stripe: We will need to do something smarter here as the shape of `value` changes based on `type`. This will be closer to what we do with BalanceTransaction and `source`

paymentintent.go
```diff
@@ -0,0 +1,174 @@
...
+
+	var err error
+	*s = PaymentIntentSourceAction(v)
+
+	fmt.Printf("%v", s.Type)
+
+	switch s.Type {
+	case PaymentIntentNextActionAuthorizeWithURL:
+		fmt.Printf("%v", data)
+		err = json.Unmarshal(data, &s.Value.AuthorizeWithUrl)
```
remi-stripe: I know this part is incorrect because at that point `data` still contains the top-level info but I haven't yet figured out out I can fix it.

paymentintent.go
```diff
@@ -0,0 +1,167 @@
...
+	ApplicationFee       int64                           `json:"application_fee"`
+	CanceledAt           int64                           `json:"canceled_at"`
+	CaptureMethod        PaymentIntentCaptureMethod      `json:"capture_method"`
+	Charges              *ChargeList                     `json:"charges"`
+	ClientSecret         string                          `json:"client_secret"`
+	ConfirmationMethod   PaymentIntentConfirmationMethod `json:"confirmation_method"`
+	Created              int64                           `json:"created"`
+	Currency             string                          `json:"currency"`
+	Customer             *Customer                       `json:"customer"`
+	Livemode             bool                            `json:"livemode"`
```
brandur-stripe: Missing `Description`.

paymentintent.go
```diff
@@ -0,0 +1,167 @@
...
+	ID                   string                          `json:"id"`
+	Metadata             map[string]string               `json:"metadata"`
+	NextSourceAction     *PaymentIntentSourceAction      `json:"next_source_action"`
+	OnBehalfOf           *Account                        `json:"on_behalf_of"`
+	ReceiptEmail         string                          `json:"receipt_email"`
+	ReturnURL            string                          `json:"return_url"`
+	SaveSourceToCustomer bool                            `json:"save_source_to_customer"`
```
brandur-stripe: I'm not seeing `SaveSourceToCustomer` in the latest documentation. Can you double-check that it's necessary?

client/api.go
```diff
@@ -148,6 +149,8 @@ type API struct {
 	PaymentSource *paymentsource.Client
 	// ExchangeRates is the client used to invoke /exchange_rates APIs.
 	ExchangeRates *exchangerate.Client
+	// PaymentIntents is the client used to invoke /exchange_rates APIs.
```
brandur-stripe: Slight fix needed in the `/exchange_rates` part of the comment.

brandur-stripe: BTW, we should try to remember to alphabetize both these lists later :)

paymentintent.go
```diff
@@ -0,0 +1,167 @@
...
+	TransferGroup        *string                          `form:"transfer_group"`
+}
+
+// PaymentIntentListParams is the set of parameters that can be used when listing payment intents.
+// For more details see https://stripe.com/docs/api#list_payouts.
+type PaymentIntentListParams struct {
+	ListParams `form:"*"`
+}
+
+type PaymentIntentSourceActionAuthorizeWithUrl struct {
```
brandur-stripe: Should probably make this `URL` instead of `Url`.

paymentintent.go
```diff
@@ -0,0 +1,167 @@
...
+	URL string `json:"url"`
+}
+
+// PaymentIntentSourceActionValue describes the `value` property in `next_source_action`
+// The `type` in the parent should indicate which object is fleshed out.
+type PaymentIntentSourceActionValue struct {
+	AuthorizeWithUrl *PaymentIntentSourceActionAuthorizeWithUrl `json:"-"`
```
brandur-stripe: And similarly, with the name of this field: `AuthorizeWithURL`.

paymentintent_test.go
```diff
@@ -0,0 +1,53 @@
...
+	assert.NoError(t, err)
+
+	assert.Equal(t, "pi_123", intent.ID)
+
+	assert.Equal(t, 1, len(intent.AllowedSourceTypes))
+	assert.Equal(t, 2, len(intent.Charges.Data))
+
+	assert.Equal(t, PaymentIntentNextActionAuthorizeWithURL, intent.NextSourceAction.Type)
+	assert.Equal(t, "https://stripe.com", intent.NextSourceAction.Value.AuthorizeWithUrl.URL)
+}
```
brandur-stripe: Would you mind adding a test specifically for `PaymentIntentSourceAction`'s `UnmarshalJSON`? IMO, we should try to have a specific test for every type that has a custom `UnmarshalJSON` implementation.

It's fairly similar to `ExternalAccount`'s unmarshaling, so I'd use the test `TestExternalAccount_UnmarshalJSON` in `account_test.go` for inspiration.

We could also then remove the `next_source_action` from `TestPaymentIntent_UnmarshalJSON` now that there's a more local test to check the same thing, but it's up to you.

paymentintent/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+// PaymentIntent returns the most recent PaymentIntent
+// visited by a call to Next.
+func (i *Iter) PaymentIntent() *stripe.PaymentIntent {
+	return i.Current().(*stripe.PaymentIntent)
+}
+
+func getC() Client {
+	return Client{stripe.GetBackend(stripe.APIBackend), stripe.Key}
+}
```
brandur-stripe: Would you mind taking a stab at getting the new `paymentintent/` package to pass Golint? Basically you'd be doing what I did [in the PR that activated it for account and charge](https://github.com/stripe/stripe-go/pull/593/files):

1. Get Golint running locally:

    ```
    go get -u golang.org/x/lint/golint
    golint ./paymentintent
    ```

2. Fix the linting problems (using the same style of comment from `./charge` basically:

    ```
    paymentintent/client.go:1:1: package comment should be of the form "Package paymentintent ..."
    paymentintent/client.go:23:1: exported method Client.New should have comment or be unexported
    paymentintent/client.go:24:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:35:1: exported method Client.Get should have comment or be unexported
    paymentintent/client.go:37:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:48:1: exported method Client.Update should have comment or be unexported
    paymentintent/client.go:50:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:61:1: exported method Client.Cancel should have comment or be unexported
    paymentintent/client.go:63:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:74:1: exported method Client.Capture should have comment or be unexported
    paymentintent/client.go:76:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:87:1: exported method Client.Confirm should have comment or be unexported
    paymentintent/client.go:89:2: don't use underscores in Go names; var payment_intent should be paymentIntent
    paymentintent/client.go:100:1: exported method Client.List should have comment or be unexported
    ```

3. Add the package to the whitelist in `Makefile`:

	```
	lint:
		golint -set_exit_status ./account
		golint -set_exit_status ./charge
		golint -set_exit_status ./paymentintent
	```

paymentintent.go
```diff
@@ -0,0 +1,167 @@
...
+
+	var err error
+	*s = PaymentIntentSourceAction(v)
+
+	switch s.Type {
+	case PaymentIntentNextActionAuthorizeWithURL:
+		err = json.Unmarshal(data, &s.Value.AuthorizeWithUrl)
+	}
+
+	return err
```
brandur-stripe: This `UnmarshalJSON` implementation is fine by me!

paymentintent/client.go
```diff
@@ -0,0 +1,129 @@
...
+
+// New POSTs a new payment_intent.
+// For more details see https://stripe.com/docs/api#create_payment_intent.
+func New(params *stripe.PaymentIntentParams) (*stripe.PaymentIntent, error) {
+	return getC().New(params)
+}
+
+func (c Client) New(params *stripe.PaymentIntentParams) (*stripe.PaymentIntent, error) {
+	payment_intent := &stripe.PaymentIntent{}
```
brandur-stripe: And LOL, I didn't even notice these right away, but saw them in the Linter's error messages, but these of course should be `paymentIntent` instead of `payment_intent`.

client/api.go
remi-stripe: yep I can take after this PR, I did not want to mix both

paymentintent.go
remi-stripe: it's a param only

brandur-stripe:
@remi-stripe Okay, here's a working `UnmarshalJSON` implementation that I came up with:

``` go
func (s *PaymentIntentSourceAction) UnmarshalJSON(data []byte) error {
	type paymentIntentSourceAction PaymentIntentSourceAction
	var v paymentIntentSourceAction
	if err := json.Unmarshal(data, &v); err != nil {
		return err
	}

	var err error
	*s = PaymentIntentSourceAction(v)
	s.Value = &PaymentIntentSourceActionValue{}

	// Unmarshal data a second time so that we can get the raw bytes for the
	// `value` field
	var rawObject map[string]*json.RawMessage
	if err := json.Unmarshal(data, &rawObject); err != nil {
		return err
	}

	switch s.Type {
	case PaymentIntentNextActionAuthorizeWithURL:
		err = json.Unmarshal(*rawObject["value"], &s.Value.AuthorizeWithUrl)
	}

	return err
}
```

And also, change the struct definition so that we have no JSON annotation for `Value`:

``` go
type PaymentIntentSourceAction struct {
	Type  PaymentIntentNextActionType     `json:"type"`
	Value *PaymentIntentSourceActionValue `json:"-"` // See custom UnmarshalJSON implementation
}
```

I think overall it's pretty clean, with just a couple caveats:

1. It does have to call `Unmarshal` a second time. The first pass isn't doing much work though, so it's probably not too bad.
2. It's not ideal that we have to come up with yet another pattern for unmarshaling this late in the game :( I'm not sure there's much we can do about it here, but we it'd be kind of nice to get faster feedback from API library implementations for API review ... when there's a smell, it tends to be more noticeable here. The good news is that if we add strongly-typed structs for sources, `Source`'s `UnmarshalJSON` will probably look a lot like this one's (although not exactly).

remi-stripe:
Okay addressed all the comments and fixed the unmarshal logic thanks to Brandur. Only thing left is to decide changes to the `/confirm` parameters shape.

paymentintent/client.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/stripe/stripe-go/form"
+)
+
+// Client is used to invoke /payment_intents APIs.
+type Client struct {
+	B   stripe.Backend
+	Key string
+}
+
+// New creates a payment_intent.
```
brandur-stripe: Nit, but would you mind if we used the more human-readable "payment intent" for these comments?

brandur-stripe:
Thanks for getting all that review feedback in so quickly!

paymentintent/client.go
remi-stripe: damn search and replace. Nice catch and fixed!

paymentintent.go
```diff
@@ -0,0 +1,175 @@
...
+	Description         string                          `json:"description"`
+	Livemode            bool                            `json:"livemode"`
+	ID                  string                          `json:"id"`
+	Metadata            map[string]string               `json:"metadata"`
+	NextSourceAction    *PaymentIntentSourceAction      `json:"next_source_action"`
+	OnBehalfOf          *Account                        `json:"on_behalf_of"`
+	ReceiptEmail        string                          `json:"receipt_email"`
+	ReturnURL           string                          `json:"return_url"`
+	Shipping            ShippingDetails                 `json:"shipping"`
+	Source              *PaymentSource                  `json:"source"`
```
remi-stripe: Since we only support real Sources, should this be Source directly? I think so

paymentintent.go
```diff
@@ -0,0 +1,175 @@
...
+	Description         string                          `json:"description"`
+	Livemode            bool                            `json:"livemode"`
+	ID                  string                          `json:"id"`
+	Metadata            map[string]string               `json:"metadata"`
+	NextSourceAction    *PaymentIntentSourceAction      `json:"next_source_action"`
+	OnBehalfOf          *Account                        `json:"on_behalf_of"`
+	ReceiptEmail        string                          `json:"receipt_email"`
+	ReturnURL           string                          `json:"return_url"`
+	Shipping            ShippingDetails                 `json:"shipping"`
+	Source              *Source                         `json:"source"`
```
remi-stripe: Since PaymentIntent will only support Sources and nothing else, I think it's the right call to use `Source` here.

paymentintent.go
brandur-stripe: Arg, I just took a look at the source and it's allows to be either a source _or_ a card.

I really like the idea of it just being a `Source` as well, but we might have to check with the team for advice on how to handle this.

remi-stripe: yep, talked to them about it and we want to support Cards moving forward so it will stay a PaymentSource unfortunately

remi-stripe:
Okay now that stripe-mock is updated the tests run and I think it is now ready to merge but needs a full review to make sure we did not miss anything

paymentintent/client.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		return ret, list.ListMeta, err
+	})}
+}
+
+// Iter is an iterator for lists of PaymentIntents.
+// The embedded Iter carries methods with it;
+// see its documentation for details.
+type Iter struct {
```
brandur-stripe: I think we ended up updating all these `Iter` comments in your other Lint PRs. Can we change this one too? (And maybe the comment on `PaymentIntent()` directly below too.

brandur-stripe:
Left a minor comment on comments, but otherwise looks great.

paymentintent/client.go
remi-stripe: Nice catch, fixed other comments too

remi-stripe:
PTAL @brandur-stripe

brandur-stripe:
Thanks Remi!

brandur-stripe:
Released as 36.0.0.

brandur-stripe:
And I just realized that I jumped the gun on this a little bit because we're probably going to tweak a few parts of this API. I think it's okay though since this is still behind a gate and it won't really be possible to start relying on it yet, so we can make those changes without breaking anyone.

remi-stripe:
@brandur-stripe My understanding was that the API was stable and changes would be versioned/punted to later on so I think it was ready to be merged. At least that's the feeling I got internally!

brandur-stripe:
@remi-stripe Oh great! I'd thought for some reason that some of the changes we'd talked about a few weeks ago were still WIP. Very happy to hear that they're already good to go. Thanks.

PR #607, opened 2018-06-28
Adds support for 'partner_id' to 'SetAppInfo'

zachwick:
This adds support for an optional partner_id attribute in the SetAppInfo hash.

### DIFF
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -157,7 +157,7 @@ func TestUserAgent(t *testing.T) {
 func TestUserAgentWithAppInfo(t *testing.T) {
 	appInfo := &stripe.AppInfo{
 		Name:      "MyAwesomePlugin",
-		PartnerId: "partner_1234",
+		PartnerID: "partner_1234",
 		URL:       "https://myawesomeplugin.info",
 		Version:   "1.2.34",
 	}
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -196,6 +196,7 @@ func TestUserAgentWithAppInfo(t *testing.T) {
 	assert.Equal(t, "MyAwesomePlugin", application["name"])
 	assert.Equal(t, "https://myawesomeplugin.info", application["url"])
 	assert.Equal(t, "1.2.34", application["version"])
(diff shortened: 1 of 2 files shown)


remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

stripe_test.go
```diff
@@ -156,9 +156,10 @@ func TestUserAgent(t *testing.T) {
 
 func TestUserAgentWithAppInfo(t *testing.T) {
 	appInfo := &stripe.AppInfo{
-		Name:    "MyAwesomePlugin",
-		URL:     "https://myawesomeplugin.info",
-		Version: "1.2.34",
+		Name:      "MyAwesomePlugin",
+		PartnerId: "partner_1234",
```
remi-stripe: since you changed the input here, you should update the test below though to ensure it's passed properly

brandur-stripe: @remi-stripe The confusion here is partly my mistake — Zach had originally added the new partner ID to `User-Agent`, but I asked him to take it out again because it didn't really seem to fit the format of that field very well. Therefore, the new field is _only_ sent as a new field in the specially encoded `X-Stripe-User-Agent` header instead.

The test below only checks `User-Agent`, which is why it wasn't updated I think.

remi-stripe: Oh I see. Should we add a test for it to be safe since it's not really an API parameter?

brandur-stripe: Yeah I guess it can't hurt. I added #609 to bring in some basic tests for the existing parameters. We can merge that first and then have Zach rebase and add one new assertion for "partner ID".

zachwick:
ptal @brandur-stripe

brandur-stripe:
LGTM Zach.

One more thing: I just merged #609. Would you mind rebasing on `master` and then adding one assertion for the new partnerID field for that test? Thanks!

zachwick:
ptal @brandur-stripe

brandur-stripe:
Thanks!

brandur-stripe:
Released as 35.8.0.

PR #609, opened 2018-06-28
Add some "app info" tests for `X-Stripe-Client-User-Agent`

brandur:
The current tests for "app info" test against `User-Agent`, but don't
test against the `X-Stripe-Client-User-Agent` header. This patch adds a
few extra tests for safety.

Related: #607.

r? @remi-stripe


brandur-stripe:
And thanks for the quick review!

PR #610, opened 2018-06-28
Lint a few more packages

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Only did a few packages. Minor but I don't think `go test ./...` is able to catch linter errors which makes it easy to miss later. I'm saying this because I added my packages to the whitelist and only saw errors when explicitly using `golint` and not when doing `go test ./...`

### DIFF
--- a/applepaydomain/client.go
+++ b/applepaydomain/client.go
@@ -8,28 +8,30 @@ import (
 	"github.com/stripe/stripe-go/form"
 )
 
-// Client is used to invoke /apple_pay/domains and ApplePayDomain-related APIs.
+// Client is used to invoke /apple_pay/domains and Apple Pay domain-related APIs.
 type Client struct {
 	B   stripe.Backend
 	Key string
 }
 
-// New POSTs new ApplePayDomains.
+// New creates a new Apple Pay domain.
 func New(params *stripe.ApplePayDomainParams) (*stripe.ApplePayDomain, error) {
 	return getC().New(params)
 }
 
+// New creates a new Apple Pay domain.
 func (c Client) New(params *stripe.ApplePayDomainParams) (*stripe.ApplePayDomain, error) {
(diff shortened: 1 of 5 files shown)


applepaydomain/client.go
```diff
@@ -14,22 +14,24 @@ type Client struct {
 	Key string
 }
 
-// New POSTs new ApplePayDomains.
+// New creates a new ApplePayDomain.
```
brandur-stripe: Minor, but I kind of think that we should try to use the "human style" form of "Apple Pay domain" in these comments (it's all intended for documentation). The function signature makes it obvious that a `stripe.ApplePayDomain` will be created.

balance/client.go
```diff
@@ -66,7 +66,7 @@ type Iter struct {
 	*stripe.Iter
 }
 
-// Charge returns the most recent BalanceTransaction
+// BalanceTransaction returns the most recent BalanceTransaction
```
brandur-stripe: And similarly "most recent balance transaction".

card/client.go
```diff
@@ -15,12 +15,12 @@ type Client struct {
 	Key string
 }
 
-// New POSTs new cards either for a customer or recipient.
-// For more details see https://stripe.com/docs/api#create_card.
+// New creates a new card
```
brandur-stripe: Can you add punctuation on this one and the other `New`?

remi-stripe:
> The default make target actually does a pretty good job of just running every known check for a slightly more thorough look to make sure everything is working.

Ohhhhh. I totally did not think I could just use `make` but it does work as expected so I'll know for next time.

remi-stripe:
@brandur-stripe fixed the comments, thanks for catching all of those. PTAL!

brandur-stripe:
LGTM. Thanks!

PR #611, opened 2018-06-29
Introduce `GetBackendWithConfig` and make logging configurable per backend

### LINKED ISSUE #608: Feature request: non-global logging
In a project that I am working on we have a policy of not allowing any globals or init functions and would prefer to avoid having a `setupStripeLogging(logger *log.Logger)` function. Instead, we would like a way to pass in a logger when we create our Clients or Backends (similar to how we pass dependencies down the stack as function arguments to everything else that needs them) instead of relying on global state and functions with side effects.

It seems to me like it would make the most sense to add a logger on the `BackendConfiguration` struct since that's where all logging appears to happen. This could be checked first and if it was nil the global logger could be used to maintain backwards compatibility.

To use it, a new function to create a backend with a provided configuration could be added (as far as I can tell right now the `BackendConfiguration` struct isn't usable and doesn't need to be exported, this would change that):

> func GetBackendCfg(backend SupportedBackend, config BackendConfiguration) Backend

If this is too broad (it would also let you configure everything else in BackendConfiguration) something that only let you setup logging is an option:

> func GetLoggingBackend(backend SupportedBackend, logger *log.Logger) Backend

brandur:
Introduces a new system for getting backends with a set of custom
configuration that's a little cleaner than adding a new setter for every
option that we add in the future.

`BackendConfig` is introduced a user configurable struct which is then
passed to `GetBackendWithConfig`, which returns a backend configured to
spec. `GetBackend` remains, is backward compatible, and continues to set
a default backend for either type.

A new option on `BackendConfig` is being able to set a per-backend
`Logger` and `LogLevel`. This is the impetus that sparked this change as
documented in #608.

So the unfortunate part is that `BackendConfig` is confusingly named
similarly to the very poorly named `BackendConfiguration`, which is the
final implementation of a backend. I wanted to keep this patch backward
compatible, but I want to make a breaking change in the future to rename
this to `backendImplementation` and make it unexported at the same time
because it shouldn't be needed outside of the core package.

We should also unexport `NewBackends` at the same time. It was
introduced to allow some configuration for backends, but it was a pretty
bad idea.

Fixes #608.

r? @remi-stripe

### DIFF
--- a/order/client_test.go
+++ b/order/client_test.go
@@ -50,12 +50,6 @@ func TestOrderReturn(t *testing.T) {
 }
 
 func TestOrderReturn_RequestParams(t *testing.T) {
-	// Restore the backend when this test cases completes.
-	backend := stripe.GetBackend("api")
-	defer func() {
-		stripe.SetBackend("api", backend)
-	}()
-
 	// Create an ephemeral test server so that we can inspect request attributes.
 	var lastRequest *http.Request
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
@@ -65,16 +59,15 @@ func TestOrderReturn_RequestParams(t *testing.T) {
 	defer ts.Close()
 
 	// Configure the stripe client to use the ephemeral backend.
-	stripe.SetBackend("api", &stripe.BackendConfiguration{
-		Type:       stripe.APIBackend,
-		URL:        ts.URL + "/v1",
-		HTTPClient: &http.Client{},
+	backend := stripe.GetBackendWithConfig(stripe.APIBackend, &stripe.BackendConfig{
+		URL: ts.URL,
 	})
+	orderClient := Client{B: backend, Key: stripe.Key}
 
 	p := &stripe.OrderReturnParams{}
 	p.SetStripeAccount("acct_123")
 
-	order, err := Return("or_123", p)
+	order, err := orderClient.Return("or_123", p)
 	assert.Nil(t, err)
 	assert.NotNil(t, order)
 
--- a/stripe.go
+++ b/stripe.go
@@ -61,7 +61,10 @@ var LogLevel = 2
 
 // Logger controls how stripe performs logging at a package level. It is useful
 // to customise if you need it prefixed for your application to meet other
-// requirements
+// requirements.
+//
+// This Logger will be inherited by any backends created by default, but will
+// be overridden if a backend is created with GetBackendWithConfig.
 var Logger Printfer
 
 //
@@ -101,12 +104,54 @@ type Backend interface {
 	SetMaxNetworkRetries(maxNetworkRetries int)
 }
 
-// BackendConfiguration is the internal implementation for making HTTP calls to Stripe.
+// BackendConfig is used to configure a new Stripe backend.
+type BackendConfig struct {
+	// HTTPClient is an HTTP client instance to use when making API requests.
+	//
+	// If left unset, it'll be set to a default HTTP client for the package.
+	HTTPClient *http.Client
+
+	// LogLevel is the logging level of the library and defined by:
+	//
+	// 0: no logging
+	// 1: errors only
+	// 2: errors + informational (default)
+	// 3: errors + informational + debug
+	//
+	// Defaults to 0 (no logging), so please make sure to set this if you want
+	// to see logging output in your custom configuration.
+	LogLevel int
+
+	// Logger is where this backend will write its logs.
+	//
+	// If left unset, it'll be set to Logger.
+	Logger Printfer
+
+	// MaxNetworkRetries sets maximum number of times that the library will
+	// retry requests that appear to have failed due to an intermittent
+	// problem.
+	//
+	// Defaults to 0.
+	MaxNetworkRetries int
+
+	// URL is the base URL to use for API paths.
+	//
+	// If left empty, it'll be set to the default for the SupportedBackend.
+	URL string
+}
+
+// BackendConfiguration is the internal implementation for making HTTP calls to
+// Stripe.
+//
+// The public use of this struct is deprecated. It will be renamed and changed
+// to unexported in a future version.
 type BackendConfiguration struct {
 	Type              SupportedBackend
 	URL               string
 	HTTPClient        *http.Client
 	MaxNetworkRetries int
+	LogLevel          int
+	Logger            Printfer
 }
 
 // Call is the Backend.Call implementation for invoking Stripe APIs.
@@ -185,8 +230,8 @@ func (s *BackendConfiguration) NewRequest(method, path, key, contentType string,
 
 	req, err := http.NewRequest(method, path, body)
 	if err != nil {
-		if LogLevel > 0 {
-			Logger.Printf("Cannot create Stripe request: %v\n", err)
+		if s.LogLevel > 0 {
+			s.Logger.Printf("Cannot create Stripe request: %v\n", err)
 		}
 		return nil, err
 	}
@@ -233,8 +278,8 @@ func (s *BackendConfiguration) NewRequest(method, path, key, contentType string,
 // the backend's HTTP client to execute the request and unmarshals the response
 // into v. It also handles unmarshaling errors returned by the API.
 func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
-	if LogLevel > 1 {
-		Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
+	if s.LogLevel > 1 {
+		s.Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
 	}
 
 	start := time.Now()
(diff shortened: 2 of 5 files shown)


stripe_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestBearerAuth(t *testing.T) {
-	c := &stripe.BackendConfiguration{URL: stripe.APIURL}
+	c := stripe.GetBackend(stripe.APIBackend).(*stripe.BackendConfiguration)
```
brandur-stripe: Dang it. I kind of punted on finding the right way to solve this for now.

It's an improvement that we now use `GetBackend` to get a default backend with good values set, but it's still bad that we then need to cast it back to `BackendConfiguration`. The problem is that each of these tests uses `BackendConfiguration.NewRequest`, and that is a function that's not part of the `Backend` interface.

We should either add it to the interface, or find a better way to write these tests, with a preference for the latter.

order/client_test.go
```diff
@@ -65,16 +59,15 @@ func TestOrderReturn_RequestParams(t *testing.T) {
...
 
 	// Configure the stripe client to use the ephemeral backend.
-	stripe.SetBackend("api", &stripe.BackendConfiguration{
-		Type:       stripe.APIBackend,
-		URL:        ts.URL + "/v1",
-		HTTPClient: &http.Client{},
+	backend := stripe.GetBackendWithConfig(stripe.APIBackend, &stripe.BackendConfig{
+		URL: ts.URL,
 	})
+	orderClient := Client{B: backend, Key: stripe.Key}
```
brandur-stripe: Instantiating a new client turned out to be a far cleaner way to implement this test anyway (the `defer` to reset the default backend at the top is no longer needed).

stripe.go
```diff
@@ -101,12 +104,53 @@ type Backend interface {
 	SetMaxNetworkRetries(maxNetworkRetries int)
 }
 
-// BackendConfiguration is the internal implementation for making HTTP calls to Stripe.
+// BackendConfig is used to configure a new Stripe backend.
+type BackendConfig struct {
+	// HTTPClient is an HTTP client instance to use when maxing requests.
```
SamWhited: s/maxing/making/

stripe.go
```diff
@@ -101,12 +104,53 @@ type Backend interface {
...
+	// LogLevel is the logging level of the library and defined by:
+	//
+	// 0: no logging
+	// 1: errors only
+	// 2: errors + informational (default)
+	// 3: errors + informational + debug
+	//
+	// Defaults to 0 (no logging), so please make sure to set this if you want
+	// to see logging output in your custom configuration.
+	LogLevel int
```
SamWhited: I don't think this is ever used; should it have a value in `BackendConfiguration` and be used there too? Given that it will default to zero it may not be possible to determine if we should use this or the global LogLevel, I'm not sure how to do that in a backwards compatible way.

stripe.go
```diff
@@ -101,12 +104,53 @@ type Backend interface {
...
+	//
+	// If left unset, it'll be set to Logger.
+	Logger Printfer
+
+	// MaxNetworkRetries sets maximum number of times that the library will
+	// retry requests that appear to have failed due to an intermittent
+	// problem.
+	//
+	// Defaults to 0.
+	MaxNetworkRetries int
```
SamWhited: Same as `LogLevel`.

SamWhited:
Wow, that was fast; thanks for implementing this!

stripe.go
brandur-stripe: Thanks! Fixed.

stripe.go
brandur-stripe: > I don't think this is ever used; should it have a value in BackendConfiguration and be used there too?

Haha, that's what I get for trying to push through a change like this near the end of the day!

Yep, that's a major omission. I just added `LogLevel` to `BackendConfiguration` and changed all existing checks on `LogLevel` to use it.

stripe.go
brandur-stripe: Thanks! Fixed that as well. (Its value was already on `BackendConfiguration`, but it was being carried over from the config struct.)

brandur-stripe:
> Given that it will default to zero it may not be possible to determine if we should use this or the global LogLevel, I'm not sure how to do that in a backwards compatible way.

Yeah, the inability to check on `0` versus unset is frustrating.

I believe I've implemented this to be backwards compatible though — when getting a default backend (i.e., _not_ through `GetBackendWithConfig`), we simply always set `backendConfig.LogLevel` to `stripe.LogLevel`, so it'll carry through.

There are still problems in other areas though. For example, if we want to give either `LogLevel` or `MaxNetworkRetries` a default value that's _not_ `0` when passing a `BackendConfig` (which is quite plausible), we then won't be able to tell whether a user has left them unset or explicitly set them to `0`.

One possibility is to set these to `*int64` like we do in our `*Params` structs and ask people to use the `stripe.Int64(...)` helper to set them. It's still a little weird given that `HTTPClient`, `Logger` etc. aren't set that way, but maybe it's fine. Thoughts?

brandur-stripe:
(Oh, and thanks for the review @SamWhited!)

stripe.go
```diff
@@ -233,8 +278,8 @@ func (s *BackendConfiguration) NewRequest(method, path, key, contentType string,
 // the backend's HTTP client to execute the request and unmarshals the response
 // into v. It also handles unmarshaling errors returned by the API.
 func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
-	if LogLevel > 1 {
-		Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
+	if s.LogLevel > 1 {
```
brandur-stripe: We should also make helpers for `logInfo`, `logError`, `logDebug` so we don't have these ugly checks on LogLevel everywhere, but I'll tackle that in another PR.

SamWhited:
> We should also make helpers for logInfo, logError, logDebug so we don't have these ugly checks on LogLevel everywhere, but I'll tackle that in another PR.
> …
> One possibility is to set these to *int64 like we do in our *Params structs and ask people to use the stripe.Int64(...) helper to set them. It's still a little weird given that HTTPClient, Logger etc. aren't set that way, but maybe it's fine. Thoughts?

Depending on how closely you want this to match what's already being done globally, one option might be to get rid of `LogLevel` in the config struct and have separate loggers for each level, eg. `logInfo *log.Logger`, `logError *log.Logger`, etc. and set each one to a nil Printerf implementation by default.

brandur-stripe:
> Depending on how closely you want this to match what's already being done globally, one option might be to get rid of LogLevel in the config struct and have separate loggers for each level, eg. LogInfo *log.Logger, LogError *log.Logger, etc. and set each one to a nop Printerf implementation by default.

Ah, true. I'd be a little hesitant to introduce a new paradigm here though — leveled logging where the wanted level is set just once is a pretty widespread paradigm (even in Go). 

I just looked at Zap and Logrus, two popular libraries, and they both basically use an identical convention here:

``` go
log.SetLevel(log.WarnLevel)
```

(Which is probably roughly what we should be doing as well, even if it looks like just setting `LogLevel` on `BackendConfig`.)

The one major departure compared to what we have is that normally the level is set on the logging object itself instead of as a separate field on the construct that holds the logger. I kind of want to change `Printfer` to include more functions to do something similar:

``` go
type Printfer interface {
	Errorf(format string, v ...interface{})
	Infof(format string, v ...interface{})
        SetLogLevel(...)
	Warnf(format string, v ...interface{})
}
```

This would make using Zap and Logrus easy, but then of course Go's built in logging module would no longer comply to the interface. Sigh.

I _think_ the right thing to do for now is to leave it more or less as I have here, and we can decide on a larger breaking change later if it comes up again.

stripe.go
```diff
@@ -462,32 +509,99 @@ func FormatURLPath(format string, params ...string) string {
...
+		if config.URL == "" {
+			config.URL = apiURL
+		}
+
+		// Add the /v1/ prefix because all client packages expect it. We should
+		// probably change this to just make it explicit wherever it's needed
+		// to fulfill the principle of least astonishment.
+		config.URL += "/v1"
+
+		return &BackendConfiguration{
```
remi-stripe: this like the code is identical between both switch except for the url, could we generalize that?

stripe.go
```diff
@@ -462,32 +509,99 @@ func FormatURLPath(format string, params ...string) string {
...
+	}
+	backends.mu.RUnlock()
+	if backend != nil {
+		return backend
+	}
+
+	backend = GetBackendWithConfig(
+		backendType,
+		&BackendConfig{
+			HTTPClient:        httpClient,
```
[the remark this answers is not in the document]
brandur-stripe: Good eye. Yeah, I went back in forth on this — I settled for what you see because I felt that the explicitness is somewhat advantageous compared to setting a bunch of `0`/`nil` and then having to follow the code down to `GetBackendWithConfig` to see what's going to end up in there.

I don't feel too strongly about it though, and am happy to go with the other way too. What do you think?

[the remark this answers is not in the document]
remi-stripe: I don't feel strongly either so I'm fine leaving as is

stripe.go
brandur-stripe: Great point. I just updated the code and it looks way cleaner now.

brandur-stripe:
> To be honest I don't 100% follow the logic/usage of the code though but I think I understood the changes

Yeah sorry, but unfortunately the relations of `Backend`, `BackendConfig`, `BackendConfiguration`, `SupportedBackend`, etc. are all still pretty confusing right now :/

I think I can clean this up a lot in the future and make this way easier to wrap your head around, but I wanted to target a non-breaking change for this PR and see if we can improve things a bit later.

ptal @remi-stripe Updated the patch, please take one more look.

brandur-stripe:
Thanks Remi!

brandur-stripe:
Released as 35.9.0.

SamWhited:
> Released as 35.9.0.

Fantastic, thanks again for your quick response! This will let me clean up a lot of hacky race-condition-prone stuff on my end where I had to make some logging global to pass it between modules and apply it to Stripe. Looking forward to cleaning that up!

brandur-stripe:
> Fantastic, thanks again for your quick response! This will let me clean up a lot of hacky race-condition-prone stuff on my end where I had to make some logging global to pass it between modules and apply it to Stripe. Looking forward to cleaning that up!

No worries! I hope it works out.

PR #613, opened 2018-07-03
More packages properly linted

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/event/client.go
+++ b/event/client.go
@@ -14,12 +14,12 @@ type Client struct {
 	Key string
 }
 
-// Get returns the details of an event
-// For more details see https://stripe.com/docs/api#retrieve_event.
+// Get returns the details of an event.
 func Get(id string, params *stripe.EventParams) (*stripe.Event, error) {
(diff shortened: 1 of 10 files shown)


event/client.go
```diff
@@ -14,12 +14,12 @@ type Client struct {
...
 }
 
-// Get returns the details of an event
-// For more details see https://stripe.com/docs/api#retrieve_event.
+// Get returns the details of an event.
 func Get(id string, params *stripe.EventParams) (*stripe.Event, error) {
 	return getC().Get(id, params)
 }
 
+// Get returns the details of an event.
```
remi-stripe: do we care about this wording? We don't always say "the details of". I think it's mostly fine but wanted to flag

brandur-stripe: Yeah, seems fine to me, especially in areas like this where the resource name is somewhat vague.

brandur-stripe:
LGTM.

Issue #614, opened 2018-07-04
UsageRecord missing from client.API struct.


andrewchambers:
It seems this is a new API, my impression it should be initialized and present in the client.API struct.

remi-stripe:
@andrewchambers Thanks for the report! I just opened a PR to fix this issue.

brandur-stripe:
Remi's fixed has been released as 35.10.0.

PR #616, opened 2018-07-04
Add missing packages to the list of clients

remi-stripe:
This PR starts by re-ordering all packages alphabetically to make it easier to maintain. It then adds all the packages that were missing from the list.

This fixes https://github.com/stripe/stripe-go/issues/614

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -43,111 +43,78 @@ import (
 
 // API is the Stripe client. It contains all the different resources available.
 type API struct {
-	// Charges is the client used to invoke /charges APIs.
-	// For more details see https://stripe.com/docs/api#charges.
-	Charges *charge.Client
-	// Customers is the client used to invoke /customers APIs.
-	// For more details see https://stripe.com/docs/api#customers.
-	Customers *customer.Client
-	// Cards is the client used to invoke /cards APIs.
-	// For more details see https://stripe.com/docs/api#cards.
-	Cards *card.Client
-	// Subscriptions is the client used to invoke /subscriptions APIs.
-	// For more details see https://stripe.com/docs/api#subscriptions.
-	Subscriptions *sub.Client
-	// SubscriptionItems is the client used to invoke subscription's items-related APIs.
-	// For more details see https://stripe.com/docs/api#subscription_items.
-	SubscriptionItems *subitem.Client
-	// Plans is the client used to invoke /plans APIs.
-	// For more details see https://stripe.com/docs/api#plans.
-	Plans *plan.Client
-	// Coupons is the client used to invoke /coupons APIs.
-	// For more details see https://stripe.com/docs/api#coupons.
-	Coupons *coupon.Client
-	// Discounts is the client used to invoke discount-related APIs.
-	// For mode details see https://stripe.com/docs/api#discounts.
-	Discounts *discount.Client
-	// Invoices is the client used to invoke /invoices APIs.
-	// For more details see https://stripe.com/docs/api#invoices.
-	Invoices *invoice.Client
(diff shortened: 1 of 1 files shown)


client/api.go
```diff
@@ -35,119 +37,99 @@ import (
...
 	"github.com/stripe/stripe-go/token"
+	"github.com/stripe/stripe-go/topup"
 	"github.com/stripe/stripe-go/transfer"
+	"github.com/stripe/stripe-go/usagerecord"
 )
 
 // API is the Stripe client. It contains all the different resources available.
 type API struct {
+	// Account is the client used to invoke /accounts APIs.
+	Account *account.Client
```
remi-stripe: Do you know why it says `Account` and not `Accounts`? Should we fix it?

client/api.go
```diff
@@ -35,119 +37,99 @@ import (
...
-	// Disputes is the client used to invoke dispute-related APIs.
-	// For more details see https://stripe.com/docs/api#disputes.
-	Disputes *dispute.Client
-	// Transfers is the client used to invoke /transfers APIs.
-	// For more details see https://stripe.com/docs/api#transfers.
-	Transfers *transfer.Client
+	// Orders is the client used to invoke /orders APIs.
+	Orders *order.Client
+	// OrderReturns is the client used to invoke /order_returns APIs.
+	OrderReturns *orderreturn.Client
```
remi-stripe: Was not sure if `Orders` was before `OrderReturns` (as in whether the case matters in alphabetical ordering). It seemed easier/cleaner this way though. (true for other packages like sub below)

client/api.go
```diff
@@ -35,119 +37,99 @@ import (
...
-	// ExchangeRates is the client used to invoke /exchange_rates APIs.
-	ExchangeRates *exchangerate.Client
+	// SourceTransactions is the client used to invoke source transaction related APIs.
+	SourceTransactions *sourcetransaction.Client
+	// Subscriptions is the client used to invoke /subscriptions APIs.
+	Subscriptions *sub.Client
+	// SubscriptionItems is the client used to invoke subscription's items related APIs.
+	SubscriptionItems *subitem.Client
+	// Tokens is the client used to invoke /tokens APIs.
+	Tokens *token.Client
```
remi-stripe: I did not add `threedsecure` since it's been deprecated for well over a year. Is it the right call?

client/api.go
brandur-stripe: Hm, I'd guess because maybe the original package started out by just doing `GET /account`, which was singular.

Yeah, I think we should fix it, but maybe not now since it's breaking — maybe add a TODO or something here?

client/api.go
brandur-stripe: Technically I think `OrderReturns` would come first. I don't think it matters that much though — like you said, this will be more of a sane ordering for users reading through this list.

client/api.go
brandur-stripe: Yeah, if no one's asked for it by now, let's just leave it out.

brandur-stripe:
Minor comment about possibly adding a TODO comment or something above, but otherwise LGTM. Thanks for the quick fix.

ptal @remi-stripe

client/api.go
remi-stripe: Sounds good. I think we can punt on the TODO and I'll change it when we go ahead and remove the old 3DS endpoint

remi-stripe:
I think we should punt on the TODO and fix later, I'll take note of it so that we don't forget it.

brandur-stripe:
> I think we should punt on the TODO and fix later, I'll take note of it so that we don't forget it.

Works for me.

PR #620, opened 2018-07-05
Add `Quantity` and `UnitAmount` to `InvoiceItemParams`

### LINKED ISSUE #619: Missing fields in InvoiceItemParams
The fields `unit_amount` and `quantity` are both missing from InvoiceItemParams. These are definitely supported by the server because they are used by the admin UI to create invoice items (as shown by examining the POST'd JSON).

brandur:
Both these parameters were missing from the params struct, but are
documented in the API [1]. This patch adds them in.

Fixes #619.

[1] https://stripe.com/docs/api/php#create_invoiceitem

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/invoiceitem.go
+++ b/invoiceitem.go
@@ -12,7 +12,9 @@ type InvoiceItemParams struct {
 	Description  *string `form:"description"`
 	Discountable *bool   `form:"discountable"`
 	Invoice      *string `form:"invoice"`
+	Quantity     *int64  `form:"quantity"`
 	Subscription *string `form:"subscription"`
+	UnitAmount   *int64  `form:"unit_amount"`
 }
 
 // InvoiceItemListParams is the set of parameters that can be used when listing invoice items.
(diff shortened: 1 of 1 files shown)


invoiceitem.go
```diff
@@ -12,7 +12,9 @@ type InvoiceItemParams struct {
 	Description  *string `form:"description"`
 	Discountable *bool   `form:"discountable"`
 	Invoice      *string `form:"invoice"`
+	Quantity     *int64  `form:"quantity"`
 	Subscription *string `form:"subscription"`
+	UnitAmount   *int64  `form:"unit_amount"`
 }
```
remi-stripe: I think you want to add them to the `InvoiceItem` struct too

brandur-stripe: Yes ... that would make sense. I just pushed that change (only `UnitAmount` was missing).

brandur-stripe:
Thanks Remi.

Issue #621, opened 2018-07-06
DeclineChargeOn present on LegalEntityParams instead of AccountParams


sfriedel:
This is a follow-up issue to #617.

It appears that the `DeclineChargeOn` field was added to `LegalEntityParams` while it probably should go to `AccountParams`.

brandur-stripe:
Sent Remi's fix out on 35.13.0.

sfriedel:
Thank you once again

PR #622, opened 2018-07-06
Fix decline_charge_on on the Account ressource

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/621

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
Oops. Alright, I'm going to push this as a minor because it was new and in the wrong place.

PR #623, opened 2018-07-09
Changed Payout.Destination from string to *PayoutDestination to support expanding

slavick:


remi-stripe:
@slavick Thank you so much for the fix! Sorry we missed this one. We do have custom code to handle the deserialization but it seems I incorrectly moved it to a `string` in a recent major version.

r? @brandur-stripe 
cc @stripe/api-libraries

brandur-stripe:
Thank you Slavick!

brandur-stripe:
Released as 36.0.0.

PR #628, opened 2018-07-18
Use HTTPS and HTTP/2 from the test suite

brandur:
Switches the test suite over to use HTTPS and HTTP/2 against stripe-mock
using its new `-https` flag and built-in HTTP/2 support in Go.

Note that this won't quite work until we've merged stripe/stripe-mock#82 which
adds the `-https` option. I've tested it locally though, and verified that it
does indeed use HTTP/2.


brandur-stripe:
Thanks Remi!

Alright, this now passes @ stripe-mock 0.20.1. Pulling it in.

PR #630, opened 2018-07-18
Switch to port 12112 as default stripe-mock port for HTTPS

brandur:
See stripe/stripe-mock#84 for context, but I'm going to switch to port
`12112` as the new default for stripe-mock running HTTPS. This will
allow us to have a setup where we can run API libraries that use HTTP
and those that use HTTPS with just a single instance of stripe-mock.

Depends on stripe/stripe-mock#84 being merged first, and a new
stripe-mock release being cut.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
       tar -zxf "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz" -C "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/"
     fi
   - |
-    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock -https > /dev/null &
+    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock -https-port 12112 > /dev/null &
     STRIPE_MOCK_PID=$!
 
 cache:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,7 +20,7 @@ before_install:
       tar -zxf "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz" -C "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/"
     fi
   - |
-    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock -https > /dev/null &
+    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock -https-port 12112 > /dev/null &
```
ob-stripe: You're missing a space character before the `-port` option.

brandur-stripe: Doh, okay sorry this new set of options can be a little bit confusing, but in this case it's actually allowed to be either `-https -port 12112` _or_ `-https-port 12112`. The latter is required if you want simultaneous HTTP and HTTPS, but it can also be used by itself.

ob-stripe: Ah right, of course 🤦‍♂️

brandur-stripe:
Thanks OB!

I've upgraded stripe-mock here to 0.21.1 and confirmed that it plays nicely with a Homebrew-installed stripe-mock (while also allowing stripe-ruby and other libraries that are not using HTTPS to also play nicely with it).

PR #633, opened 2018-07-24
Fix encoding of list params for bank accounts and cards

### LINKED ISSUE #632: Using ListParams in BankAccountListParams results in unexpected query params in URL
While mocking part of the Stripe API for use in our test suite I was noticing that setting `ListParams` in `BankAccountListParams` produces unexpected and possibly invalid query params in the requested URL.

I prepared a short test program to demonstrate the issue:

```go
package main

import (
	"fmt"
	"net/http"
	"net/http/httptest"

	"github.com/stripe/stripe-go"
	"github.com/stripe/stripe-go/client"
)

func main() {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Println("URI:", r.RequestURI)
		fmt.Println("QUERY:", r.URL.Query())
		w.WriteHeader(500)
	}))
	backends := &stripe.Backends{
		API: &stripe.BackendConfiguration{
			Type:       stripe.APIBackend,
			URL:        server.URL,
			HTTPClient: http.DefaultClient,
		},
		Uploads: &stripe.BackendConfiguration{
			Type:       stripe.UploadsBackend,
			URL:        server.URL,
			HTTPClient: http.DefaultClient,
		},
	}
	c := client.New("sk_xxxxx", backends)
	p := &stripe.BankAccountListParams{
		Account: stripe.String("acc_xxxxxx"),
		ListParams: stripe.ListParams{
			StartingAfter: stripe.String("ba_xxxxxx"),
			Limit:         stripe.Int64(1),
		},
	}
	c.BankAccounts.List(p)
}
```

When this is run, it will produce the following output:

```
$ go run test.go
URI: /accounts/acc_xxxxxx/external_accounts?object=bank_account?limit=1&starting_after=ba_xxxxxx
QUERY: map[object:[bank_account?limit=1] starting_after:[ba_xxxxxx]]
```

As can be seen in the QUERY output, the `object` param actually also contains the `limit` because they are separated by `?` instead of `&`.

It does not seem to cause any issues when used with the regular Stripe API but it might be undesired behavior anyway.

brandur:
Fixes a bug discovered in #632 whereby because we were manually adding a
filter like `?object=bank_account` to the end of paths, we'd clash with
the `form` package's attempt at an encoding if both came into play. This
would have the effect of producing an invalid URL like this:

```
/accounts/acc_xxxxxx/external_accounts?object=bank_account?limit=1&starting_after=ba_xxxxxx
```

(Note the multiple `?`s in the query string.)

Here we stop manually adding filters to URL paths and move them into a
custom `AppendTo` override for the `form` package. Bank accounts and
cards were the only places that we were doing this, so only they were
affected.

We also add tests to make sure that we're doing the right thing.

Fixes #632.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/bankaccount_test.go
+++ b/bankaccount_test.go
@@ -29,6 +29,26 @@ func TestBankAccount_UnmarshalJSON(t *testing.T) {
 	}
 }
 
+func TestBankAccountListParams_AppendTo(t *testing.T) {
+	// Adds `object` for account (this will hit the customer sources endpoint)
+	{
+		params := &BankAccountListParams{Account: String("acct_123")}
+		body := &form.Values{}
+		form.AppendTo(body, params)
+		t.Logf("body = %+v", body)
+		assert.Equal(t, []string{"bank_account"}, body.Get("object"))
+	}
+
+	// Adds `object` for customer (this will hit the external accounts endpoint)
+	{
+		params := &BankAccountListParams{Customer: String("cus_123")}
+		body := &form.Values{}
+		form.AppendTo(body, params)
+		t.Logf("body = %+v", body)
+		assert.Equal(t, []string{"bank_account"}, body.Get("object"))
+	}
+}
+
 func TestBankAccountParams_AppendToAsSourceOrExternalAccount(t *testing.T) {
 	// We should add more tests for all the various corner cases here ...
 
--- a/card_test.go
+++ b/card_test.go
@@ -5,7 +5,38 @@ import (
 	"testing"
 
 	assert "github.com/stretchr/testify/require"
+	"github.com/stripe/stripe-go/form"
 )
+
+func TestCardListParams_AppendTo(t *testing.T) {
+	// Adds `object` for account (this will hit the external accounts endpoint)
+	{
+		params := &CardListParams{Account: String("acct_123")}
+		body := &form.Values{}
+		form.AppendTo(body, params)
+		t.Logf("body = %+v", body)
+		assert.Equal(t, []string{"card"}, body.Get("object"))
+	}
+
+	// Adds `object` for customer (this will hit the sources endpoint)
+	{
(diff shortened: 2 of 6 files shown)


card_test.go
```diff
@@ -5,8 +5,39 @@ import (
...
+		params := &CardListParams{Account: String("acct_123")}
+		body := &form.Values{}
+		form.AppendTo(body, params)
+		t.Logf("body = %+v", body)
+		assert.Equal(t, []string{"card"}, body.Get("object"))
+	}
+
+	// Adds `object` for customer (this will hit the sources endpoint)
+	{
+		params := &CardListParams{Account: String("acct_123")}
```
remi-stripe: This should be `Customer: String("cus_123")` instead

bankaccount_test.go
```diff
@@ -29,6 +29,16 @@ func TestBankAccount_UnmarshalJSON(t *testing.T) {
...
 }
 
+func TestBankAccountListParams_AppendTo(t *testing.T) {
+	{
+		params := &BankAccountListParams{}
+		body := &form.Values{}
+		form.AppendTo(body, params)
+		t.Logf("body = %+v", body)
+		assert.Equal(t, []string{"bank_account"}, body.Get("object"))
+	}
```
remi-stripe: should you be testing Account and Customer like you do for cards?

card_test.go
brandur-stripe: Oops! Thank you. Fixed.

bankaccount_test.go
brandur-stripe: Yeah, I skipped it because the same conditional branch just adds `object` regardless of the circumstances, but I guess it can't hurt to be explicit. Broke it into two test cases like cards.

PR #637, opened 2018-07-28
Add support for ScheduledQueryRun

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/sigma/scheduledqueryrun/client_test.go
+++ b/sigma/scheduledqueryrun/client_test.go
@@ -1,0 +1,26 @@
+package scheduledqueryrun
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	stripe "github.com/stripe/stripe-go"
+	_ "github.com/stripe/stripe-go/testing"
+)
+
+func TestSigmaScheduledQueryRunGet(t *testing.T) {
+	run, err := Get("sqr_123", nil)
+	assert.Nil(t, err)
+	assert.NotNil(t, run)
+	assert.Equal(t, "scheduled_query_run", run.Object)
+}
+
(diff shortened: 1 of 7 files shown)


sigma/scheduledqueryrun/client_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	stripe "github.com/stripe/stripe-go"
+	_ "github.com/stripe/stripe-go/testing"
+)
+
+func TestSigmaScheduledQueryRunGet(t *testing.T) {
+	t.Skip("skipping test as stripe-mock does not support this resource")
```
remi-stripe: It would be really cool if we could add some way to stub requests like we do in other libraries!

brandur-stripe: Yes, that would be cool! Mocking isn't really possible in Go, but I think we can accomplish something similar with a helper to boot up a little JSON test server that responds with a blob you can hand it.

In this case though, I think we should make sure that stripe-mock supports scheduled query runs. I'm just winding that through the backend now.

brandur-stripe: Okay, can you try bumping stripe-mock to 0.24.0 and seeing if you can activate these tests? Thanks!

ptal @remi-stripe

remi-stripe: oh neat, thanks for shipping this one!

remi-stripe:
@brandur-stripe fixed, PTAL!

brandur-stripe:
Released as 37.0.0.

PR #639, opened 2018-07-30
Add support for the most recent API version

remi-stripe:
This ensures the most recent API version is set in stripe-go as there were a few breaking changes.

* Move to API version 2018-07-27
* Remove SKUs list on the product
* Remove the ability to pass a Source on subscription creation/update
* Change percent_off on the coupon from int to float

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/coupon.go
+++ b/coupon.go
@@ -15,15 +15,15 @@ const (
 // For more details see https://stripe.com/docs/api#create_coupon.
 type CouponParams struct {
 	Params           `form:"*"`
-	AmountOff        *int64  `form:"amount_off"`
-	Currency         *string `form:"currency"`
-	Duration         *string `form:"duration"`
-	DurationInMonths *int64  `form:"duration_in_months"`
-	ID               *string `form:"id"`
-	MaxRedemptions   *int64  `form:"max_redemptions"`
-	Name             *string `form:"name"`
-	PercentOff       *int64  `form:"percent_off"`
-	RedeemBy         *int64  `form:"redeem_by"`
+	AmountOff        *int64   `form:"amount_off"`
+	Currency         *string  `form:"currency"`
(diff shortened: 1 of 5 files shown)


coupon.go
```diff
@@ -15,15 +15,15 @@ const (
...
-	AmountOff        *int64  `form:"amount_off"`
-	Currency         *string `form:"currency"`
-	Duration         *string `form:"duration"`
-	DurationInMonths *int64  `form:"duration_in_months"`
-	ID               *string `form:"id"`
-	MaxRedemptions   *int64  `form:"max_redemptions"`
-	Name             *string `form:"name"`
-	PercentOff       *int64  `form:"percent_off"`
-	RedeemBy         *int64  `form:"redeem_by"`
+	AmountOff        *float64 `form:"amount_off"`
```
brandur-stripe: Was changing `AmountOff` and not `PercentOff` in parameters intentional? It kind of looks like they should both be the same (and I would've thought they'd be floats themselves, but integers seem to still be recommended in the docs).

ptal @remi-stripe

remi-stripe: @brandur-stripe Sorry, I changed the wrong one I think. Only `percent_off` should be a float. It's for the latest API version we did: https://stripe.com/docs/upgrades#2018-07-27

> The percent_off field of coupons was changed from Integer to Float, with a precision of two decimal places.

remi-stripe:
@brandur-stripe Sorry for the mixup, I fixed the issue and changed `percent_off` as expected. `amount_off` is still an int64 since it's always the amount in the smallest denomination for the currency

brandur-stripe:
Ah yes, that looks right now. Thanks!

brandur-stripe:
Released as 37.0.0.

PR #641, opened 2018-07-30
Enforce linter on all files and packages

remi-stripe:
This enforces the linter on all files and packages and fixes all the current errors

* Fix the makefile to run `golint ./...`
* Add comments to all const blocks to describe what they represent
* Fix comments on all structs and exported methods
* A few breaking changes due to naming of properties or variables:
  * `Account` resource where `IdentityVerificationDetailsCodeScanIdCountryNotSupported` is now `IdentityVerificationDetailsCodeScanIDCountryNotSupported`
  * `Account` resource where `IdentityVerificationDetailsCodeScanIdTypeNotSupported` is now `IdentityVerificationDetailsCodeScanIDTypeNotSupported`
  * `BitcointReceiver` resource where `BitcoinUri` is now `BitcoinURI`.
  * `IssuingAuthorization` resource where `NetworkId` is now `NetworkID`

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,64 +9,8 @@ build:
 check-gofmt:
 	scripts/check_gofmt.sh
 
-# We're trying to get linting activated, but there are so many errors that
-# we're doing so incrementally instead of pushing it through as part of one
-# giant patch.
-#
-# As new packages get cleaned up, add them here so that we don't regress. When
-# all packages are here, switch to a `./...` instead of linting every package
-# individually.
 lint:
-	golint -set_exit_status ./account
-	golint -set_exit_status ./applepaydomain
-	golint -set_exit_status ./balance
-	golint -set_exit_status ./bankaccount
-	golint -set_exit_status ./bitcoinreceiver
-	golint -set_exit_status ./bitcointransaction
-	golint -set_exit_status ./card
-	golint -set_exit_status ./charge
-	golint -set_exit_status ./countryspec
-	golint -set_exit_status ./coupon
-	golint -set_exit_status ./customer
-	golint -set_exit_status ./discount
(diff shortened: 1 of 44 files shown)


remi-stripe:
Flagging that there's one left that I'm not sure how to fix
```
client/api.go:5:2: should not use dot imports
```

remi-stripe:
Flagging that I got a "success" email for the build (2 really) though I expected it to fail since we now run the linter on the whole library.

brandur-stripe:
> Flagging that there's one left that I'm not sure how to fix

It just doesn't like this line:

``` go
import (
	. "github.com/stripe/stripe-go"
```

It's considered poor practice to use `.` because you're polluting the package's global namespace rather than referring to names via a symbol. Try changing this to a:

``` go
import (
```

And putting `stripe.` in front of any references that broke.

> Flagging that I got a "success" email for the build (2 really) though I expected it to fail since we now run the linter on the whole library.

You still need to specify the `-set_exit_status` flag for Golint. It defaults to exiting with `0` even if linting problems were found.

brandur-stripe:
And also: wow :) This is a serious piece of work!!

remi-stripe:
@brandur-stripe ha I tried a few things but did not think to rename the .! Thanks for the quick answer.

PTAL

remi-stripe:
@brandur-stripe Thanks for the quick review! Just fixed all of those, PTAL!

brandur-stripe:
Thanks Remi! LGTM.

brandur-stripe:
Released as 38.0.0.

Well, I hate to do two major versions in one day, but I figure that we may as well release this one while very few people have picked up version 37.

Issue #642, opened 2018-08-01
invalid character '<' looking for beginning of value


yns01:
Hey,

Some requests are failing with the following error:

`invalid character '<' looking for beginning of value`

I configured the number of retries to 3. Here's a timeline of what happened:

On the Stripe dashboard:

```
12:42:50 POST /charges ends with 200
12:42:52 POST /charges ends with 409
```

On my logs with `LogLevel = 5`:
```

[app logs] 2018-08-01T10:42:53.489Z\",\"caller\":\"logger/logger.go:36\",\"message\":\"Stripe request ended with an error: invalid character '<' looking for beginning of value\", 

[stripe-go logs] 
"2018/08/01 10:42:53 Retry request https://api.stripe.com/v1/charges 1 time.",
"2018/08/01 10:42:53 Stripe Response: \"{\\\\"id\\\": \\\"ch_*\\\",\
"2018/08/01 10:42:53 Completed in 2.784733977s",
"2018/08/01 10:42:52 Request failed with error: <nil>. Response: &{409....
"2018/08/01 10:42:52 Requesting POST api.stripe.com/v1/charges"
"2018/08/01 10:42:50 Requesting POST api.stripe.com/v1/charges"
```

One of the two requests ended with a 409 and was retried and failed with the error in the title. As it does not appear on the dashboard, it seems it did not reach your servers.

AFAIK, it should have been retried.

Similar but maybe not related to #528 

**go version: 1.10.1
stripe-go: 35.1.0**

Thanks!

yns01:
I suspect that `<` is the start of `<nil>`

brandur-stripe:
@yns01 You do know if the the decoding error occurred on the first request or the second? (It looks like the second I think?)

brandur-stripe:
@yns01 I went through and audited all the retry code here, but it's still not clear to me exactly where your problem is occurring.

It looks like what happened here is that you had a first requested which 409ed due to a conflicting request already in flight. The code retried and you got back what looks like a valid response, but decoding failed somewhere (I believe the `invalid character ...` was thrown by an `json.Unmarshal`).

Can you do two things:

1. It looks like you truncated the log like `Stripe Response: \"{\\\\"id\\\": \\\"ch_*\\\",\`. Could you examine the entire thing and make sure there's nothing fishy in there that would lead to an unmarshaling error?
[...]

PR #643, opened 2018-08-01
Bug fix + various code/logging improvements to retry code

brandur:
This one is a variety of improvements to the retry code:

* Fixed one bug whereby for a failed request, we were not reading the
  entire response body before trying to log it, which results in an
  uninformative output like `&{409....`. Here we add a `ReadAll`.
* I denested some code so that we have less indentation.
* I changed measuring request time to measuring individual requests
  around `s.HTTPClient.Do` instead of measuring the entire span
  including all the sleeping the retry block is doing.
* I added logging for the amount of time the program is about to go to
  sleep before a retry. You can still get the total execution time by
  adding the time for all requests plus the time for all sleeps.
* Generally changed some logs around for consistency around format,
  casing, and punctuation.

The impetus is that I was trying to understand the code to debug #642.
This might help get some visibility into a solution, but won't solve the
problem.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -283,34 +283,51 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
 		s.Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
 	}
 
-	start := time.Now()
-
 	var res *http.Response
 	var err error
 	for retry := 0; ; {
+		start := time.Now()
+
 		res, err = s.HTTPClient.Do(req)
-		if s.shouldRetry(err, res, retry) {
-			if s.LogLevel > 0 {
-				s.Logger.Printf("Request failed with error: %v. Response: %v\n", err, res)
-			}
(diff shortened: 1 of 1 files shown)


stripe.go
```diff
@@ -329,7 +346,7 @@ func (s *BackendConfiguration) Do(req *http.Request, v interface{}) error {
 	}
 
 	if s.LogLevel > 2 {
-		s.Logger.Printf("Stripe Response: %q\n", resBody)
+		s.Logger.Printf("Response: %s\n", string(resBody))
```
brandur-stripe: Changed the `%q` to a `%s` with a `string(...)` cast because the `%q` produces lots of noisy escaping that's not really needed for anything and detrimental to legibility:

``` go
Stripe Response: \"{\\\\"id\\\": \\\"ch_*\\\",\
```

brandur-stripe:
Thanks Remi!

brandur-stripe:
Released as 38.1.0.

PR #644, opened 2018-08-02
Add the File Link resource

remi-stripe:
r? @ob-stripe @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/filelink.go
+++ b/filelink.go
@@ -1,0 +1,60 @@
+package stripe
+
+import (
+	"encoding/json"
+)
+
+// FileLinkParams is the set of parameters that can be used when creating or updating a file link.
+type FileLinkParams struct {
+	Params    `form:"*"`
+	ExpiresAt *int64  `form:"expires_at"`
+	File      *string `form:"file"`
+}
+
+// FileLinkListParams is the set of parameters that can be used when listing file links.
+type FileLinkListParams struct {
+	ListParams   `form:"*"`
+	Created      *int64            `form:"created"`
+	CreatedRange *RangeQueryParams `form:"created"`
+	Expired      *bool             `form:"expired"`
+	File         *string           `form:"file"`
+}
+
+// FileLink is the resource representing a Stripe file link.
+// For more details see https://stripe.com/docs/api#file_links.
+type FileLink struct {
+	Created   int64             `json:"created"`
+	Expired   bool              `json:"expired"`
+	ExpiresAt int64             `json:"expires_at"`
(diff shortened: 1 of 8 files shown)


filelink.go
```diff
@@ -0,0 +1,60 @@
...
+
+	type fileLink FileLink
+	var v fileLink
+	if err := json.Unmarshal(data, &v); err != nil {
+		return err
+	}
+
+	*c = FileLink(v)
+	return nil
+}
```
brandur-stripe: Do we need the custom unmarshal here? (i.e., Is file link expanded in any other class?)

brandur-stripe:
Minor comment, but otherwise looks great! Thanks for the fast turnaround on all of these Remi.

ptal @remi-stripe

filelink.go
remi-stripe: @brandur-stripe While it's not the case yet it likely will be at some point when we ship more resources. I can imagine this being included in another resource and since I don't think we would have tests for this we would not catch the fact that the deserialization is missing right?

Deferring to you on whether it's better to add it in the future!

ob-stripe: Actually, `FileLink` is expanded as a list in the `links` attribute of `FileUpload` objects.

brandur-stripe: Okay, fair enough! Do you mind if we add a `filelink_test.go` to test this custom unmarshal logic then? You can copy the tests from dispute almost exactly:

``` go
package stripe

import (
	"encoding/json"
	"testing"

	assert "github.com/stretchr/testify/require"
)

func TestDispute_UnmarshalJSON(t *testing.T) {
	// Unmarshals from a JSON string
	{
		var v Dispute
		err := json.Unmarshal([]byte(`"dp_123"`), &v)
		assert.NoError(t, err)
		assert.Equal(t, "dp_123", v.ID)
	}

	// Unmarshals from a JSON object
	{
		v := Dispute{ID: "dp_123"}
		data, err := json.Marshal(&v)
		assert.NoError(t, err)

		err = json.Unmarshal(data, &v)
		assert.NoError(t, err)
		assert.Equal(t, "dp_123", v.ID)
	}
}
```

remi-stripe:
Okay added the test and also added the changes to the FileUpload resource to properly have a collection of FileLink.
PTAL

brandur-stripe:
> Okay added the test and also added the changes to the FileUpload resource to properly have a collection of FileLink.

Bonus! Thanks Remi!

brandur-stripe:
Released as 38.2.0.

PR #646, opened 2018-08-04
Set request body before every retry

brandur:
This patch changes the interfaces of `NewRequest` and `Do` around a
little so that we can set a new request body with every request.

In the era of HTTP (1), it was safe to reuse a `Request` object, but
with the addition of HTTP/2, it's now only sometimes safe. Reusing a
`Request` with a body will break.

See some more information here:

https://github.com/golang/go/issues/19653#issuecomment-341539160

Fixes https://github.com/stripe/stripe-go/issues/642.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -184,12 +184,12 @@ func (s *BackendConfiguration) Call(method, path, key string, params ParamsConta
 func (s *BackendConfiguration) CallMultipart(method, path, key, boundary string, body io.Reader, params *Params, v interface{}) error {
 	contentType := "multipart/form-data; boundary=" + boundary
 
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Actually, I just realized that I think I can do this with less interface churn. Let me take one more shot at this ...

brandur-stripe:
Okay, I think there's a race condition in one of our tests (arg, had to restart one of the builds), but I'll fix that separately. For now, let's try to get this patch out.

r? @remi-stripe

brandur-stripe:
NP. Thanks Remi!

PR #648, opened 2018-08-06
Introduce buffers so a request's body can be read multiple times

### LINKED ISSUE #647: On retries, body is empty
Following PR #646, the body of POST requests on retries is empty:

You can still use https://github.com/yns01/stripe-go-invalid-char to reproduce:
```
2018/08/06 08:51:18 Requesting POST api.stripe.com/v1/charges
2018/08/06 08:51:18 Requesting POST api.stripe.com/v1/charges
2018/08/06 08:51:19 Request completed in 1.459352076s (retry: 0)
2018/08/06 08:51:19 Request failed with: {"error":{"code":"idempotency_key_in_use","doc_url":"https://stripe.com/docs/error-codes/idempotency-key-in-use","message":"There is currently another in-progress request using this Idempotent Key (that probably means you submitted twice, and the other request is still going through): 770434815mw69wd8iuv. Please try again later.","type":"invalid_request_error"}} (error: <nil>)
2018/08/06 08:51:19 Initiating retry 1 for request POST api.stripe.com/v1/charges after sleeping 500ms
2018/08/06 08:51:20 Request completed in 476.943826ms (retry: 1)
2018/08/06 08:51:20 Request failed with: {"error":{"code":"idempotency_key_in_use","doc_url":"https://stripe.com/docs/error-codes/idempotency-key-in-use","message":"There is currently another in-progress request using this Idempotent Key (that probably means you submitted twice, and the other request is still going through): 770434815mw69wd8iuv. Please try again later.","type":"invalid_request_error"}} (error: <nil>)
2018/08/06 08:51:20 Initiating retry 2 for request POST api.stripe.com/v1/charges after sleeping 917.846449ms
2018/08/06 08:51:21 Request completed in 3.364051524s (retry: 0)
2018/08/06 08:51:21 Response: {
  "id": "ch_1Cw2YkC1oEzKtlmeIfcTHGVz",
Stripe charge OK ch_1Cw2YkC1oEzKtlmeIfcTHGVz
2018/08/06 08:51:22 Request completed in 309.157064ms (retry: 2)
2018/08/06 08:51:22 Error encountered from Stripe: {"status":400,"message":"Keys for idempotent requests can only be used with the same parameters they were first used with. Try using a key other than '770434815mw69wd8iuv' if you meant to execute a different request.","request_id":"req_wDGUanPuWf9gxm","type":"idempotency_error"}
Stripe charge failed {"status":400,"message":"Keys for idempotent requests can only be used with the same parameters they were first used with. Try using a key other than '770434815mw69wd8iuv' if you meant to execute a different request.","request_id":"req_wDGUanPuWf9gxm","type":"idempotency_error"}
```

Checking the Stripe dashboard, we can see that both the first retry (ended with a 409) and the second (ended with 400) have empty POST bodies

stripe-go: 39.0.0

brandur:
In #646, I attempted to fix a problem whereby when using HTTP/2 we were
sending invalid requests during automatic retries because a `Request`
struct with a body cannot be reused in the context of HTTP/2.

Unfortunately, it wasn't blowing up quite as spectacularly as before,
the fix didn't quite work either. It would set a new `Body` before
retries, but that body would be pointing to an `io.Reader` that was
already exhausted as it had been fully consumed during the original
request, thereby producing an empty request body.

Here we modify the `Do` and `CallMultipart` interfaces so that they take
body buffers instead of just readers. When making a request, we create a
new reader for those buffers every time, thus ensuring a fresh one.

This is relatively trivial when making non-multipart requests (which is
most requests) because we were already producing a buffer in almost the
right place. It's a little more complicated for multipart requests (file
uploads) because we had encoding scheme that wasn't quite compatible,
but I've done some refactoring there (and added more tests) to bring
things in line.

We also add a much improved test framework this time around that
verifies that the problem is fixed and will definitively stay fixed.

Fixes #647.

r? @remi-stripe (Sorry, I know these are annoying to review, but LMK if you
have any questions.)
cc @stripe/api-libraries


stripe.go
```diff
@@ -487,6 +507,34 @@ func (s *BackendConfiguration) shouldRetry(err error, resp *http.Response, numRe
 	return false
 }
 
+// sleepTime calculates sleeping/delay time in milliseconds between failure and a new one request.
+func (s *BackendConfiguration) sleepTime(numRetries int) time.Duration {
```
brandur-stripe: This moved to a function the struct so that we can check `s.networkRetriesSleep`.

PR #649, opened 2018-08-06
Rename `BackendConfiguration` to `BackendImplementation`

brandur:
I recently introduced `BackendConfig` to give users a way of fine-tuning
a backend that's created. Unfortunately, that produced a close name
collision with the existing `BackendConfiguration`, which isn't really a
configuration, but more of the main implementation of the `Backend`
interface.

Here we rename `BackendConfiguration` to `BackendImplementation` to make
this less confusing. It's a breaking change, but I have to do a major
release as part of #648 anyway, so it's not a bad time to do it.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -140,12 +140,12 @@ type BackendConfig struct {
 	URL string
 }
 
-// BackendConfiguration is the internal implementation for making HTTP calls to
-// Stripe.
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Released as 40.0.0.

PR #652, opened 2018-08-07
Change the type of FileUpload's Links fld to FileLinkList

spastorelli-stripe:
The `Links` field of `FileUpload` struct is currently set to `[]*FileLink`.
This can cause an unmarshalling error when calling for example the Create FileUpload method, as the  `FileUpload` object return as response can include a list of file links:
```
{
  id: "file_xxxxxxxx",
  object: "file_upload",
  created: 1533640021,
  filename: "file.jpg",
  links: {
    object: "list",
    data: [ ... ],
    has_more: false,
    url: "/v1/file_links?file=file_xxxxxxxx"
  },
  ....
}
```
This PR attempts to fix this issue by changing the type of FileUpload's Links fld to `FileLinkList`.

r? @ob-stripe @brandur-stripe
cc @stripe/api-libraries

### DIFF
--- a/fileupload.go
+++ b/fileupload.go
@@ -47,7 +47,7 @@ type FileUpload struct {
 	Created  int64             `json:"created"`
 	ID       string            `json:"id"`
 	Filename string            `json:"filename"`
-	Links    []*FileLink       `json:"links"`
+	Links    *FileLinkList     `json:"links"`
 	Purpose  FileUploadPurpose `json:"purpose"`
 	Size     int64             `json:"size"`
 	Type     string            `json:"type"`


filelink.go
```diff
@@ -53,7 +53,7 @@ func (c *FileLink) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// FileLinkList is a list of file links as retrieved from a list endpoint.
+// FileLinkList is a list object for file links.
```
remi-stripe: I think we should keep that comment since we tried to make it consistent across packages

fileupload.go
```diff
@@ -47,7 +47,7 @@ type FileUpload struct {
 	Created  int64             `json:"created"`
 	ID       string            `json:"id"`
 	Filename string            `json:"filename"`
-	Links    []*FileLink       `json:"links"`
+	Links    *FileLinkList     `json:"links"`
```
remi-stripe: Nice catch, totally missed that one when I shipped the FileLink resource

brandur-stripe:
Geeze, I should've caught that one on review too. Thanks @spastorelli-stripe!

Let me know if you've taken a look at Remi's comment, but will get this released ASAP after.

filelink.go
spastorelli-stripe: Makes sense. Reverted to the original comment.

spastorelli-stripe:
Thanks @remi-stripe and @brandur-stripe for your reviews. I believe I have addressed Remi's comment. PTAL

brandur-stripe:
Thanks @spastorelli-stripe.

Even though this is an interface change, I'm going to release this as a patch because I don't think that the list of file links could ever have successfully deserialized into `[]*FileLink`.

fileupload.go
spastorelli-stripe: I can't get the credit on this one as I didn't catch this. I only came across as it was reported in a ticket.

brandur-stripe:
Released as 40.0.2.

PR #656, opened 2018-08-10
Split webhook payload validation logic into separate ValidatePayload* functions

jfheller:
ConstructEvent and related functions now use the new validatePayload functionality.

This allows consumers of the library to not have to Unmarshal the json
payload in order to validate it. This way a library user can choose to only
validate a signature of a message and pass it to downstream systems as
trusted without being concerned with the actual structure of the events
yet. This is useful because the Event structure is tied to the user's
Stripe-account-wide API version which must be upgraded in an all-or-nothing
fashion. With this commit a library user may choose to write the webhook
event validation portion of their code in a way that will not be broken by
forwards or backwards- incompatible event structure changes

### DIFF
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -89,6 +89,68 @@ func parseSignatureHeader(header string) (*signedHeader, error) {
 	return sh, nil
 }
 
+// ValidatePayload validates the payload against the Stripe-Signature header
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks!

brandur-stripe:
Released as 40.1.0.

PR #657, opened 2018-08-14
Integer-index encode all arrays

brandur:
Changes all arrays from classic Rack encoding:

``` sh
arr[]=...&arr[]=...&arr[]=...
```

To integer-indexed encoding:

``` sh
arr[0]=...&arr[1]=...&arr[2]=...
```

See some additional background in: https://github.com/stripe/stripe-ruby/pull/674

cc @stevene-stripe @zanker-stripe

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -56,10 +56,15 @@ func TestAccountNew(t *testing.T) {
 			AdditionalOwners: []*stripe.AdditionalOwnerParams{
 				{
 					FirstName: stripe.String("Jane"),
+					LastName:  stripe.String("Doe"),
 					Verification: &stripe.IdentityVerificationParams{
 						Document:     stripe.String("file_345"),
 						DocumentBack: stripe.String("file_567"),
 					},
+				},
+				{
+					FirstName: stripe.String("John"),
+					LastName:  stripe.String("Doe"),
 				},
 			},
 			DOB: &stripe.DOBParams{
--- a/form/form.go
+++ b/form/form.go
@@ -49,11 +49,6 @@ type field struct {
 }
 
 type formOptions struct {
-	// IndexedArray indicates that contrary to standard "Rack-style" form
-	// encoding, array items should be index like `arr[0]=...&arr[1]=...`
-	// (normally it'd be `arr[]=...`).
-	IndexedArray bool
-
 	// Empty indicates that a field's value should be emptied in that its value
 	// should be an empty string. It's used to workaround the fact that an
 	// empty string is a string's zero value and wouldn't normally be encoded.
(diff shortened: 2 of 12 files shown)


account/client_test.go
```diff
@@ -56,11 +56,16 @@ func TestAccountNew(t *testing.T) {
...
+					LastName:  stripe.String("Doe"),
 					Verification: &stripe.IdentityVerificationParams{
 						Document:     stripe.String("file_345"),
 						DocumentBack: stripe.String("file_567"),
 					},
 				},
+				{
+					FirstName: stripe.String("John"),
+					LastName:  stripe.String("Doe"),
+				},
```
brandur-stripe: Added a little more flavor to this test just so that I could see multiple array elements being encoded for myself.

form/form.go
```diff
@@ -454,12 +442,6 @@ func parseTag(tag string) (string, *formOptions) {
 			}
 			options.Empty = true
 
-		case "indexed":
```
brandur-stripe: Removed the `indexed` option because everything is indexed now.

form/form_test.go
```diff
@@ -295,44 +289,6 @@ func TestAppendTo(t *testing.T) {
...
-			values := form.ToValues()
-			//fmt.Printf("values = %+v", values)
-
-			// This is the only difference between this test case and the one
-			// above: we used square brackets to grab a []string instead of
-			// just a single value.
-			assert.Equal(t, tc.want, values[tc.field])
-		})
-	}
-}
```
brandur-stripe: I removed this test because it's largely here to make sure that duplicate keys are tolerated — duplicate keys like you'd see for an array encoded without integer indexes. Now that we do have integer indexes everywhere, this functionality is no longer important.

form/form.go
remi-stripe: Flagging that we likely could/should remove all options. The only one really used is `empty` specifically for additional_owners I think.

params_test.go
```diff
@@ -196,7 +196,7 @@ func TestListParams_Expand(t *testing.T) {
 		{
 			InitialBody:  [][2]string{{"foo", "bar"}, {"foo", "baz"}},
 			Expand:       []string{"data", "data.foo"},
-			ExpectedBody: [][2]string{{"foo", "bar"}, {"foo", "baz"}, {"expand[]", "data"}, {"expand[]", "data.foo"}},
+			ExpectedBody: [][2]string{{"foo", "bar"}, {"foo", "baz"}, {"expand[0]", "data"}, {"expand[1]", "data.foo"}},
```
remi-stripe: Oh that's neat, I was worried that expand would for some reason work differently. Glad to hear it does not.

sub.go
```diff
@@ -42,7 +42,7 @@ type SubscriptionParams struct {
 	Coupon                      *string                    `form:"coupon"`
 	Customer                    *string                    `form:"customer"`
 	DaysUntilDue                *int64                     `form:"days_until_due"`
-	Items                       []*SubscriptionItemsParams `form:"items,indexed"`
+	Items                       []*SubscriptionItemsParams `form:"items"`
```
remi-stripe: Minor but is it worth making sure all those arrays have an explicit test for the multiple element one? I checked and for example the subscription test only adds one item in the `items` hash.

form/form.go
brandur-stripe: +1. Let's do that as a separate cleanup, but yep, we should be able to kill all of them at this point.

sub.go
brandur-stripe: Yeah, can't hurt. I added multi-item tests for all of order, order return, plans, and subscriptions.

brandur-stripe:
Released as 40.2.0.

PR #658, opened 2018-08-15
Remove the form package's `zero` option

brandur:
Removes the form package's `zero` option, which was previously used to
encode a true zero for an integer value before we introduced pointers
everywhere.

I also looked into removing the `empty` option, but I think we still
need it for now. We could move it into a custom `AppendTo`
implementation specifically on `LegalEntityParams`, but it doesn't seem
like a clear win to me.

r? @remi-stripe (As we discussed a little yesterday.)

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -53,12 +53,6 @@ type formOptions struct {
 	// should be an empty string. It's used to workaround the fact that an
 	// empty string is a string's zero value and wouldn't normally be encoded.
 	Empty bool
-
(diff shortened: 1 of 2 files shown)


remi-stripe:
@brandur-stripe seems my review never made it through, really sorry for the delays :(

LGTM

brandur:
> @brandur-stripe seems my review never made it through, really sorry for the delays :(

No worries! This one is super low priority.

PR #661, opened 2018-08-17
Fix type error in issuing_card.go

jack-stripe:
`eta` is a timestamp, like `created`.


brandur-stripe:
Oops, yep, that looks more right. Thanks Jack!

brandur-stripe:
Released as 41.0.0.

PR #662, opened 2018-08-17
Rename `IssuingCardShipping`'s `Eta` to `ETA` to fit conventions

brandur:
Convention throughout Go and this library is to upcase field names that
are acronyms. Here we do that for "ETA". We are changing its type anyway
in #661, so it's a good time to make this change.


brandur-stripe:
Going to self-approve this one because it's very minor:

LGTM.

brandur-stripe:
Released as 41.0.0.

PR #663, opened 2018-08-20
Add support for usage record summary

remi-stripe:
cc @stripe/api-libraries


remi-stripe:
Okay confirmed the resources can't be expanded (invoice and subscription item) so ready to review.

r? @brandur-stripe

usage_record_summary.go
```diff
@@ -0,0 +1,25 @@
+package stripe
+
+// UsageRecordSummary represents a usage record summary.
+// See https://stripe.com/docs/api#usage_records
+type UsageRecordSummary struct {
+	ID               string  `json:"id"`
+	Live             bool    `json:"livemode"`
```
brandur-stripe: I think we agreed on the fully expanded `Livemode` for these types of properties, but doh! I just noticed when grepping that we missed a couple places — usage records, issuer fraud records, and source transactions.

I'm not sure what to do with those ones right now, but can we make this one `Livemode`?

usage_record_summary.go
```diff
@@ -0,0 +1,25 @@
...
+	Invoice          string  `json:"invoice"`
```
brandur-stripe: Also, I think for most structs we went with pure alphabetical order — `ID` and `Livemode` don't get special treatment.

Again, it looks like usage records are a little weird in this respect.

usagerecordsummary/client.go
```diff
@@ -0,0 +1,50 @@
...
+package usagerecordsummary
+
+import (
+	"net/http"
+
+	stripe "github.com/stripe/stripe-go"
+	"github.com/stripe/stripe-go/form"
+)
+
+// Client is used to invoke /plans APIs.
```
brandur-stripe: I think for most of these we went with something like:

``` go
// Client is used to invoke APIs related to usage record summaries.
```

usage_record_summary.go
remi-stripe: Damn nice catch, I copy/pasted from usagerecord and when I double checked I saw other matches and did not look further

remi-stripe:
@brandur-stripe  Nice catch on all this. I took this as an opportunity to fix all `Live` and fix the naming of the files to remove the `_` which kind of indicates a namespace in theory (in our API).

PTAL

brandur-stripe:
LGTM. Thanks!

brandur-stripe:
Released as 42.0.0.

PR #667, opened 2018-08-23
Add support for the forgive parameter on the Pay Invoice API

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM.

Side note: the API reference docs are kind of fail for this parameter. It doesn't actually say anywhere that it expects a boolean and you should pass `true` or `false`. Users kind of just need to figure that out for themselves based on the naming.

brandur-stripe:
Released as 42.3.0.

PR #669, opened 2018-08-28
Add support for authorization_code on the Charge resource

remi-stripe:
Note that this property is internal and not usually returned in the API.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
> Note that this property is internal and not usually returned in the API.

Thanks for the heads up.

LGTM.

brandur-stripe:
Released as 43.1.0.

PR #673, opened 2018-08-28
Remove racy test in favor of simpler one

brandur:
Removes a test that's producing racy results in CI. As noted in #672, it
probably became racy as it started using the new HTTP implementation
in Go's HTTP/2 package.

My justification here is that cancellable contexts are already well
tested in Go's core. What we really want to test is just that the
context supplied by our users does indeed make it into the Go `Request`
that we're about to execute -- from there we can expect Go to do the
right thing. We already have a test that checks that just a few lines
above.

r? @remi-stripe What do you think about this approach?
cc @stripe-internal/api-libraries

### DIFF
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -33,38 +33,10 @@ func TestContext(t *testing.T) {
 	req, err := c.NewRequest("", "", "", "", p)
 	assert.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks Remi!

PR #674, opened 2018-08-28
Fix payment intent parameters

remi-stripe:
* Capture takes a different set of parameters
* application_fee is now called application_fee_amount

r? @brandur-stripe 
cc @stripe/api-libraries 

(Flagging the latest stripe-mock broke the test suite so also going to fix that in a separate PR in a few minutes)

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -44,6 +44,14 @@ const (
 	PaymentIntentStatusSucceeded            PaymentIntentStatus = "succeeded"
 )
 
+// PaymentIntentCaptureParams is the set of parameters that can be used when capturing a payment intent.
+type PaymentIntentCaptureParams struct {
+	Params               `form:"*"`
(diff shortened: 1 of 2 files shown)


paymentintent.go
```diff
@@ -44,6 +44,14 @@ const (
 	PaymentIntentStatusSucceeded            PaymentIntentStatus = "succeeded"
 )
 
+// PaymentIntentCaptureParams is the set of parameters that can be used when capturing a payment intent.
+type PaymentIntentCaptureParams struct {
```
remi-stripe: Is this a better approach than using the common one and adding the custom parameter `AmountToCapture` at the bottom like we do for Pay Invoice?

brandur-stripe: I'm honestly not sure, but I have a slight preference for this way I think. Let's try it for now.

PR #675, opened 2018-08-28
Fix two issues with List tests

remi-stripe:
* Ensure that `subscription_item` is not sent as a parameter when listing usage record summaries
* Update the test for usage record summary
* Fix the bank account list test for Accounts to pass the right parameter

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/bankaccount/client_test.go
+++ b/bankaccount/client_test.go
@@ -37,7 +37,7 @@ func TestBankAccountGet_ByCustomer(t *testing.T) {
 }
 
 func TestBankAccountList_ByAccount(t *testing.T) {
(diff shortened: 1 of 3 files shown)


brandur-stripe:
LGTM.

> Ensure that subscription_item is not sent as a parameter when listing usage record summaries

Just for release purposes: was this wrong, or was it causing a problem?

remi-stripe:
Fairly sure it was wrong and the API would have returned an error so it's likely no one used it yet. The issue is that stripe-mock did not validate extra parameters on GET requests so I did not catch it.

brandur-stripe:
Ah, I see! Okay, thanks.

PR #677, opened 2018-08-30
Remove the right test

brandur:
In #676 I accidentally removed the wrong test — I was trying to get the
"list" test for accounts, not the "get" test for accounts.

Here we resurrect the test that was removed accidentally and remove the
right one instead. Oops.

cc @remi-stripe


brandur-stripe:
> damn sorry I should have caught this

You and me both :)

PR #678, opened 2018-08-30
Fix subscription item listing test

brandur:
When listing subscription items a subscription ID is required. Our test
wasn't previously doing that, so here we add one in.

Caught while testing against a newer version of stripe-mock that
validates query parameters.


brandur-stripe:
Thanks Remi!

PR #680, opened 2018-08-30
SubscriptionTaxPercent should be a float64, not int64

EVODelavega:
The TaxPercent field is a *float64, the SubscriptionTaxPercent field should match, if not it's impossible to apply something like 6.5% tax on a subscription.


remi-stripe:
cc @brandur-stripe for release

This one might be a breaking change because it kind of works for people today that have tax percent as an int.

brandur-stripe:
Thanks everyone!

> This one might be a breaking change because it kind of works for people today that have tax percent as an int.

+1. I'll do a major.

brandur-stripe:
Released as 45.0.0.

Issue #682, opened 2018-08-30
support for order.List with status_transitions filters


rothskeller:
The API documentation for listing orders says that it supports a status_transition parameter, but the Go language bindings don't seem to have any way to set this parameter.  Could this be added?

(Actually what I'd really like is a filter on "updated", so that I could efficiently poll for order changes, but that would be an API change as well as a language binding change.)

PR #683, opened 2018-08-31
Add status_transitions filters to the List Orders params

spastorelli-stripe:
Add `status_transitions` filters to the List Orders params.
Attempts to fix https://github.com/stripe/stripe-go/issues/682  

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/order.go
+++ b/order.go
@@ -139,12 +139,25 @@ type OrderList struct {
 
 // OrderListParams is the set of parameters that can be used when listing orders.
 type OrderListParams struct {
-	ListParams   `form:"*"`
-	Created      *int64            `form:"created"`
-	CreatedRange *RangeQueryParams `form:"created"`
-	Customer     *string           `form:"customer"`
-	IDs          []*string         `form:"ids"`
-	Status       *string           `form:"status"`
+	ListParams        `form:"*"`
+	Created           *int64                         `form:"created"`
+	CreatedRange      *RangeQueryParams              `form:"created"`
+	Customer          *string                        `form:"customer"`
+	IDs               []*string                      `form:"ids"`
+	Status            *string                        `form:"status"`
+	StatusTransitions *StatusTransitionsFilterParams `form:"status_transitions"`
+}
+
+// StatusTransitionsFilterParams are parameters that can used to filter on status_transition when listing orders.
+type StatusTransitionsFilterParams struct {
+	Canceled       *int64            `form:"canceled"`
+	CanceledRange  *RangeQueryParams `form:"canceled"`
+	Fulfilled      *int64            `form:"fulfilled"`
+	FulfilledRange *RangeQueryParams `form:"fulfilled"`
+	Paid           *int64            `form:"paid"`
+	PaidRange      *RangeQueryParams `form:"paid"`
+	Returned       *int64            `form:"returned"`
+	ReturnedRange  *RangeQueryParams `form:"returned"`
 }
 
 // StatusTransitions are the timestamps at which the order status was updated.


order.go
```diff
@@ -139,20 +139,30 @@ type OrderList struct {
...
+	Created           *int64                         `form:"created"`
+	CreatedRange      *RangeQueryParams              `form:"created"`
+	Customer          *string                        `form:"customer"`
+	IDs               []*string                      `form:"ids"`
+	Status            *string                        `form:"status"`
+	StatusTransitions *StatusTransitionsFilterParams `form:"status_transitions"`
+}
+
+// StatusTransitionsFilterParams are parameters that can used to filter on status_transition when listing orders.
+type StatusTransitionsFilterParams struct {
```
brandur-stripe: What do you think about calling this `StatusTransitionsParams` for a little more brevity? Also cc @remi-stripe in case you want in on the naming bikeshed ;)

order.go
```diff
@@ -139,20 +139,30 @@ type OrderList struct {
...
+	StatusTransitions `form:"*"`
+	CanceledRange     *RangeQueryParams `form:"canceled"`
```
brandur-stripe: Could you also add a field for each of `Canceled`, `Fulfilled`, `Paid`, and `Returned` which is a `*int64`? This is how we generally support either a range or a basic Unix timestamp like for `Created`. From the docs:

> A filter on the list based on the object canceled field. The value can be a string with an integer Unix timestamp, or it can be a dictionary with the following options:

order.go
```diff
@@ -139,20 +139,30 @@ type OrderList struct {
...
 // StatusTransitions are the timestamps at which the order status was updated.
 type StatusTransitions struct {
-	Canceled  int64 `json:"canceled"`
-	Fulfilled int64 `json:"fulfiled"`
-	Paid      int64 `json:"paid"`
-	Returned  int64 `json:"returned"`
+	Canceled  int64 `json:"canceled" form:"canceled"`
+	Fulfilled int64 `json:"fulfiled" form:"fulfilled"`
+	Paid      int64 `json:"paid" form:"paid"`
+	Returned  int64 `json:"returned" form:"returned"`
```
brandur-stripe: And let's drop the `form` annotations from this struct — they're not needed.

order/client_test.go
```diff
@@ -19,7 +19,13 @@ func TestOrderGet(t *testing.T) {
 }
 
 func TestOrderList(t *testing.T) {
-	i := List(&stripe.OrderListParams{})
+	i := List(&stripe.OrderListParams{
+		StatusTransitions: &stripe.StatusTransitionsFilterParams{
+			PaidRange: &stripe.RangeQueryParams{
+				GreaterThan: int64(123456),
+			},
+		},
```
brandur-stripe: I don't think adding the new predicate is strictly needed here because both the `form` annotations and `RangeQueryParams` serialization are pretty well tested already (see `form/form_test.go` and `params_test.go`).

order.go
remi-stripe: The issue with `StatusTransitionsParams` is that if we ever added `status_transitions` as a parameter on the create/update then there's a naming conflict. So either `StatusTransitionsFilterParams` or `StatusTransitionsListParams` is likely safer

brandur-stripe: Ah true! Okay, I'm okay with just leaving as is then. Thanks.

order.go
spastorelli-stripe: I had implicitly provided these fields by embedding `StatusTransitions` in `StatusTransitionsFilterParams`.
Should I instead explicitly include each of the fields in `StatusTransitionsFilterParams`?

order/client_test.go
spastorelli-stripe: Makes sense. Removed.

order.go
spastorelli-stripe: Done

order.go
spastorelli-stripe: Leaving it like this.

spastorelli-stripe:
Thanks @brandur-stripe for the review. PTAL

order.go
spastorelli-stripe: Thanks. Done.

spastorelli-stripe:
Thanks @brandur-stripe for the follow up. PTAL

order.go
```diff
@@ -139,12 +139,25 @@ type OrderList struct {
...
+	IDs               []*string                      `form:"ids"`
+	Status            *string                        `form:"status"`
+	StatusTransitions *StatusTransitionsFilterParams `form:"status_transitions"`
+}
+
+// StatusTransitionsFilterParams are parameters that can used to filter on status_transition when listing orders.
+type StatusTransitionsFilterParams struct {
+	Canceled       *int64            `form:"canceled"`
+	CanceledRange  *RangeQueryParams `form:"canceled"`
+	Fulfilled      *int64            `form:"fulfiled"`
```
brandur-stripe: Looks great!

One more very minor bug here: this form annotation should be "fulfilled" (with two Ls) like the one below it.

ptal @spastorelli-stripe

spastorelli-stripe: Thanks for catching this. Done.

spastorelli-stripe:
Thanks @brandur-stripe. PTAL

brandur-stripe:
LGTM. Thanks!

brandur-stripe:
Released as 48.1.0.

Issue #684, opened 2018-09-03
`SourceObjectParams` missing `receiver` field


ddosic:
Hello, 

I have tried using the `source.New`, but it seems that I cannot set the `receiver.refund_attributes_method` as declared in https://stripe.com/docs/api#create_source

Am I missing something in the stripe-go api, or should the `SourceObjectParams` be enriched with another field?

Thanks,
Dejan

remi-stripe:
@ddosic Thanks for the report! I just opened a PR https://github.com/stripe/stripe-go/pull/686 that will fix this and we will merge it later this week.

ddosic:
That is great to hear, thanks!

PR #686, opened 2018-09-03
Add support for `mandate` and `receiver` hashes on Source creation

remi-stripe:
This is a breaking change as `date` was incorrectly categorized as a String.

This fixes https://github.com/stripe/stripe-go/issues/684

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM.

brandur-stripe:
> This is a breaking change as `date` was incorrectly categorized as a String.

+1. Thanks!

Released as 46.0.0.

PR #688, opened 2018-09-04
Encode `Params` in `AppendToAsSourceOrExternalAccount`

### LINKED ISSUE #685: API.BankAccounts.New() doesn't add Metadata to created bank account.
Using go 1.11 and stripe-go v45.0.0.

```
bankAcct, err := s.stripe.BankAccounts.New(&stripe.BankAccountParams{
	Token:    &token.StripeBankAccountToken,
	Customer: &account.StripeCustomerID,
	Params: stripe.Params{
		Context: ctx,
		Metadata: map[string]string{
			"ABC": "123",
		},
	},
})
if err != nil {
	log.Printf("CreatePaymentMethod: create bank account: %v", err)
	return nil, err
}
log.Println(bankAcct.Metadata)
```
prints
```
2018/09/03 14:42:58 paymentmethod.go:215: map[]
```
My workaround is to update the bank account with metadata once it's been created, but it took me a while to figure out what was going on, and it would be nice to save others from the same trouble in the future.

brandur:
`BankAccount` has a special encoding function that's invoked from its
`New` called `AppendToAsSourceOrExternalAccount`. It works quite a bit
differently from any other resource (with the exception of card) because
this special encoding is invoked then the function uses a `CallRaw`
instead of a `Call`, so no general `AppendTo` is ever used.

As described in #685, this caused a problem whereby the standard set of
fields in `Params` were not being included because the function didn't
have any consideration for them (although strangely, we *were* handling
them from card's equivalent). Here we add an allowance for `Params` and
a test to make sure that metadata is being encoded correctly.

Fixes #685.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -71,6 +71,11 @@ type BankAccountParams struct {
 // because the bank accounts endpoint is a little unusual. There is one other
 // resource like it, which is cards.
 func (a *BankAccountParams) AppendToAsSourceOrExternalAccount(body *form.Values) {
+	// Rather than being called in addition to `AppendTo`, this function
+	// *replaces* `AppendTo`, so we must also make sure to handle the encoding
+	// of `Params` so metadata and the like is included in the encoded payload.
+	form.AppendTo(body, a.Params)
+
 	isCustomer := a.Customer != nil
 
 	var sourceType string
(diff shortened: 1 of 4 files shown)


brandur-stripe:
> Since the Card one works a bit differently it might be worth adding an extra tests but deferring to you on that

Good point! It turns out this had super minimal testing. Added a few more tests in ae63bc3, although they're by no means exhaustive.

remi-stripe:
Assuming tests still pass, LGTM. Thank you for the extra tests.

brandur-stripe:
NP! Thanks Remi.

brandur-stripe:
Released as 46.1.0.

PR #690, opened 2018-09-05
Add support for Reporting resources

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries @brahn-stripe

### DIFF
--- a/reporting_reportrun.go
+++ b/reporting_reportrun.go
@@ -1,0 +1,65 @@
+package stripe
+
+// ReportRunStatus is the possible values for status on a report run.
+type ReportRunStatus string
+
+// List of values that ReportRunStatus can take.
+const (
+	ReportRunStatusFailed    ReportRunStatus = "failed"
+	ReportRunStatusPending   ReportRunStatus = "pending"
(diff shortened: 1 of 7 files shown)


reporting_reportrun.go
```diff
@@ -0,0 +1,65 @@
...
+	ConnectedAccount  string   `json:"connected_account"`
+	Currency          Currency `json:"currency"`
+	IntervalEnd       int64    `json:"interval_end"`
+	IntervalStart     int64    `json:"interval_start"`
+	Payout            string   `json:"payout"`
+	ReportingCategory string   `json:"reporting_category"`
+}
+
+// ReportRun is the resource representing a report run.
+type ReportRun struct {
```
remi-stripe: I decided not to prefix everything with `Reporting` in this case because it felt redundant and unlikely we'd have a conflict in the future.

Let me know if you think it's cleaner to keep the prefix though like we do for Issuing

brandur-stripe: Yeah, makes sense in this situation.

brandur-stripe:
LGTM!

brandur-stripe:
Released as 48.2.0.

PR #695, opened 2018-09-17
Bindings for Terminal endpoints

daz-stripe:
- Connection Tokens
- Locations
- Readers


daz-stripe:
r? @remi-stripe
cc @stripe/api-libraries

Will hit rebuild once we release stripe-mocks! Still working on that

terminal_connectiontoken.go
```diff
@@ -0,0 +1,12 @@
...
+
+// TerminalConnectionTokenParams is the set of parameters that can be used when creating a terminal connection token.
+type TerminalConnectionTokenParams struct {
+	Params             `form:"*"`
+	OperatorAccount    *string         `form:"operator_account"`
+}
+
+// TerminalConnectionToken is the resource representing a Stripe terminal connection token.
+type TerminalConnectionToken struct {
+	ConnectionToken    string          `json:"connection_token"`
```
brandur-stripe: So a `ConnectionToken` comes back with a single field that's also called `ConnectionToken` and there's no `object` field?

This doesn't seem all that great from an API perspective because (1) activation token is both a resource and a field in a resource, and (2) it doesn't future proof things very well in case you want to add more fields to this resource in the future. Did you guys consider naming this `secret` after the field of that name in ephemeral keys?

terminal_reader.go
```diff
@@ -0,0 +1,47 @@
+package stripe
+
+// TerminalReaderRegisterParams is the set of parameters that can be used for creating a terminal reader.
+type TerminalReaderRegisterParams struct {
```
brandur-stripe: Convention throughout the library is to keep the parameters for "new" and "update" in the same struct — let's combine this with `TerminalReaderUpdateParams`.

brandur-stripe:
Left a few comments above! Two other things that apply across a number of files:

1. Can you alphabetize the fields in your structs? `Params` comes at the top, but everything else in there should be in order.
2. The build won't succeed until you run gofmt on your files to line everything up according to Go's stylistic conventions. You'll probably want something like this `gofmt -w -s ./...`.

ptal @daz-stripe

issuing_dispute.go
```diff
@@ -44,11 +44,11 @@ type IssuingDisputeEvidenceParams struct {
...
-	Params      `form:"*"`
-	Amount      *int64                        `form:"amount"`
-	Evidence    *IssuingDisputeEvidenceParams `form:"evidence"`
-	Reason      *string                       `form:"reason"`
-	Transaction *string                       `form:"transaction"`
+	Params              `form:"*"`
+	Amount              *int64                        `form:"amount"`
+	Evidence            *IssuingDisputeEvidenceParams `form:"evidence"`
+	Reason              *string                       `form:"reason"`
+	DisputedTransaction *string                       `form:"disputed_transaction"`
```
daz-stripe: This got changed in pay-sever so I'm just updating it here to pass the test

remi-stripe: This makes it a breaking change though so tagging as such!

terminal_connectiontoken.go
```diff
@@ -0,0 +1,13 @@
...
+// TerminalConnectionTokenParams is the set of parameters that can be used when creating a terminal connection token.
+type TerminalConnectionTokenParams struct {
+	Params          `form:"*"`
+	OperatorAccount *string `form:"operator_account"`
+}
+
+// TerminalConnectionToken is the resource representing a Stripe terminal connection token.
+type TerminalConnectionToken struct {
+	Object string `json:"object"`
+	Secret string `json:"connection_token"`
```
brandur-stripe: Can you change `json:"connection_token"` here to `json:"secret"`?

issuing_dispute.go
brandur-stripe: Thanks Remi!

brandur-stripe:
Added one more minor ask here, but otherwise LGTM!

terminal_connectiontoken.go
daz-stripe: Good catch! will do

daz-stripe:
Update :)

brandur-stripe:
Thanks! Released as 49.3.0.

brandur-stripe:
And ... I screwed up the major despite Remi's warning, haha. Going to pull this release as go with 50.0.0.

brandur-stripe:
Released as 50.0.0.

PR #698, opened 2018-09-26
Update for API version 2018-09-24

ob-stripe:
cc @stripe/api-libraries 

Update for API version 2018-09-24:
- `FileUpload` is renamed to `File` (and generally, all `FileUploadSomething` structs are renamed to `FileSomething`)
- change `uploads.stripe.com` to `files.stripe.com`
- fix file links to use `api.stripe.com`

Ideally, we'd only use `files.stripe.com` for the file creation request, and `api.stripe.com` for file retrieve / list requests, but the library's architecture makes that difficult. It's not a huge deal since we will likely keep supporting retrieve and list requests on `files.stripe.com` in perpetuity.


brandur-stripe:
LGTM as well!

PR #700, opened 2018-10-03
Add "OnBehalfOf" to subscriptions

klauspost:
It seems like `OnBehalfOf` was added to SubscriptionParams but not the subscription objects themselves.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -155,6 +155,7 @@ type Charge struct {
 	Level3              ChargeLevel3        `json:"level3"`
 	Livemode            bool                `json:"livemode"`
 	Metadata            map[string]string   `json:"metadata"`
+	OnBehalfOf          *Account            `json:"on_behalf_of"`
 	Outcome             *ChargeOutcome      `json:"outcome"`
 	Paid                bool                `json:"paid"`
 	ReceiptEmail        string              `json:"receipt_email"`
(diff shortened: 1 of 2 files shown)


remi-stripe:
@klauspost Thank you for the pull request! Since this field represents an Account id (acct_123) it can be [Expanded](https://stripe.com/docs/api#expanding_objects) and return an Account object too. Would you be up for changing your PR to support this? You can see how we do this on the Charge resource [here](https://github.com/stripe/stripe-go/blob/master/charge.go#L148) for `destination`. This makes me realize we should also support `on_behalf_of` on the Charge resource too.

Let me know if it'd be easier for us to do the changes!

klauspost:
> @klauspost Thank you for the pull request! Since this field represents an Account id (acct_123) it can be [Expanded](https://stripe.com/docs/api#expanding_objects) and return an Account object too. 

@remi-stripe I don't really see how that works. Unless you always "expand", I can't really figure out how it falls back to the ID.

> Let me know if it'd be easier for us to do the changes!

Thanks - I would appreciate that since I don't really get how the expand functionality works.

remi-stripe:
@klauspost This idea is that the library will automatically deserialize the value, if it's a string it will only set `OnBehalfOf.Id` and if it's an object it will deserialize all the Account properties automatically. You only need to replace `string` by `*Account` in your PR and then update the tests.

Let me know if that clarifies how to do this. Otherwise I can take over in a separate PR instead to show you!

klauspost:
Ahh.. The `ParseID` in the Account unmarhal! 👍  I'll give it a shot :)

klauspost:
@remi-stripe Updated :)

remi-stripe:
@klauspost Released as 51.1.0. Thanks for working on this, that's awesome!

PR #702, opened 2018-10-09
Add support for the `delivery_success` param on the list event endpoint

karlr-stripe:
r? @remi-stripe 
cc @stripe/api-libraries


remi-stripe:
Released as 51.2.0

PR #706, opened 2018-10-17
Add support for the Webhook Endpoint resource

remi-stripe:
cc @stripe/api-libraries


remi-stripe:
> I think you need a separate struct for update parameters, with disabled and without connect.

That's not something we do in Go today. We don't have a separate struct for create and update as the vast majority of params are shared.

I added `disabled` though since you're correct that it was missing.

PR #707, opened 2018-10-18
Add new invoice methods and fixes to the Issuing Cardholder resource

remi-stripe:
This PR ships multiple breaking changes
* Move to `2018-11-08` as the new API version
* Updates the Invoice resource to support new properties, parameters and methods
* Add support for `default_source` on Subscription.

cc @stripe/api-libraries

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -27,9 +27,22 @@ type InvoiceBillingReason string
 const (
 	InvoiceBillingReasonManual             InvoiceBillingReason = "manual"
 	InvoiceBillingReasonSubscription       InvoiceBillingReason = "subscription"
+	InvoiceBillingReasonSubscriptionCreate InvoiceBillingReason = "subscription_create"
 	InvoiceBillingReasonSubscriptionCycle  InvoiceBillingReason = "subscription_cycle"
 	InvoiceBillingReasonSubscriptionUpdate InvoiceBillingReason = "subscription_update"
 	InvoiceBillingReasonUpcoming           InvoiceBillingReason = "upcoming"
+)
+
+// InvoiceBillingStatus is the reason why a given invoice was created
+type InvoiceBillingStatus string
+
+// List of values that InvoiceBillingStatus can take.
+const (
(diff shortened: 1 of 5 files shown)


remi-stripe:
Flagging that this is a breaking change and we'll need the new API version too before we can release.

invoice.go
```diff
@@ -47,11 +60,13 @@ type InvoiceUpcomingInvoiceItemParams struct {
 // For more details see https://stripe.com/docs/api#create_invoice, https://stripe.com/docs/api#update_invoice.
 type InvoiceParams struct {
 	Params              `form:"*"`
+	AutoAdvance         *bool    `form:"auto_advance"`
```
brandur-stripe: I think `auto_advance` is just a "finalize" parameter.

invoice.go
```diff
@@ -47,11 +60,13 @@ type InvoiceUpcomingInvoiceItemParams struct {
...
 	ApplicationFee      *int64   `form:"application_fee"`
 	Billing             *string  `form:"billing"`
 	Closed              *bool    `form:"closed"`
 	Customer            *string  `form:"customer"`
 	DaysUntilDue        *int64   `form:"days_until_due"`
+	DefaultSource       *string  `form:"default_source"`
```
brandur-stripe: I notice that `DefaultSource` on invoice and subscription are marked as expandable. Should we support that?

remi-stripe: Nice catch, fixed! PTAL

invoice.go
remi-stripe: Where did you see that one? `auto_advance` is the new world and has been planned for months. Just confirmed with the team internally that it's `auto_advance` and `finalize` does not exist as a parameter (but is a method)

brandur-stripe: (Spoke with Remi offline: this seems to be a documentation problem in that `auto_advance` isn't showing up in some places at the moment. Fine by me if we know it's supposed to be here.)

PR #709, opened 2018-10-19
Only run Golint on supported versions

brandur:
Unfortunately Go only supports the last two major versions of the
language for Golint, and aren't afraid of making changes that will break
older ones, which they seem to have just done (and thus breaking our
master build).

Here we track which versions of Go Golint can support, and only run it
if the build is happening on one of them.

[1] https://golang.org/doc/devel/release.html#policy


hone:
:+1: in bash we trust.

PR #711, opened 2018-10-29
Set `Request.GetBody`

### LINKED ISSUE #710: Closed connection with `define Request.GetBody` error message
Running a script to delete subscriptions in a loop, I get:

```
http2: server sent GOAWAY and closed the connection; LastStreamID=XXXX, ErrCode=NO_ERROR, debug=""
Delete https://api.stripe.com/v1/subscriptions/sub_XXXXXXXXXXXXXX: http2: Transport: cannot retry err [http2: Transport received Server's graceful shutdown GOAWAY] after Request.Body was written; define Request.GetBody to avoid this error
```

I guess I was kind of expecting the SDK to handle closed connections, but if that's not happening, is there a recommended way to capture these specific errors and reconnect?

brandur:
Here we set `Request.GetBody`, which is used under certain circumstances
by the internal `net/http` core like in the case of a redirect. This
will usually not end up being used, but it doesn't hurt to set it in
case it is (it's run lazily, and few additional resources are needed in
case it's not run).

Fixes #710.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,10 +52,10 @@ env:
     - STRIPE_MOCK_VERSION=0.33.0
 
 go:
-  - "1.7"
   - "1.8"
   - "1.9"
   - "1.10"
+  - "1.11"
   - tip
 
 language: go
--- a/stripe.go
+++ b/stripe.go
@@ -324,6 +324,18 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			reader := bytes.NewReader(body.Bytes())
 
 			req.Body = nopReadCloser{reader}
+
+			// And also add the same thing to `Request.GetBody`, which allows
+			// `net/http` to get a new body in cases like a redirect. This is
+			// usually not used, but it doesn't hurt to set it in case it's
+			// needed. See:
+			//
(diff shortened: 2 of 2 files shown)


remi-stripe:
@brandur-stripe looks like this does not work on Go 1.7

brandur-stripe:
@remi-stripe Doh! Yeah, I just saw that.

Any FUD for dropping support for <= 1.7? It's not officially supported by the Go core team, and hasn't been since the release of Go 1.9 in August 2017 (1+ year) according to their [version support policy](https://golang.org/doc/devel/release.html).

Also, looking at OB's language metrics page, <= 1.7 is roughly ~= 1.5% of our users, which is quite small. Those users will also likely be able to upgrade easily because there are roughly no breaking changes on the path to a more modern Go version.

remi-stripe:
@brandur-stripe Based on your stats that sounds fine to me (though it becomes a major release then)

brandur-stripe:
Thanks Remi. Alright, I just added 1.11 to the matrix and dropped 1.7. I'll release as a major.

brandur-stripe:
Released as 52.0.0.

PR #714, opened 2018-11-02
`Do` function panics when HTTPClient returns a nil response

### LINKED ISSUE #713: `Do` function panics when HTTPClient returns a nil response
Assume `res, err = s.HTTPClient.Do(req)` results in `res` that is `nil`. 
( https://github.com/stripe/stripe-go/blob/master/stripe.go#L341 )

*Current behaviour:*
The `Do` function tries to read `res.Body` which results in a panic "invalid memory address or nil pointer dereference".
( https://github.com/stripe/stripe-go/blob/master/stripe.go#L354 )

*Expected behaviour:*
The Do() function retries until success or MaxNetworkRetries threshold is met, or otherwise returns an error. 

---
go version go1.10.4 linux/amd64
stripe-go version 43.0.0

agatamigalska-newstore:
This PR covers the scenario when `res, err = s.HTTPClient.Do(req)` results in `res` that is `nil` (https://github.com/stripe/stripe-go/blob/master/stripe.go#L341 ). This can happen for example in case of timeout.

* a test was added to verify no panic in case of timeout
* in the `Do` function an additional check was added to prevent panic

Closes #713

---
[EDIT] Initial motivation:

Integration tests in the codebase I'm working on are randomly failing with the following stack trace:
```
error_cause=runtime error: invalid memory address or nil pointer dereference 
stack_trace=[
  {
  "function": "runtime.call32",
  "location": "/usr/local/go/src/runtime/asm_amd64.s:573"
  },
  {
  "function": "runtime.gopanic",
  "location": "/usr/local/go/src/runtime/panic.go:503"
  },
  {
  "function": "runtime.panicmem",
  "location": "/usr/local/go/src/runtime/panic.go:63"
  },
  {
  "function": "runtime.sigpanic",
  "location": "/usr/local/go/src/runtime/signal_unix.go:388"
  },
  {
  "function": "github.com/newstore/newstore/vendor/github.com/stripe/stripe-go.(*BackendImplementation).Do",
  "location": "/go/src/github.com/newstore/newstore/vendor/github.com/stripe/stripe-go/stripe.go:342"
  },
  {
  "function": "github.com/newstore/newstore/vendor/github.com/stripe/stripe-go.(*BackendImplementation).CallRaw",
  "location": "/go/src/github.com/newstore/newstore/vendor/github.com/stripe/stripe-go/stripe.go:224"
  },
  {
  "function": "github.com/newstore/newstore/vendor/github.com/stripe/stripe-go.(*BackendImplementation).Call",
  "location": "/go/src/github.com/newstore/newstore/vendor/github.com/stripe/stripe-go/stripe.go:186"
  },
  {
  "function": "github.com/newstore/newstore/vendor/github.com/stripe/stripe-go/customer.Client.Get",
  "location": "/go/src/github.com/newstore/newstore/vendor/github.com/stripe/stripe-go/customer/client.go:38"
  },
  ...
```

In this PR I'm trying to capture the situation in which this panic can happen in client.go:38. The best hypothesis was that res is nil, and so the attempt to access res.Body causes a panic. 

The goal in the test was to simulate the scenario where HttpClient.Do() function returns nil response. 

The timeout scenario for the test was chosen simply because, given the infrastructure used for testing stripe-go, it was the easiest to simulate.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -362,10 +362,10 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 				if s.LogLevel > 0 {
 					s.Logger.Printf("Cannot read response: %v\n", err)
 				}
-				return err
-			}
-			if s.LogLevel > 0 {
-				s.Logger.Printf("Request failed with body: %s (status: %v)\n", string(resBody), res.StatusCode)
+			} else {
+				if s.LogLevel > 0 {
+					s.Logger.Printf("Request failed with body: %s (status: %v)\n", string(resBody), res.StatusCode)
+				}
 			}
 		}
 
--- a/stripe.go
+++ b/stripe.go
@@ -351,13 +351,21 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			break
 		}
 
-		resBody, err := ioutil.ReadAll(res.Body)
-		res.Body.Close()
-		if err != nil {
-			if s.LogLevel > 0 {
-				s.Logger.Printf("Cannot read response: %v\n", err)
+		var (
+			resBody []byte
+			err error
+		)
+		// Checking res instead of err because s.HTTPClient.Do(req) has one special case for Go 1 compatibility
+		// when both the response and an error are returned
+		if res != nil {
+			resBody, err = ioutil.ReadAll(res.Body)
+			res.Body.Close()
+			if err != nil {
+				if s.LogLevel > 0 {
+					s.Logger.Printf("Cannot read response: %v\n", err)
+				}
+				return err
 			}
-			return err
 		}
 
 		if s.LogLevel > 0 {
--- a/stripe.go
+++ b/stripe.go
@@ -365,7 +365,7 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 				return err
 			}
 			if s.LogLevel > 0 {
-				s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
+				s.Logger.Printf("Request failed with body: %s (status: %v)\n", string(resBody), res.StatusCode)
 			}
 		}
 
--- a/stripe.go
+++ b/stripe.go
@@ -351,12 +351,8 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			break
 		}
 
-		var (
-			resBody []byte
-			err     error
-		)
-		// Checking res instead of err because s.HTTPClient.Do(req) has one special case for Go 1 compatibility
-		// when both the response and an error are returned
+		var resBody []byte
+
 		if res != nil {
 			resBody, err = ioutil.ReadAll(res.Body)
 			res.Body.Close()
@@ -389,6 +385,7 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 		}
 		return err
 	}
+
 	defer res.Body.Close()
 
 	resBody, err := ioutil.ReadAll(res.Body)
--- a/stripe.go
+++ b/stripe.go
@@ -351,10 +351,12 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			break
 		}
 
-		var resBody []byte
-
-		if res != nil {
-			resBody, err = ioutil.ReadAll(res.Body)
+		if err != nil {
+			if s.LogLevel > 0 {
+				s.Logger.Printf("Request failed with error: %v\n", err)
+			}
+		} else {
+			resBody, err := ioutil.ReadAll(res.Body)
 			res.Body.Close()
 			if err != nil {
 				if s.LogLevel > 0 {
@@ -362,10 +364,9 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 				}
 				return err
 			}
-		}
-
-		if s.LogLevel > 0 {
-			s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
+			if s.LogLevel > 0 {
+				s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
+			}
 		}
 
 		sleepDuration := s.sleepTime(retry)
--- a/stripe.go
+++ b/stripe.go
@@ -353,7 +353,7 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 
 		var (
 			resBody []byte
-			err error
+			err     error
 		)
 		// Checking res instead of err because s.HTTPClient.Do(req) has one special case for Go 1 compatibility
 		// when both the response and an error are returned
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -137,7 +137,6 @@ func (s *SafeCounter) Get() int {
 	return s.requestNum
 }
 
-// Tests client retries when HTTP returns timeout.
 func TestDo_RetryOnTimeout(t *testing.T) {
 	type testServerResponse struct {
 		Message string `json:"message"`
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	assert "github.com/stretchr/testify/require"
-	stripe "github.com/stripe/stripe-go"
+	"github.com/stripe/stripe-go"
 	. "github.com/stripe/stripe-go/testing"
 )
 
@@ -119,10 +119,9 @@ func TestDo_Retry(t *testing.T) {
 	assert.Equal(t, 2, requestNum)
 }
 
-
 // Types for TestDo_RetryOnTimeout test
 type SafeCounter struct {
-	mux sync.Mutex
+	mux        sync.Mutex
 	requestNum int
 }
 
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -13,7 +13,7 @@ import (
 	"time"
 
 	assert "github.com/stretchr/testify/require"
-	stripe "github.com/stripe/stripe-go"
+	"github.com/stripe/stripe-go"
 	. "github.com/stripe/stripe-go/testing"
 )
 
@@ -153,7 +153,7 @@ func TestDo_RetryOnTimeout(t *testing.T) {
 			LogLevel:          3,
 			MaxNetworkRetries: 2,
 			URL:               testServer.URL,
-			HTTPClient:  &http.Client{Timeout: timeout},
+			HTTPClient:        &http.Client{Timeout: timeout},
 		},
 	).(*stripe.BackendImplementation)
 
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -10,6 +10,7 @@ import (
 	"runtime"
 	"sync"
 	"testing"
+	"time"
 
 	assert "github.com/stretchr/testify/require"
 	stripe "github.com/stripe/stripe-go"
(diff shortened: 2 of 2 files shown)


remi-stripe:
@agatamigalska-newstore Thanks a lot for opening a PR so quickly and adding a test. This one is quite specific to some Go internals so I'll defer to @brandur-stripe who can look next week and he'll follow up with feedback!

stripe.go
```diff
@@ -351,13 +351,21 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
-		res.Body.Close()
-		if err != nil {
-			if s.LogLevel > 0 {
-				s.Logger.Printf("Cannot read response: %v\n", err)
+		var (
+			resBody []byte
+			err     error
+		)
+		// Checking res instead of err because s.HTTPClient.Do(req) has one special case for Go 1 compatibility
+		// when both the response and an error are returned
```
brandur-stripe: Can we clarify this comment a bit? It should read something like "when both the response and error are returned as nil" I think?

Also, what exactly do you mean by "Go 1 compatibility"? All Go versions right now are Go 1 right? Do you mean Go 1.10 or something like that?

stripe_test.go
```diff
@@ -118,6 +119,71 @@ func TestDo_Retry(t *testing.T) {
 	assert.Equal(t, 2, requestNum)
 }
 
+// Tests client retries when HTTP returns timeout.
```
brandur-stripe: Can we either elaborate on this comment or remove it? e.g. You could add something like "Timeouts can result in an edge whereby both a request's result and error come back as nil, which can fail the retry infrastructure if it's not handling it."

stripe_test.go
```diff
@@ -118,6 +119,71 @@ func TestDo_Retry(t *testing.T) {
...
+//
+// To run only this test:
+//
+//     go test . -run TestDo_RetryOnTimeout -test.v
+//
```
brandur-stripe: And I think this "run only this test" comment was copied from `TestDo_Retry` above. Let's just remove this part. I included it in `TestDo_Retry` because it gives you a hint at how to better understand what the test is doing, but in general there's sufficient instruction for this type of thing [in the README](https://github.com/stripe/stripe-go/issues/713).

stripe_test.go
```diff
@@ -118,6 +119,71 @@ func TestDo_Retry(t *testing.T) {
...
+
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		err := r.ParseForm()
+		assert.NoError(t, err)
+
+		// The body should always be the same with every retry. We've
+		// previously had regressions in this behavior as we switched to HTTP/2
+		// and `Request` became non-reusable, so we want to check it with every
+		// request.
+		assert.Equal(t, "bar", r.Form.Get("foo"))
```
brandur-stripe: Let's strip out this `assert` and just rely on the one in `TestDo_Retry` above (it and the comment above it were copied from there).

stripe_test.go
```diff
@@ -118,6 +119,71 @@ func TestDo_Retry(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	var body = bytes.NewBufferString("foo=bar")
+	var response testServerResponse
+
+	// make sure that there was no panic
+	defer func() {
+		var recovery = recover()
+		assert.Nil(t, recovery)
+	}()
```
brandur-stripe: I think it's sufficient to just let the test framework catch the panic for us and fail the run if one occurs. Let's strip out this `defer`.

stripe_test.go
```diff
@@ -118,6 +119,71 @@ func TestDo_Retry(t *testing.T) {
...
+		err := r.ParseForm()
+		assert.NoError(t, err)
+
+		// The body should always be the same with every retry. We've
+		// previously had regressions in this behavior as we switched to HTTP/2
+		// and `Request` became non-reusable, so we want to check it with every
+		// request.
+		assert.Equal(t, "bar", r.Form.Get("foo"))
+		time.Sleep(timeout)
+		requestNum++
```
brandur-stripe: Okay, this is a little subtle, but there's a race here which is causing the Travis build to fail on this test. Check there for details, but you should also be able to reproduce it locally by using `go test` with the `-race` flag (which is what we do in Travis):

```
go test . -run TestDo_RetryOnTimeout -test.v -race
```

I believe what's happening is that you have multiple server Goroutines that are active in this handler concurrently. Because the client is timing out early, the first request is still active on the server when the second is handled. The fix is pretty easy though: make sure to use a mutex around any variable accesses like this one that could be touched from multiple Goroutines.

stripe.go
agatamigalska-newstore: My comment was based on the following statement made in spripe-go's dependency `go-1.10/src/net/http/client.go` (lines 602 - 605) : 
```
				// Special case for Go 1 compatibility: return both the response
				// and an error if the CheckRedirect function failed.
				// See https://golang.org/issue/3795
				// The resp.Body has already been closed.
```
Current github repo don't have this comment anymore. Anyway, since I understood this library belongs to Go's core, I assumed that must make sense and didn't give it more thought. I still don't know what exactly they meant, so I updated my comment.

agatamigalska-newstore:
The build for Go 1.8 has failed due to 
```
The command "go get -u golang.org/x/net/http2" failed 3 times.
```
Any ideas?

agatamigalska-newstore:
@brandur-stripe Please review, thanks.

brandur-stripe:
@agatamigalska-newstore Ah! Okay, I misunderstood the premise of the PR a bit — re-reading the code now, I can see that we are not correctly handling the `err` case during a retry, which is obviously incorrect.

Which you mind changing your implementation a bit so that we very explicitly handle the case on `err` and print accordingly? Here's the diff I came up with if you just want to apply it directly:

``` patch
diff --git a/stripe.go b/stripe.go
index e356177..dc6fcf1 100644
--- a/stripe.go
+++ b/stripe.go
@@ -351,17 +351,23 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			break
 		}
 
-		resBody, err := ioutil.ReadAll(res.Body)
-		res.Body.Close()
 		if err != nil {
 			if s.LogLevel > 0 {
-				s.Logger.Printf("Cannot read response: %v\n", err)
+				s.Logger.Printf("Request failed with error: %v\n", err)
+			}
+		} else {
+			resBody, err := ioutil.ReadAll(res.Body)
+			res.Body.Close()
+			if err != nil {
+				if s.LogLevel > 0 {
+					s.Logger.Printf("Cannot read response: %v\n", err)
+				}
+				return err
 			}
-			return err
-		}
 
-		if s.LogLevel > 0 {
-			s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
+			if s.LogLevel > 0 {
+				s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
+			}
 		}
 
 		sleepDuration := s.sleepTime(retry)
```

Regarding your test, I gave you a little bit of bad advice. Go actually has an atomic counter built in, so would you mind if we switched to just using that. [Here's an example of using it](https://gobyexample.com/atomic-counters), but your implementation should look something like this:

``` go
var counter uint64

...

atomic.AddUint64(&counter, 1)

...

atomic.LoadUint64(&counter)
```

Thanks!

stripe_test.go
```diff
@@ -118,6 +119,74 @@ func TestDo_Retry(t *testing.T) {
...
+		"sk_test_123",
+		"application/x-www-form-urlencoded",
+		nil,
+	)
+	assert.NoError(t, err)
+
+	var body = bytes.NewBufferString("foo=bar")
+	var response testServerResponse
+
+	// make sure that there was no panic
```
brandur-stripe: Can we strip this comment? I think it's somewhat self-explanatory that we don't want a panic.

stripe_test.go
```diff
@@ -118,6 +119,74 @@ func TestDo_Retry(t *testing.T) {
...
+	err = backend.Do(request, body, &response)
+
+	// there should be an error returned
+	assert.Error(t, err)
```
brandur-stripe: And this comment too — it's pretty obvious what this assertion is trying to accomplish.

stripe_test.go
```diff
@@ -118,6 +119,74 @@ func TestDo_Retry(t *testing.T) {
...
+	// timeout should not prevent retry
+	assert.Equal(t, 2, counter.Get())
+	var recovery = recover()
+	assert.Nil(t, recovery)
```
brandur-stripe: And can we take out the `recover()` and subsequent assert? If there's a panic, the test run will fail automatically.

brandur-stripe:
> The build for Go 1.8 has failed due to 
> ```
> The command "go get -u golang.org/x/net/http2" failed 3 times.
> ```
> Any ideas? 

@agatamigalska-newstore Unfortunately an upstream change in Go's HTTP/2 package ended up breaking it for 1.8, and that took our build down with it. I dropped 1.8 from the build matrix in #716 and that's now merged, so you should be able to rebase on that to get things passing again.

agatamigalska-newstore:
Thanks for your comments @brandur-stripe. The code is ready for the next round :)

stripe.go
```diff
@@ -351,17 +351,22 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
+			resBody, err := ioutil.ReadAll(res.Body)
+			res.Body.Close()
+			if err != nil {
+				if s.LogLevel > 0 {
+					s.Logger.Printf("Cannot read response: %v\n", err)
+				}
+				return err
+			}
+			if s.LogLevel > 0 {
+				s.Logger.Printf("Request failed with: %s (error: %v)\n", string(resBody), err)
```
brandur-stripe: @agatamigalska-newstore Sorry, I know this is code I gave you, but reading it again, I now realize that on this last log line `err` will never be non-`nil` because we're now handling that case above. It would also be kind of nice if we logged the response status code because that's one of the factors that the library will use to determine the retry.

Would you mind changing it to:

``` go
s.Logger.Printf("Request failed with body: %s (status: %v)\n", string(resBody), res.StatusCode)

Comment (brandur-stripe) [reply stripe.go]: And thanks for making all those other changes!

After this last one, this looks good to me.

Comment (agatamigalska-newstore) [reply stripe.go]: Updated as requested, this time I also read it instead of just copy-pasting ;)

File: stripe.go [hunk 11]
```diff
@@ -351,17 +351,22 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
+				s.Logger.Printf("Request failed with error: %v\n", err)
+			}
+		} else {
+			resBody, err := ioutil.ReadAll(res.Body)
+			res.Body.Close()
+			if err != nil {
+				if s.LogLevel > 0 {
+					s.Logger.Printf("Cannot read response: %v\n", err)
+				}
+				return err
```
agatamigalska-newstore: @brandur-stripe This return statement was here before I started messing with this code, and I'm wondering - why not let it retry? It seems like the only case when (apart from shouldRetry() returning false) the retrial is not permitted, and I wonder why?

brandur-stripe: I suspect that it wasn't originally retried because it's kind of an "error inside of an error" condition, and also probably somewhat unlikely to occur (usually once you have the response you'll get back a response body).

If you want, you can start retrying it as well since you're right in that it'd be strictly more correct.

agatamigalska-newstore: I removed the return statement and amended logged messages. Please have a look @brandur-stripe

brandur-stripe:
Excellent! Thanks for all the work here.

brandur-stripe:
Released as 53.0.1.

PR #715, opened 2018-11-06
Ship changes to the Payment Intent resource

remi-stripe:
The Payment Intent resource had some changes. This PR addresses all the changes. This will be released as a major version and we will automatically return the old shape based on previous stripe-go versions.

cc @danwang-stripe @stripe/api-libraries 
r? @brandur-stripe

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -4,6 +4,16 @@ import (
 	"encoding/json"
 )
 
+// PaymentIntentCancellationReason is the list of allowed values for the cancelation reason.
+type PaymentIntentCancellationReason string
+
+// List of values that PaymentIntentCaptureMethod can take.
+const (
+	PaymentIntentCancellationReasonDuplicate           PaymentIntentCancellationReason = "duplicate"
+	PaymentIntentCancellationReasonFraudulent          PaymentIntentCancellationReason = "fraudulent"
+	PaymentIntentCancellationReasonRequestedByCustomer PaymentIntentCancellationReason = "requested_by_customer"
+)
+
 // PaymentIntentCaptureMethod is the list of allowed values for the capture method.
 type PaymentIntentCaptureMethod string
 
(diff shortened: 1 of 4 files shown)


brandur-stripe:
LGTM.

brandur-stripe:
Arg, I think the core team has dropped support for 1.8 in the HTTP/2 package :/

It's not great to drop 1.8 from our matrix so soon after 1.7, but technically the Go team doesn't support it, and since they've now broken compatibility, we might not have a choice.

remi-stripe:
@brandur-stripe yeah do we have stats on how many people use Go 1.8?

brandur-stripe:
@remi-stripe Just under 10% :/

remi-stripe:
That seems really high to drop support for it. Is the issue fixable?

brandur-stripe:
> That seems really high to drop support for it. Is the issue fixable?

Barring trying to lobby the core team to not break old versions or dropping HTTP/2 support, unfortunately I don't think so.

The good news is that upgrades in Go are _way_ easier than almost any other language. 10% may be a lot, but I don't think they're going to experience any significant lock in to 1.8.

remi-stripe:
@brandur-stripe Can I leave to you to open a separate PR for dropping support? Or should we do it in the same PR?

brandur-stripe:
@remi-stripe Yeah, I can open one. Just sec.

brandur-stripe:
@remi-stripe #716 should probably have a major release, and so should this, so I didn't release for the time being so that we can roll these up together.

remi-stripe:
@brandur-stripe While working on other additions I discovered another thing I missed before and this made me rethink the way I do params on Payment Intents. We now have one struct for create/update and then one struct per endpoint (to match what we did on invoices in the other PR).

Can you re-review?

brandur-stripe:
@remi-stripe Yeah, I like the new approach better and it's good that it's more consistent — sorry I missed it before too.

LGTM.

PR #716, opened 2018-11-06
Drop Go 1.8 from the build matrix

brandur:
Unfortunately, it appears that new changes in the
`golang.org/x/net/http2` package have caused it to no longer build under
Go 1.8.

As noted in Go's release policy [1], the latest two Go versions are
officially supported, making that Go 1.11 and Go 1.10 currently.

We do see ~10% users on Go 1.8 still which is larger than we'd hope when
dropping an old version, but we don't have too much choice here. We
believe that an upgrade to Go 1.9 for those users should be
straightforward given that there generally no backward incompatible
changes between new Go versions.

See also discussion in #715.

[1] https://golang.org/doc/devel/release.html#policy

r? @remi-stripe


remi-stripe:
LGTM

brandur-stripe:
Thanks Remi!

PR #721, opened 2018-11-10
Add ValueList and ValueListItem for Radar and other Invoice changes

remi-stripe:
cc @stripe/api-libraries 

While I removed 2 properties from `Invoice`, those were not supported anymore because of the API version change so not sure if it's a breaking change but flagging just in case.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -39,6 +39,8 @@ import (
 	"github.com/stripe/stripe-go/payout"
 	"github.com/stripe/stripe-go/plan"
 	"github.com/stripe/stripe-go/product"
+	"github.com/stripe/stripe-go/radar/valuelist"
+	"github.com/stripe/stripe-go/radar/valuelistitem"
 	"github.com/stripe/stripe-go/recipient"
 	"github.com/stripe/stripe-go/refund"
 	"github.com/stripe/stripe-go/reporting/reportrun"
@@ -170,6 +172,10 @@ type API struct {
 	Transfers *transfer.Client
 	// UsageRecords is the client used to invoke usage record related APIs.
 	UsageRecords *usagerecord.Client
+	// ValueLists is the client used to invoke /radar/value_lists APIs.
+	ValueLists *valuelist.Client
+	// ValueListItems is the client used to invoke /radar/value_list_items APIs.
+	ValueListItems *valuelistitem.Client
 	// WebhookEndpoints is the client used to invoke usage record related APIs.
 	WebhookEndpoints *webhookendpoint.Client
 }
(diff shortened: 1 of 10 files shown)


client/api.go
```diff
@@ -170,6 +172,10 @@ type API struct {
 	Transfers *transfer.Client
 	// UsageRecords is the client used to invoke usage record related APIs.
 	UsageRecords *usagerecord.Client
+	// ValueLists is the client used to invoke /radar/value_lists APIs.
+	ValueLists *valuelist.Client
+	// ValueListItems is the client used to invoke /radar/value_list_items APIs.
+	ValueListItems *valuelistitem.Client
```
brandur-stripe: I think we're generally prefixing these with the project namespace, so like `RadarValueLists`.

radar_valuelist.go
```diff
@@ -0,0 +1,55 @@
...
+	ValueListItemTypeCardFingerprint     ValueListItemType = "card_fingerprint"
+	ValueListItemTypeCountry             ValueListItemType = "country"
+	ValueListItemTypeEmail               ValueListItemType = "email"
+	ValueListItemTypeIPAddress           ValueListItemType = "ip_address"
+	ValueListItemTypeString              ValueListItemType = "string"
+	ValueListItemTypeCaseSensitiveString ValueListItemType = "case_sensitive_string"
+)
+
+// ValueListParams is the set of parameters that can be used when creating a value list.
+type ValueListParams struct {
```
brandur-stripe: And likewise, I think for structs in these shared namespaces we're prefixing with the project like `RadarValueListParams`. See what we have in `issuing_cardholder.go` for example.

remi-stripe:
@brandur-stripe Nice catch. Fixed, PTAL

client/api.go
```diff
@@ -170,6 +172,10 @@ type API struct {
 	Transfers *transfer.Client
 	// UsageRecords is the client used to invoke usage record related APIs.
 	UsageRecords *usagerecord.Client
+	// RadarValueLists is the client used to invoke /radar/value_lists APIs.
+	RadarValueLists *valuelist.Client
+	// RadarValueListItems is the client used to invoke /radar/value_list_items APIs.
+	RadarValueListItems *valuelistitem.Client
```
brandur-stripe: Sorry for the nit, but now that these are renamed, let's shift them up a bit to fall into alphabetical order (and same with where they're initialized below).

radar_valuelist.go
```diff
@@ -0,0 +1,55 @@
...
+// RadarValueListParams is the set of parameters that can be used when creating a value list.
+type RadarValueListParams struct {
+	Params   `form:"*"`
+	Alias    *string `form:"alias"`
+	ItemType *string `form:"item_type"`
+	Name     *string `form:"name"`
+}
+
+// RadarValueListListParams is the set of parameters that can be used when listing value lists.
+type RadarValueListListParams struct {
```
brandur-stripe: We made a `ListList` :/

client/api.go
remi-stripe: Thanks, fixed!

PR #728, opened 2018-11-25
Add IssuingCard param value to ephemeral_key.go

jack-stripe:
We now accept an `issuing_card` parameter in the API when making an ephemeral key; this allows doing so from the Go bindings.

r? @remi-stripe

### DIFF
--- a/ephemeralkey.go
+++ b/ephemeralkey.go
@@ -7,8 +7,8 @@ import "encoding/json"
 // For more details see https://stripe.com/docs/api#ephemeral_keys.
 type EphemeralKeyParams struct {
 	Params        `form:"*"`
-  Customer      *string `form:"customer"`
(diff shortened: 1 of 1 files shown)


remi-stripe:
Going to re-assign to @brandur-stripe because of the build errors. Likely something else that changed recently in one of our third-party libraries.

brandur-stripe:
Thanks @jack-stripe!

I think your build errors are just because you're using an outdated `.travis.yml` file — we've since addressed these issues in various other PRs. Would you mind rebasing and re-pushing?

jack-stripe:
Oh, my bad! Just re-pushed.

brandur-stripe:
LGTM!

brandur-stripe:
Released as 53.4.0.

PR #730, opened 2018-11-28
Add support for the Review resource

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/review.go
+++ b/review.go
@@ -15,15 +15,40 @@ const (
 	ReviewReasonRule            ReviewReasonType = "rule"
 )
 
+// ReviewParams is the set of parameters that can be used when approving a review.
+type ReviewParams struct {
+	Params `form:"*"`
+}
+
+// ReviewApproveParams is the set of parameters that can be used when approving a review.
+type ReviewApproveParams struct {
+	Params `form:"*"`
+}
+
+// ReviewListParams is the set of parameters that can be used when listing reviews.
+type ReviewListParams struct {
+	ListParams   `form:"*"`
+	Created      *int64            `form:"created"`
+	CreatedRange *RangeQueryParams `form:"created"`
+}
+
 // Review is the resource representing a Radar review.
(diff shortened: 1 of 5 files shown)


review.go
```diff
@@ -15,15 +15,40 @@ const (
...
 // Review is the resource representing a Radar review.
 // For more details see https://stripe.com/docs/api#reviews.
 type Review struct {
-	Charge   *Charge          `json:"charge"`
-	Created  int64            `json:"created"`
-	ID       string           `json:"id"`
-	Livemode bool             `json:"livemode"`
-	Open     bool             `json:"open"`
-	Reason   ReviewReasonType `json:"reason"`
+	Review        *Review          `json:"review"`
```
brandur-stripe: I think this is supposed to be `Charge`/`*Charge`/`json:"charge"` right?

review/client.go
```diff
@@ -0,0 +1,75 @@
...
+	Key string
+}
+
+// Approve approves a review.
+func Approve(id string, params *stripe.ReviewApproveParams) (*stripe.Review, error) {
+	return getC().Approve(id, params)
+}
+
+// Approve approves a review.
+func (c Client) Approve(id string, params *stripe.ReviewApproveParams) (*stripe.Review, error) {
```
brandur-stripe: Thanks for following convention with the empty params struct here! It feels a little funny to do, but I like that we're staying consistent.

brandur-stripe:
Minor comment, but LGTM.

ptal @remi-stripe

review/client.go
remi-stripe: yeah and I do feel like we'll have params soon!

review.go
remi-stripe: Damn nice catch, I messed up my string replace

remi-stripe:
@brandur-stripe PTAL

brandur-stripe:
LGTM!

PR #732, opened 2018-11-28
Added support for creating Customers with custom ID

garygrace:
Added support for creating Customers with a custom ID

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -31,6 +31,7 @@ type CustomerParams struct {
 	DefaultSource  *string                        `form:"default_source"`
 	Description    *string                        `form:"description"`
 	Email          *string                        `form:"email"`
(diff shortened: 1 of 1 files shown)


remi-stripe:
@garygrace This feature has been deprecated for almost 5 years now and it's definitely not something we want to support moving forward. It's available for legacy reasons but we discourage anyone from trying to use this. If you're just getting started with Stripe, I would highly recommend not using this at all, even if it seems to work, and rely on our generation of ids for you (cus_XXXX) and then do your own mapping in your database or set your own id inside the `metadata` [property](https://stripe.com/docs/api#metadata) instead.

Because of this, we would not merge this feature in the library for anyone to find and assume it's okay to use. I hope it makes sense!

PR #733, opened 2018-11-30
Fix uploads API when using NewBackends

Oralordos:
When testing uploading files using an API client, I found that it would fail consistently with a strange error:

```go
{"code":"invalid_utf8_in_post_body","status":400,"message":"Invalid UTF-8 characters found in POST body: \"application/octet-stream\".  For assistance, please contact support@stripe.com.","type":"invalid_request_error"}
```

Adding some log statements to the `backendImplementation.Do` function, I found that it was trying to send a `files.New` request to `https://api.stripe.com/v1/v1/files`

Eventually I tracked the problem down to the `NewBackends` function. It reused the config object when creating the api and uploads backends. The `GetBackendWithConfig` function then modified the config object. So when it was called again with the same config object, it would already have the URL field filled out.

Simply put, we need two copies of the config object to avoid causing problems with the modified config object.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -689,10 +689,11 @@ func Int64Value(v *int64) int64 {
 // NewBackends creates a new set of backends with the given HTTP client. You
 // should only need to use this for testing purposes or on App Engine.
 func NewBackends(httpClient *http.Client) *Backends {
-	config := &BackendConfig{HTTPClient: httpClient}
+	apiConfig := &BackendConfig{HTTPClient: httpClient}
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Geeze, sorry about the trouble here, and really excellent debugging work to get to the bottom of it.

I think I'm going to try and unwind the `/v1` string prepend thing — it really feels ugly and unnecessary., and we should really just be explicit in our URLs instead. Because `GetBackendWithConfig` is fundamentally mutating though, it looks like we should be duplicating the config object anyway.

Thanks! LGTM.

PR #737, opened 2018-12-05
Add support for capabilities on the `Account` resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/account_test.go
+++ b/account_test.go
@@ -33,6 +33,10 @@ func TestAccountExternalAccountParams_AppendTo(t *testing.T) {
 func TestAccount_Unmarshal(t *testing.T) {
 	accountData := map[string]interface{}{
 		"id": "acct_123",
+		"capabilities": map[string]interface{}{
+			"card_payments":     "active",
+			"platform_payments": "inactive",
+		},
 		"decline_charge_on": map[string]interface{}{
 			"avs_failure": true,
 			"cvc_failure": false,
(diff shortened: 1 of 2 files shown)


account_test.go
```diff
@@ -160,6 +164,10 @@ func TestAccount_Unmarshal(t *testing.T) {
 
 	assert.Equal(t, "acct_123", account.ID)
 
+	assert.Equal(t, stripe.AAccountCapabilityStatusActive, account.Capabilities.CardPayments)
+	// TODO: Figure out how to test that account.Capabilities.LegacyPayments is null/not present.
```
remi-stripe: @brandur-stripe Do you know how to approach this in Go?

brandur-stripe: Yeah, it's a little hacky, but basically `AccountCapability` is defined as a type based off string, which means that like string, it's possible for it to have an empty value `""` which is also the zero value for the string type. When `json` sees an unset/null value for a non-pointer type like string, it sets it to the type's zero value, so the equivalent of `""`.

Therefore, you can check an empty by converting an empty string to an `AccountCapabilityStatus` like this:

``` go
assert.Equal(t, AccountCapabilityStatus(""), account.Capabilities.LegacyPayments)
```

remi-stripe: gotcha. Is that clean/go-like though? Or would it be better to have a default constant `AccountCapabilityStatusNone` mapping to `""`?

PR #739, opened 2018-12-11
Use ApplicationFee struct for FeeRefund.Fee

drtz:
Adds support for expanding fee_refund.fee


remi-stripe:
Looks good to me! Assigning to @brandur-stripe to confirm and decide if we should release now since it's a breaking change!

brandur-stripe:
Thanks @drtz!

Going to go for the release. We don't have any other major changes rolling out anytime soon that this one can be grouped with, so we may as well do it now. It'll be a very minor major version.

brandur-stripe:
Released as 55.0.0.

PR #740, opened 2018-12-20
Add support for the Checkout Session resource

remi-stripe:
This resource is now ready to be released. The PR has been updated to match the latest spec.

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/checkout/session/client_test.go
+++ b/checkout/session/client_test.go
@@ -1,0 +1,51 @@
+package session
+
+import (
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	stripe "github.com/stripe/stripe-go"
+	_ "github.com/stripe/stripe-go/testing"
+)
+
+func TestCheckoutSessionGet(t *testing.T) {
+	session, err := Get("cs_123", nil)
+	assert.Nil(t, err)
+	assert.NotNil(t, session)
+}
+
+func TestCheckoutSessionNew(t *testing.T) {
+	session, err := New(&stripe.CheckoutSessionParams{
+		CancelURL:         stripe.String("https://stripe.com/cancel"),
+		ClientReferenceID: stripe.String("1234"),
+		LineItems: []*stripe.CheckoutSessionLineItemParams{
+			{
+				Amount:      stripe.Int64(1234),
+				Currency:    stripe.String(string(stripe.CurrencyUSD)),
+				Description: stripe.String("description"),
+				Images: []*string{
+					stripe.String("https://stripe.com/image1"),
+				},
+				Name:     stripe.String("name"),
+				Quantity: stripe.Int64(2),
+			},
+		},
+		PaymentIntentData: &stripe.CheckoutSessionPaymentIntentDataParams{
+			Description: stripe.String("description"),
+			Shipping: &stripe.ShippingDetailsParams{
+				Address: &stripe.AddressParams{
+					Line1: stripe.String("line1"),
+					City:  stripe.String("city"),
+				},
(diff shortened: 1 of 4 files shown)


remi-stripe:
Related to https://github.com/stripe/stripe-dotnet/pull/1440

checkout/session/client_test.go
```diff
@@ -0,0 +1,45 @@
...
+			},
+		},
+		PaymentMethodTypes: []*string{
+			stripe.String("card"),
+		},
+		SuccessURL: stripe.String("https://stripe.com/success"),
+	})
+	assert.Nil(t, err)
+	assert.NotNil(t, session)
+}
```
brandur-stripe: Should we try to exercise `Get` as well?

checkout_session.go
```diff
@@ -0,0 +1,134 @@
...
+type CheckoutSessionSubscriptionDataItemsParams struct {
+	Plan     *string `form:"plan"`
+	Quantity *int64  `form:"quantity"`
+}
+
+// CheckoutSessionSubscriptionDataParams is the set of parameters allowed for the subscription
+// creation on a checkout session.
+type CheckoutSessionSubscriptionDataParams struct {
+	Items           *CheckoutSessionSubscriptionDataItemsParams `form:"items"`
+	TrialEnd        *int64                                      `form:"trial_end"`
```
brandur-stripe: This might also need its own `Params` so that it can get a `Metadata`.

checkout_session.go
```diff
@@ -0,0 +1,134 @@
...
+	Description *string   `form:"description"`
+	Images      []*string `form:"images"`
+	Name        *string   `form:"name"`
+	Quantity    *int64    `form:"quantity"`
+}
+
+// CheckoutSessionPaymentIntentDataTransferDataParams is the set of parameters allowed for the
+// transfer_data hash.
+type CheckoutSessionPaymentIntentDataTransferDataParams struct {
+	Destination *string `form:"destination"`
```
brandur-stripe: Should this also have `Amount`?

checkout_session.go
remi-stripe: Ah yes I forgot to ask you: does it just work? Do we explicitly go through nested `Params` the right way?

checkout_session.go
remi-stripe: No `amount` is gated in theory (and there for PI because it was ungated before launch)

checkout_session.go
brandur-stripe: You'd have to write a test to be 100% sure, but yeah, `Params` doesn't do anything special really, so it should work fine even if nested down on a substructure like this.

The one downside of course is that `Params` has a bunch of things on it that are not relevant to this struct (e.g. idempotency key, headers, `Stripe-Account` ...), so another alternative would be just to add `Metadata map[string]string `form:"metadata"`` directly. The downside of that of course is that you'd also have to manually add the `AddMetadata` helper to the struct if you wanted to support that.

checkout_session.go
brandur-stripe: Ah, got it. Okay, thanks!

checkout_session.go
remi-stripe: Woot that magically just works. Nice catch! I think it's likely easier and more explicit that way than having to do something custom that users would not expect.

remi-stripe:
@brandur-stripe Thanks for catching those issues. I just fixed each one. PTAL

checkout_session.go
brandur-stripe: Okay, WFM.

PR #741, opened 2018-12-26
Add support for invoice_now and prorate on subscription cancelation

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM. Thanks Remi!

brandur-stripe:
(And yet another case where I'm really glad we started giving _all_ API methods their own parameter struct, even if they take no parameters initially.)

Released as 55.2.0.

PR #743, opened 2018-12-27
Add ClearUsage param to SubscriptionItemParams

eclipse1985:
Accoring with the [documentation](https://stripe.com/docs/api/subscription_items/delete?lang=go), we need a ClearUsage parameter.

Right now is not possible to delete subscription items due to error:

> Error encountered from Stripe: {"status":400,"message":"When deleting the subscription item ITEM_ID with a metered plan, all related usage records must be cleared too by setting `clear_usage=true`.","request_id":"REQ_ID","type":"invalid_request_error"}


remi-stripe:
Released in 55.1.0

PR #763, opened 2019-01-09
Add support for `application_fee_amount` on Charge

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM.

Just flagging that it seems that the way this was implemented, `application_fee_amount` parameters on create and capture won't show up in the docs (and probably not the OpenAPI spec either), which isn't super great because I think stripe-mock would reject the request if you included that parameter.

brandur-stripe:
Released as 55.8.0.

PR #765, opened 2019-01-11
Add new properties to `TransferReversal`

remi-stripe:
Add support for `destination_payment_refund` and `source_refund` on `TransferReversal`.

cc @stripe/api-libraries 

NOT asking for review yet as those properties are not expandable and I want to make sure it's expected to avoid a future breaking change.

### DIFF
--- a/reversal.go
+++ b/reversal.go
@@ -18,13 +18,15 @@ type ReversalListParams struct {
 
 // Reversal represents a transfer reversal.
 type Reversal struct {
-	Amount             int64               `json:"amount"`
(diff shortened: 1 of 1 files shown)


mickjermsurawong-stripe:
ah.. since I was already looking.. I can verify that the two fields are expandable resources for Refund!

remi-stripe:
Yeah, this was just added in the API and I released stripe-mock 0.43.0 for this, but stripe-go never explicitly tests this (because the deserialization of a resource is already well tested elsewhere)

brandur-stripe:
LGTM @remi-stripe!

I'll leave it to you to release unless you'd like me to.

PR #766, opened 2019-01-11
Add Stripe client telemetry to request headers

jameshageman-stripe:
Follows https://github.com/stripe/stripe-ruby/pull/696, https://github.com/stripe/stripe-php/pull/549, and https://github.com/stripe/stripe-python/pull/518 in adding telemetry metadata to request headers. 

The telemetry is disabled by default, and can be enabled per-client like so:

```go
backend := stripe.GetBackendWithConfig(
	stripe.APIBackend,
	&stripe.BackendConfig{
		URL: "https://api.stripe.com/v1",
		EnableTelemetry: true,
	},
)
```

or globally by setting

```go
stripe.EnableTelemetry = true
```

cc @dcarney-stripe @bobby-stripe

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -67,10 +67,6 @@ var LogLevel = 2
 // be overridden if a backend is created with GetBackendWithConfig.
 var Logger Printfer
 
-// EnableTelemetry allows request metrics (request id and duration) to be sent
-// to Stripe in subsequent requests via the `X-Stripe-Client-Telemetry` header.
-var EnableTelemetry = false
-
 //
 // Public types
 //
@@ -142,13 +138,24 @@ type BackendConfig struct {
 	//
 	// If left empty, it'll be set to the default for the SupportedBackend.
 	URL string
+
+	// EnableTelemetry allows request metrics (request id and duration) to be sent
+	// to Stripe in subsequent requests via the `X-Stripe-Client-Telemetry` header.
+	//
+	// Defaults to false.
+	EnableTelemetry bool
 }
 
-// RequestMetrics contains the payload sent in the `X-Stripe-Client-Telemetry`
-// header when stripe.EnableTelemetry = true.
+// RequestMetrics contains the id and duration of the last request sent
 type RequestMetrics struct {
 	RequestID         string `json:"request_id"`
 	RequestDurationMS int    `json:"request_duration_ms"`
+}
+
+// RequestTelemetry contains the payload sent in the
+// `X-Stripe-Client-Telemetry` header when stripe.EnableTelemetry = true.
+type RequestTelemetry struct {
+	LastRequestMetrics RequestMetrics `json:"last_request_metrics"`
 }
 
 // BackendImplementation is the internal implementation for making HTTP calls
@@ -169,7 +176,9 @@ type BackendImplementation struct {
 	//
 	// See also SetNetworkRetriesSleep.
 	networkRetriesSleep bool
-	lastRequestMetrics  *RequestMetrics
+
+	enableTelemetry    bool
+	lastRequestMetrics *RequestMetrics
 }
 
 // Call is the Backend.Call implementation for invoking Stripe APIs.
@@ -304,10 +313,13 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 		s.Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
 	}
 
-	if EnableTelemetry && s.lastRequestMetrics != nil {
-		metricsJSON, _ := json.Marshal(s.lastRequestMetrics)
-		payload := fmt.Sprintf(`{"last_request_metrics":%s}`, metricsJSON)
-		req.Header.Set("X-Stripe-Client-Telemetry", payload)
+	if s.enableTelemetry && s.lastRequestMetrics != nil {
+		metricsJSON, err := json.Marshal(&RequestTelemetry{LastRequestMetrics: *s.lastRequestMetrics})
+		if err == nil {
+			req.Header.Set("X-Stripe-Client-Telemetry", string(metricsJSON))
+		} else if s.LogLevel > 2 {
+			s.Logger.Printf("Unable to encode client telemetry: %s", err)
+		}
 	}
 
 	var res *http.Response
@@ -412,7 +424,7 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 		return err
 	}
 
-	if EnableTelemetry {
+	if s.enableTelemetry {
 		reqID := res.Header.Get("Request-Id")
 		if len(reqID) > 0 {
 			s.lastRequestMetrics = &RequestMetrics{
@@ -921,6 +933,7 @@ func newBackendImplementation(backendType SupportedBackend, config *BackendConfi
 		Type:                backendType,
 		URL:                 config.URL,
 		networkRetriesSleep: true,
+		enableTelemetry:     config.EnableTelemetry,
 		lastRequestMetrics:  nil,
 	}
 }
--- a/stripe.go
+++ b/stripe.go
@@ -176,7 +176,6 @@ type BackendImplementation struct {
 
 	enableTelemetry      bool
 	requestMetricsBuffer chan requestMetrics
-	requestMetricsMutex  *sync.Mutex
 }
 
 // Call is the Backend.Call implementation for invoking Stripe APIs.
@@ -435,7 +434,12 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 				RequestDurationMS: requestDurationMS,
 			}
 
-			s.recordRequestMetrics(metrics)
+			// If the metrics buffer is full, discard the new metrics. Otherwise, add
+			// them to the buffer.
+			select {
+			case s.requestMetricsBuffer <- metrics:
+			default:
+			}
 		}
 	}
 
@@ -954,34 +958,15 @@ func newBackendImplementation(backendType SupportedBackend, config *BackendConfi
 	}
 
 	return &BackendImplementation{
-		HTTPClient:          config.HTTPClient,
-		LogLevel:            config.LogLevel,
-		Logger:              config.Logger,
-		MaxNetworkRetries:   config.MaxNetworkRetries,
-		Type:                backendType,
-		URL:                 config.URL,
-		networkRetriesSleep: true,
-		enableTelemetry:     enableTelemetry,
-
-		// requestMetricsBuffer is a circular buffer of unsent metrics from previous
-		// requests. You should not write to requestMetricsBuffer without holding the
-		// requestMetricsMutex lock.
+		HTTPClient:           config.HTTPClient,
+		LogLevel:             config.LogLevel,
+		Logger:               config.Logger,
+		MaxNetworkRetries:    config.MaxNetworkRetries,
+		Type:                 backendType,
+		URL:                  config.URL,
+		networkRetriesSleep:  true,
+		enableTelemetry:      enableTelemetry,
 		requestMetricsBuffer: requestMetricsBuffer,
-		requestMetricsMutex:  &sync.Mutex{},
-	}
-}
-
-func (s *BackendImplementation) recordRequestMetrics(r requestMetrics) {
-	s.requestMetricsMutex.Lock()
-	defer s.requestMetricsMutex.Unlock()
-
-	// treat requestMetricsBuffer as a circular buffer: if it is full, pop off the
-	// oldest requestMetrics struct and insert r.
-	select {
-	case s.requestMetricsBuffer <- r:
-	default:
-		<-s.requestMetricsBuffer
-		s.requestMetricsBuffer <- r
 	}
 }
 
--- a/stripe.go
+++ b/stripe.go
@@ -67,6 +67,15 @@ var LogLevel = 2
 // be overridden if a backend is created with GetBackendWithConfig.
 var Logger Printfer
 
+// EnableTelemetry is a global override for enabling client telemetry, which
+// sends request performance metrics to Stripe via the `X-Stripe-Client-Telemetry`
+// header. If set to true, all clients will send telemetry metrics. Defaults to
+// false.
+//
+// Telemetry can also be enabled on a per-client basis by instead creating a
+// `BackendConfig` with `EnableTelemetry: true`.
+var EnableTelemetry = false
+
 //
 // Public types
 //
@@ -937,9 +946,10 @@ func isHTTPWriteMethod(method string) bool {
 // instead of this function.
 func newBackendImplementation(backendType SupportedBackend, config *BackendConfig) Backend {
 	var requestMetricsBuffer chan requestMetrics
+	enableTelemetry := config.EnableTelemetry || EnableTelemetry
 
 	// only allocate the requestMetrics buffer if client telemetry is enabled.
-	if config.EnableTelemetry {
+	if enableTelemetry {
 		requestMetricsBuffer = make(chan requestMetrics, telemetryBufferSize)
 	}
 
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -158,6 +169,7 @@ type BackendImplementation struct {
 	//
 	// See also SetNetworkRetriesSleep.
 	networkRetriesSleep bool
+	lastRequestMetrics  *RequestMetrics
```
brandur-stripe: Because `networkRetriesSleep` above has its own comment, could you just add a newline above this one just to show that it's not part of the same group?

stripe.go
```diff
@@ -140,6 +144,13 @@ type BackendConfig struct {
 	URL string
 }
 
+// RequestMetrics contains the payload sent in the `X-Stripe-Client-Telemetry`
+// header when stripe.EnableTelemetry = true.
+type RequestMetrics struct {
```
brandur-stripe: Let's call this `requestMetrics` instead so that it's not exported outside the package (whether the first letter is lower case or upper case determines this).

stripe.go
```diff
@@ -292,8 +304,15 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 		s.Logger.Printf("Requesting %v %v%v\n", req.Method, req.URL.Host, req.URL.Path)
 	}
 
+	if EnableTelemetry && s.lastRequestMetrics != nil {
+		metricsJSON, _ := json.Marshal(s.lastRequestMetrics)
```
brandur-stripe: This one is unlikely to happen, but let's not swallow errors just in case we introduce something in the future that becomes difficult to debug because we're throwing some message away. Observe `err` and handle if it's non-`nil` like elsewhere.

stripe.go
```diff
@@ -292,8 +304,15 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
+		payload := fmt.Sprintf(`{"last_request_metrics":%s}`, metricsJSON)
```
brandur-stripe: Just for cleanliness, could you create a wrapper struct for this one instead of manually assembling the JSON string (maybe `requestTelemetry`)?

stripe.go
jameshageman-stripe: I initially did make `RequestMetrics` private, however I wanted the tests to ensure that I could unmarshal the sent metrics back into this struct. Do you think it's ok to omit that check? Otherwise, I could duplicate the `requestMetrics` definition in `stripe_test.go` for the sake of unmarshaling.

jameshageman-stripe:
@brandur-stripe updated! Left one unresolved comment re: `RequestMetrics` visibility.

jameshageman-stripe:
cc @akropp-stripe

brandur-stripe:
Thanks for the updates!

And oops — unfortunately I overlooked something major on the last review, which as that as implemented, this is extremely unsafe from a concurrency perspective (which is fairly important from Go). As multiple Goroutines make requests, they'll all be vying to write `lastRequestMetrics`. Similarly as new requests are being issued, what's in `lastRequestMetrics` at any given moment is anybody's guess. Wrapping accesses in a mutex isn't enough because even with that, the values being read and written will be so volatile.

It's not super obvious to me how to solve this one well (maybe a bounded queue?), but we'll need to rethink the approach.

jameshageman-stripe:
That's a good point! Just using a mutex around the value would cause metrics to be discarded if multiple calls finished at the same time. It doesn't seem like a _huge_ deal if metrics are missed, but it would be nice to handle concurrency in some capacity.

Perhaps we could use a buffered channel to keep a finite amount of request metrics around (say, 16?), but only perform non-blocking writes to the channel and discard the metrics when the channel is full. Likewise, new requests can pull a `requestMetrics` struct off the channel if it is not empty. This would allow concurrently produced metrics to be kept around, without introducing any blocking calls and unbounded memory allocations. Let me know how that sounds to you.

While we're at it, I'd like to try to add a test that detects the existing data race.

brandur-stripe:
> Perhaps we could use a buffered channel to keep a finite amount of request metrics around (say, 16?), but only perform non-blocking writes to the channel and discard the metrics when the channel is full. Likewise, new requests can pull a `requestMetrics` struct off the channel if it is not empty. This would allow concurrently produced metrics to be kept around, without introducing any blocking calls and unbounded memory allocations. Let me know how that sounds to you.

Yep, that sounds good to me.

One minor tweak that might be worth considering: a circular buffer would probably be slightly better here than a simple buffered channel because in the case where it's full, we'd favor more recent data points instead of older data points. That said, probably not worth adding a dependency for.

jameshageman-stripe:
@brandur I think we could accomplish something similar in pure go without a dependency. On write, if the buffer is full, we could spin up a goroutine to push an element into the buffer and simulaneously pop the oldest element off. It introduces an ephemeral goroutine, but it would allow more recent metrics to have priority.

Something like this:

```go
// after a request is made in BackendImplementation#Do()

metrics := requestMetrics{
	RequestID:         reqID,
	RequestDurationMS: requestDurationMS,
}

select {
case s.prevRequestMetrics <- metrics:
	// Non-blocking insert succeeded, buffer was not full.
default:
	// Buffer was full, so pop off the oldest value and insert a newer one.
	// This is done in a goroutine to prevent blocking the client.
	go func() {
		<-s.prevRequestMetrics
		s.prevRequestMetrics <- metrics
	}()
}
```

I'm not sure how I feel about spawning a whole new goroutine for this, so I'd like to know your thoughts.

stripe.go
brandur-stripe: Oops, sorry I missed this:

I actually forgot that `stripe_test.go` is in its own `stripe_test` package. IMO, we should probably just put it in `stripe`. All the tests in subpackages are just in the same namespace as their accompanying file (`customer/`, `charge/`, `form/`, etc.) and even most of the `*_test.go` in the top-level package are as well.

brandur-stripe:
> I'm not sure how I feel about spawning a whole new goroutine for this, so I'd like to know your thoughts.

Nice idea!

I think though that you'd still have a potential race that might cause the new Goroutines to live longer than you'd want them to. These two lines are not guaranteed to run atomically:

``` go
<-s.prevRequestMetrics
s.prevRequestMetrics <- metrics
```

So after the ephemeral Goroutine pulls a value out of the channel, another Goroutine could sneak in and push its own value in, which would leave the ephemeral Goroutine waiting on `s.prevRequestMetrics <- metrics`.

My general feeling is that it's a little heavy-handed spinning up ephemeral Goroutines for this anyway.

I was just Googling and found [this article from Pivotal](https://content.pivotal.io/blog/a-channel-based-ring-buffer-in-go) on implementing a channel-based ring buffer. It's similar to what you suggested, but there's just a single Goroutine in charge of managing the ring, which avoids potential races. Honestly not sure if this is a good idea or not, but it seems pretty workable.

jameshageman-stripe:
Cool! Between the Pivotal channel-based ring buffer and using [container/ring](https://golang.org/pkg/container/ring/) with mutexes, I'd opt for the channel solution.

I'll try it out in this PR and we can decide if we like it better than non-circular buffers.

jameshageman-stripe:
The circular buffer solution (added in 7f09c4a) was passing the normal tests, but it was failing unexpectedly on the coverage tests. It would fail on this assertion: https://github.com/stripe/stripe-go/blob/7f09c4a5a5193ce2458328fefde11590ba1b7752/stripe_test.go#L252

I think this indicates that the telemetry pushed into the `inputChannel` by the first request had not yet been passed to the `outputChannel` by the ringBuffer's `run()` goroutine. 

However, this seems like it would be an unpredictable race condition based on the specific goroutine scheduling. Therefore I'm confused why it happened consistently on coverage tests but never on normal tests (in CI or on my laptop).

I'll try adding some extra logging to investigate.

jameshageman-stripe:
My guess was correct: there was a synchronization issue. My initial implementation of `ringBuffer` did asynchronous writes, so the first request's metrics hadn't necessarily been placed on the `outputChannel`  by the time the second request was kicked off (it all depended on go scheduling/time slicing).

I've added synchronous writes by having the client wait on a `done` channel, but I feel like this solution has become more complicated than simply wrapping the following in a mutex guard:
```go
<-s.prevRequestMetrics
s.prevRequestMetrics <- metrics
```

I think I'd prefer to go back to one a single channel solution and add a mutex.

jameshageman-stripe:
@brandur Alright, I think this is ready for re-review. Since you last commented, the following has changed:

- I implemented a circular buffer of `requestMetrics` using a single channel and a mutex.
- `EnableTelemetry` can be passed into `GetBackendWithConfig` via the `BackendConfig` struct, which allows individual clients to have telemetry enabled.
- I l(re-)added a global `stripe.EnableTelemetry` flag that will enable telemetry headers to all clients instead of a single `Backend`.
- I did not change the package in `stripe_test.go` from `stripe_test` to `stripe`, because that change would end up adding a large diff to an already large PR. Instead, I duplicated the `requestMetrics` and `requestTelemetry` headers into the one test that used them.

jameshageman-stripe:
@bobby-stripe @dcarney-stripe and I had some further discussion in person that I thought I'd share.

The circular buffer implementation has grown in complexity enough to warrant it's extraction from the `stripe.go` package, due to the required coordination between multiple channels, goroutines, and/or mutexes depending on the implementation. The complexity of implementing a safe, concurrent circular buffer led us to discuss the reasoning for using a circular buffer over a regular FIFO channel. Some points:
- With a circular buffer, the oldest request metrics get dropped when the buffer is full
- With a regular channel, the newest request metrics get dropped when the buffer is full
- The buffer only fills up when there are concurrently executing requests
- When we send a `requestMetrics` struct, it is already for a past request. The gap between recording a request metric and sending it could be on the order of minutes if the client is not sending many requests.
- We do not expect or require delivery of metrics for every request
- Using a circular buffer would reduce the average time between a request being recorded, but it does not stop clients from potentially sending `requestMetrics` that are many minutes old.


With these points in mind, I think it's worth reconsidering using a standard FIFO channel instead of a circular buffer, as the benefit (slightly younger request metrics on average) may not outweigh the added complexity.

brandur-stripe:
> With these points in mind, I think it's worth reconsidering using a standard FIFO channel instead of a circular buffer, as the benefit (slightly younger request metrics on average) may not outweigh the added complexity.

Yeah, fair enough.

I think the circular buffer would be slightly better implementation, but things change given that it's not that easy. I totally buy the complexity argument, and even with thorough review, there's still a much higher chance of introducing a bug with our own buffer implementation than just using a standard channel.

I'm +1 going back to FIFO channel if you guys are. Thanks for looking into it at least!

jameshageman-stripe:
Reverted back to a simple FIFO channel. 

r? @brandur-stripe

stripe.go
```diff
@@ -779,6 +835,9 @@ const minNetworkRetriesDelay = 500 * time.Millisecond
 
 const uploadsURL = "https://uploads.stripe.com"
 
+// The number of requestMetric objects to buffer for client telemetry.
+const telemetryBufferSize = 16
```
brandur-stripe: Can you move this up a couple lines? (The constant names are ordered alphabetically so let's try to keep them that way).

stripe.go
```diff
@@ -387,6 +426,23 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
+	if s.enableTelemetry {
+		reqID := res.Header.Get("Request-Id")
+		if len(reqID) > 0 {
+			metrics := requestMetrics{
+				RequestID:         reqID,
+				RequestDurationMS: requestDurationMS,
+			}
+
+			// If the metrics buffer is full, discard the new metrics. Otherwise, add
+			// them to the buffer.
```
brandur-stripe: Thanks for the comment!

stripe.go
```diff
@@ -138,6 +147,12 @@ type BackendConfig struct {
 	//
 	// If left empty, it'll be set to the default for the SupportedBackend.
 	URL string
+
+	// EnableTelemetry allows request metrics (request id and duration) to be sent
+	// to Stripe in subsequent requests via the `X-Stripe-Client-Telemetry` header.
+	//
+	// Defaults to false.
+	EnableTelemetry bool
```
brandur-stripe: Likewise, can you put `EnableTelemetry` into the right place alphabetically?

stripe.go
```diff
@@ -338,8 +371,14 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			}
 		}
 
+		// `requestStart` is used solely for client telemetry and, unlike `start`,
+		// does not account for the time spent building the request body.
+		requestStart := time.Now()
```
brandur-stripe: Can we just use `start`? All that's happening between the two is some structs being initialized locally. In practice it's likely to have a negligible effect on timing, and I don't think it's worth complicating the code over.

stripe.go
```diff
@@ -292,8 +310,23 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
+		select {
+		case metrics := <-s.requestMetricsBuffer:
+			metricsJSON, err := json.Marshal(&requestTelemetry{LastRequestMetrics: metrics})
+			if err == nil {
+				req.Header.Set("X-Stripe-Client-Telemetry", string(metricsJSON))
+			} else if s.LogLevel > 2 {
+				s.Logger.Printf("Unable to encode client telemetry: %s", err)
+			}
+		default:
+			// no metrics available, ignore.
```
dcarney: nit: might be worth amending this comment to point out that this `default` case needs to be here to enable a non-blocking receive on the channel.  Wouldn't want some over-zealous refactorer to remove it, thinking that it was an innocuous block.

stripe.go
dcarney: Also, would you mind amending the comment to mention that additional objects will be dropped if the buffer is full?  It'll make it extra clear what the lib does when the buffer is full (as opposed to other behaviors like growing the buffer, or sending metrics synchronously, etc.)

(I know you mention that at the site where the channel is actually written to, but I think it'd be helpful to have here as well).

stripe_test.go
```diff
@@ -165,6 +166,207 @@ func TestDo_RetryOnTimeout(t *testing.T) {
...
+	backend := stripe.GetBackendWithConfig(
+		stripe.APIBackend,
+		&stripe.BackendConfig{
+			LogLevel:          3,
+			MaxNetworkRetries: 0,
+			URL:               testServer.URL,
+		},
+	).(*stripe.BackendImplementation)
+
+	for i := 0; i < 2; i++ {
```
dcarney: It's initially unclear why this loop executes twice.  Can you add a quick comment that explains that it's because the telemetry comes from the _previous_ completed request/response?

stripe_test.go
```diff
@@ -165,6 +166,207 @@ func TestDo_RetryOnTimeout(t *testing.T) {
...
+// Test that telemetry metrics are not sent by default
+func TestDo_TelemetryDisabled(t *testing.T) {
+	type testServerResponse struct {
+		Message string `json:"message"`
+	}
+
+	message := "Hello, client."
+	requestNum := 0
+
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
dcarney: Great test. 👍  I _love_ Go's `httptest` package!

stripe_test.go
```diff
@@ -165,6 +166,207 @@ func TestDo_RetryOnTimeout(t *testing.T) {
...
+		requestNum++
+
+		telemetryStr := r.Header.Get("X-Stripe-Client-Telemetry")
+		switch requestNum {
```
dcarney: 👍

jameshageman-stripe:
@dcarney @brandur-stripe Thanks for your feedback! I think I've addressed all of the comments, and I just added another assertion that checks that we actually measure request duration.

Let me know if you see anything else you'd like to discuss :)

brandur-stripe:
Awesome — LGTM. Thanks for all the hard work here James!

brandur-stripe:
Released as 55.12.0.

PR #768, opened 2019-01-14
Add section in README on Go modules + simplify lead-in

brandur:
Modifies the intro of the README to add explicit instructions on how to
use stripe-go with and without Go modules, including commands that need
to be run and what imports should look like.

I've also removed the "versioning" section because I don't think this
information is that all that novel, and certainly isn't useful enough to
occupy prime real estate at the front of the file.

[Link to rendered version](https://github.com/stripe/stripe-go/blob/brandur-readme-modules/README.md).

r? @remi (Going to wait a little longer before merge just in case I have
to revert the Go modules PR, but this should be reviewable now.)


brandur-stripe:
Thanks!

PR #769, opened 2019-01-15
missing amount on the IssuingTransaction struct

mdestagnol:
The amount field of the issuing transaction object is described in the api documentation (https://stripe.com/docs/api/issuing/transactions), returned by the API, but not parsed by this library.


brandur-stripe:
Oops. Thanks for the correction!

brandur-stripe:
Released as 55.9.0.

PR #774, opened 2019-01-16
Revert Go module support

brandur:
Reverts:
https://github.com/stripe/stripe-go/pull/712

(And an earlier PR that added a `go.mod` file in the first place.)

Unfortunately, there are still some fairly bad incompatibilities between
Go modules and Dep for users still on the latter. For now, we're going
to revert Go module support because there doesn't seem to be any easy
way (i.e., anything that's not maintaining a fork) of doing a good job
of both. Projects can still use the Go module system, but will have to
pull this package in as a pre-module incompatible one.

We are hoping to eventually see basic module awareness merged into Dep:

https://github.com/golang/dep/pull/1963

If/when that gets done, we'll revert this revert after a small grace
period, and hopefully be back to a place where both packaging systems
are well supported.

I'll also modify the README to clarify the current situation in another
follow up PR.

r? @brandur ~~(WIP.)~~


brandur-stripe:
@SamWhited Does everything here look okay to you?

It's basically a revert of your patch with a couple additional changes:

* `go.mod` has been removed completely.
* I kept your change in removing `go.sum` from `.gitignore`.

Thanks!

.travis.yml
```diff
@@ -33,9 +33,9 @@ env:
     - STRIPE_MOCK_VERSION=0.40.0
 
 go:
-  - "1.9.x"
-  - "1.10.x"
-  - "1.11.x"
```
SamWhited: Probably worth testing against the latest patch release still

SamWhited:
One little nit, but otherwise LGTM. Thanks for all your hard work on this; sorry to waste so much of your time with a patch that ended up having to be reverted. I'll keep the dep issues in mind when pushing for other packages to update in the future.

.travis.yml
brandur-stripe: Oops — nice catch. Fixed!

brandur-stripe:
> One little nit, but otherwise LGTM. Thanks for all your hard work on this; sorry to waste so much of your time with a patch that ended up having to be reverted. I'll keep the dep issues in mind when pushing for other packages to update in the future.

No worries at all. I don't think the time and effort was wasted because hopefully before too long we'll just be reverting this back and it'll be back in mainline with probably very little changed.

Thanks for your hard work on this and the helpful support!

brandur-stripe:
Released as 55.10.0.

PR #775, opened 2019-01-16
Update README with information on Go modules

brandur:
Go module support will be reverted in #774. Here, we update the README
with some information about the status of Go modules in stripe-go, and
how to use the library with them for the time being.

r? @brandur ~~(WIP.)~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,23 +25,31 @@ import (
 
 ### Go Module Support
 
-The library supports Go's [experimental modules][modules]. Add stripe-go as a
-requirement in your `go.mod` along with its current version:
+The library currently *does not* ship with first-class support for Stripe
+modules. We put in support for it before, but ran into compatibility problems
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,23 +25,21 @@ import (
...
-)
-```
-
-Imports should contain the major version as a virtual component of each path:
-
-``` go
-import (
-    "github.com/stripe/stripe-go/v55"
-    "github.com/stripe/stripe-go/v55/customer"
+    github.com/stripe/stripe-go/v55
```
brandur-stripe: I don't know if this is right. The modules Wiki [has some information on this](https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-v2-package-that-has-not-opted-into-modules-what-does-incompatible-mean), but naturally provides no examples. Let me get #774 for in and just try it out to make sure I get the right invocation.

brandur-stripe:
Thanks!

PR #776, opened 2019-01-17
Add support for billing thresholds

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

This one is a bit complex as it's across multiple resources and no specific test was added for it so we need to make sure it matches OpenAPI.

cc @clintonb-stripe Could you double check that I added the right set of properties too and I'm not missing any? (Java only lists properties, not parameters).

Docs: https://stripe.com/docs/billing/subscriptions/metered-billing/thresholds
Internal PR with gate removal: #126799

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -25,12 +25,13 @@ type InvoiceBillingReason string
 
 // List of values that InvoiceBillingReason can take.
 const (
-	InvoiceBillingReasonManual             InvoiceBillingReason = "manual"
-	InvoiceBillingReasonSubscription       InvoiceBillingReason = "subscription"
-	InvoiceBillingReasonSubscriptionCreate InvoiceBillingReason = "subscription_create"
-	InvoiceBillingReasonSubscriptionCycle  InvoiceBillingReason = "subscription_cycle"
-	InvoiceBillingReasonSubscriptionUpdate InvoiceBillingReason = "subscription_update"
-	InvoiceBillingReasonUpcoming           InvoiceBillingReason = "upcoming"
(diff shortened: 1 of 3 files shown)


remi-stripe:
(like the other one, removing the approval to leave that to Brandur. Thanks for the review Clinton!)

invoice.go
```diff
@@ -217,6 +219,19 @@ type InvoiceCustomField struct {
 	Value string `json:"value"`
 }
 
+// InvoiceThresholdReason is a structure representing a reason for a billing threshold.
+type InvoiceThresholdReason struct {
+	AmountGte   int64                               `json:"amount_gte"`
```
brandur-stripe: I'm assuming gte = "greater than or equal". Can we upcase it in all instances to be `GTE`? Things like `ID`, `PDF`, or `URL` are pretty ubiquitous convention through the library.

invoice.go
```diff
@@ -217,6 +219,19 @@ type InvoiceCustomField struct {
...
+	ItemReasons []*InvoiceThresholdReasonItemReason `json:"item_reasons"`
+}
+
+// InvoiceThresholdReasonItemReason is a structure representing the line items that
+// triggered an invoice.
+type InvoiceThresholdReasonItemReason struct {
+	LineItemIds []string `json:"line_item_ids"`
```
brandur-stripe: And similarly here: `IDs`.

remi-stripe:
@brandur-stripe Fixed, PTAL (and thank you)

brandur-stripe:
LGTM!

brandur-stripe:
Released as 55.11.0.

PR #780, opened 2019-01-22
Detect custom OpenAPI spec file and start stripe-mock from test suite

brandur:
Very similar to the recent feature implemented in other languages (for
example [1]), this wraps test runs in a script which will detect a
custom OpenAPI spec and fixture in `testing/openapi/` and start a local
version of stripe-mock pointing to them. The test suite is then executed
against that stripe-mock and it's shut down afterwards.

If there isn't a custom OpenAPI spec, we continue to fall back to the
system stripe-mock to keep test runs as fast as possible.

This differs a little from implementations in other languages in that we
had to start stripe-mock from a script (as opposed to a `before` block
or the like) because Go's testing tools don't give us a place where we
can do anything before any test runs. Therefore, we start stripe-mock
then execute `go test ./...` as a subprocess.

r? @ob-stripe
cc @stripe/api-libraries

[1] https://github.com/stripe/stripe-ruby/pull/715

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@ lint:
 	golint -set_exit_status ./...
 
 test:
-	go test -race ./...
+	go run scripts/test_with_stripe_mock.go -race ./...
 
 vet:
 	go vet ./...
@@ -21,7 +21,7 @@ vet:
 coverage:
 	# go currently cannot create coverage profiles when testing multiple packages, so we test each package
 	# independently. This issue should be fixed in Go 1.10 (https://github.com/golang/go/issues/6909).
-	go list ./... | xargs -n1 -I {} -P 4 go test -covermode=count -coverprofile=../../../{}/profile.coverprofile {}
+	go list ./... | xargs -n1 -I {} -P 4 go run scripts/test_with_stripe_mock.go -covermode=count -coverprofile=../../../{}/profile.coverprofile {}
 
 clean:
 	find . -name \*.coverprofile -delete
--- a/scripts/test_with_stripe_mock.go
+++ b/scripts/test_with_stripe_mock.go
@@ -1,0 +1,172 @@
+// A script that wraps the run of the project test suite and starts stripe-mock
+// with a custom OpenAPI + fixtures bundle if one was found in the appropriate
+// spot (see `pathSpec` below).
+//
+// The script passes all its arguments to a Go's test command, and defaults to
+// `./...`. For example, both of the following are valid invocations:
+//
+//     go run test_with_stripe_mock.go
+//     go run test_with_stripe_mock.go ./charge
+//
+// The reason that we need a separate script for this is because Go's testing
+// infrastructure doesn't provide any kind of global hook that we can use to do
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -13,15 +13,15 @@ lint:
...
+	go run scripts/test_with_stripe_mock.go -race ./...
 
 vet:
 	go vet ./...
 
 coverage:
 	# go currently cannot create coverage profiles when testing multiple packages, so we test each package
 	# independently. This issue should be fixed in Go 1.10 (https://github.com/golang/go/issues/6909).
-	go list ./... | xargs -n1 -I {} -P 4 go test -covermode=count -coverprofile=../../../{}/profile.coverprofile {}
+	go list ./... | xargs -n1 -I {} -P 4 go run scripts/test_with_stripe_mock.go -covermode=count -coverprofile=../../../{}/profile.coverprofile {}
```
brandur-stripe: This is a little unfortunate because it means a new stripe-mock will get started with every invocation.

I think it's okay for now though because (1) doesn't affect most test runs (only those that use a custom spec) because we fall back on the default stripe-mock in most cases, (2) stripe-mock's start is relatively fast (in my tests the script adds 0.2-03 seconds of overhead), and (3) it should be plausible to tear this out pretty soon. When 1.12 comes out I'd probably suggest we just drop 1.9 and this hack.

(A motivated individual could also fix it now by plumbing some env vars into the matrix so that coverage only runs on >= 1.10, but it doesn't seem worth it right now to me).

scripts/test_with_stripe_mock.go
```diff
@@ -0,0 +1,172 @@
...
+
+	b := make([]byte, 1024)
+	var port string
+
+	// We store the entire captured output because the string we're looking for
+	// may have appeared across a read boundary.
+	var stdoutBuffer bytes.Buffer
+
+	for {
+		n, err := stdout.Read(b)
```
ob-stripe: Just curious, wouldn't it be simpler to use [`ReadString`](https://golang.org/pkg/bytes/#Buffer.ReadString) with `\n` as the delimiter and check each line for the port?

brandur-stripe: Good question! These IO constructs can all get a little confusing.

So the answer is that `stdout` here is not in fact a `bytes.Buffer`. `cmd.StdoutPipe` above returns an [`io.ReadCloser`](https://godoc.org/io#ReadCloser), which is a combination of a `Closer` and a `Reader`. [`Reader`](https://godoc.org/io#Reader) is an extremely simple interface that only provides us with one function: `Read`.

I think it's all stripped down this way because in this case Go has few guarantees around how that stdout pipe can be expected to behave. If you had a `ReadBytes`-type function on it, what would happen if the program never emitted another `\n`? It could either block forever, or you'd need to add a timeout mechanism.

ob-stripe: Makes sense! Thanks for the thorough answer.

brandur-stripe:
Thanks for the review Olivier!

PR #783, opened 2019-01-31
Allow stripe-mock to be suppressed with `STRIPE_MOCK_AUTOSTART=false`

brandur:
Allows stripe-mock startup to be disabled by passing
`STRIPE_MOCK_AUTOSTART` in the env with a value of `false` or `0`. This
may be useful for debugging problems where the automatically started
stripe-mock is not showing logging.

~~Allows stripe-mock startup to be disabled by passing
`STRIPE_MOCK_DISABLE` in the env with any value. This may be useful for
debugging problems where the automatically started stripe-mock is not
showing logging.~~

r? @ob-stripe Thoughts on this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ instructions for installing via Homebrew and other methods):
 
 Run all tests:
 
-    go test ./...
+    make test
 
 Run tests for one package:
 
--- a/scripts/test_with_stripe_mock.go
+++ b/scripts/test_with_stripe_mock.go
@@ -32,20 +32,30 @@ const (
 var portMatch = regexp.MustCompile(` port: (\d+)`)
 
 func main() {
-	var port string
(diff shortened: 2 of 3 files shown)


brandur-stripe:
Hm, good point.

`STRIPE_MOCK_NO_AUTOSTART` feels a little awkward to me. I like `STRIPE_MOCK_AUTOSTART=false` better, but admit that it may make usage a little harder because you have to remember the right value to pass (i.e., `=0` or `=false`). What do you think?

brandur-stripe:
Thanks OB. I tweaked this a bit to change to `STRIPE_MOCK_AUTOSTART`, which will accept either a value of `0` or `false` to be a little more forgiving for people with a bad memory (like me). I also did a little refactoring to make this code prettier and added a note to the README in `testing/openapi/`.

PR #784, opened 2019-02-03
Add support for Subscription Schedule resource

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/subschedule.go
+++ b/subschedule.go
@@ -1,0 +1,182 @@
+package stripe
+
+import (
+	"encoding/json"
+)
+
+// SubscriptionScheduleStatus is the list of allowed values for the schedule's status.
+type SubscriptionScheduleStatus string
+
+// List of values that SubscriptionScheduleStatus can take.
+const (
+	SubscriptionScheduleStatusActive    SubscriptionScheduleStatus = "active"
+	SubscriptionScheduleStatusCanceled  SubscriptionScheduleStatus = "canceled"
+	SubscriptionScheduleStatusCompleted SubscriptionScheduleStatus = "completed"
+	SubscriptionScheduleStatusPastDue   SubscriptionScheduleStatus = "not_started"
+	SubscriptionScheduleStatusTrialing  SubscriptionScheduleStatus = "released"
+)
+
+// SubscriptionScheduleRenewalBehavior describe what happens to a schedule when it ends.
+type SubscriptionScheduleRenewalBehavior string
+
+// List of values that SubscriptionScheduleRenewalBehavior can take.
+const (
+	SubscriptionScheduleRenewalBehaviorNone    SubscriptionScheduleRenewalBehavior = "none"
+	SubscriptionScheduleRenewalBehaviorRenew   SubscriptionScheduleRenewalBehavior = "release"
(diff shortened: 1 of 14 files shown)


remi-stripe:
This is now ready to be released as stripe-mock has been updated. The feature is gated but accessible for most merchants in the Dashboard and we want the libraries to support it.

r? @brandur-stripe

subschedule.go
```diff
@@ -0,0 +1,168 @@
...
+// subscription schedule.
+type SubscriptionScheduleParams struct {
+	Params            `form:"*"`
+	Billing           *string                              `form:"billing"`
+	BillingThresholds *SubscriptionBillingThresholdsParams `form:"billing_thresholds"`
+	Customer          *string                              `form:"customer"`
+	FromSubscription  *string                              `form:"from_subscription"`
+	DaysUntilDue      *int64                               `form:"days_until_due"`
+	Phases            []*SubscriptionSchedulePhaseParams   `form:"phases"`
+	RenewalBehavior   *string                              `form:"renewal_behavior"`
```
brandur-stripe: Just flagging that the docs I'm looking at don't have `DaysUntilDue`, and they `EndDate`, which seems to be missing here.

subschedulerevision.go
```diff
@@ -0,0 +1,57 @@
...
+
+// SubscriptionScheduleRevision is the resource representing a Stripe subscription schedule revision.
+type SubscriptionScheduleRevision struct {
+	Created          int64                                `json:"created"`
+	ID               string                               `json:"id"`
+	Livemode         bool                                 `json:"livemode"`
+	Metadata         map[string]string                    `json:"metadata"`
+	Object           string                               `json:"object"`
+	Phases           []*SubscriptionSchedulePhase         `json:"phases"`
+	PreviousRevision *SubscriptionSchedule                `json:"previous_revision"`
```
brandur-stripe: Is this expandable? The API ref docs are showing that it's just a string.

remi-stripe: @brandur-stripe It's not expandable but if it became expandable it'd be a breaking change so I preferred to do this. It might be the wrong approach though. What do you think is better?

subschedule.go
remi-stripe: Damn nice catch. ~daysUntilDue` should be in a separate struct/hash. I fixed this.

Can you clarify what you meant about EndDate? This should only be in the `current_phase` or `phases[]` instead

subschedulerevision.go
brandur-stripe: Yeah, that works for me. Just wanted to make sure you did it on purpose.

subschedule.go
brandur-stripe: Arg, sorry — I mixed up a word, what I meant is that it looks like there should be a `StartDate`.

All I have is an internal link, but if you scroll to "create a schedule" and look near the bottom, you should see a `start_date` underneath `renewal_behavior`.

brandur-stripe:
Thanks for the fix!

ptal @remi-stripe

subschedulerevision.go
remi-stripe: Argl yeah I remember they added this after I started the implementation. Fixing now.

remi-stripe:
@brandur-stripe Okay fixed!

brandur-stripe:
LGTM!

PR #787, opened 2019-02-12
Add support for `transfer_data[amount]` on Charge

remi-stripe:
We decided to support `transfer_data[amount]` on Charges for 2/19, at least for a few months. We think developers should use `application_fee_amount` instead but many integrations already used `destination[amount]` and would be quite blocked if we removed this.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
LGTM.

PR #789, opened 2019-02-12
Allow API arrays to be emptied by setting an empty array

### LINKED ISSUE #788: Unable to remove product attributes
I am working on a [Terraform provider](https://github.com/franckverrot/terraform-provider-stripe/pull/3), and am unable to delete product attributes. My test below demonstrates this.

Is this a bug in the client library, or am is my implementation incorrect (totally feasible given this is my first foray into Go)?

Go version: `go version go1.11.1 darwin/amd64`
Client version: `55.14.0`

```go
import (
	"github.com/stripe/stripe-go"
	"github.com/stripe/stripe-go/client"
	"log"
	"testing"
)

func TestAttributeRemoval(t *testing.T) {
	sc := &client.API{}
	sc.Init("REPLACE-ME", nil)

	tmp := "foo"
	attributes := []*string{&tmp}


	params := stripe.ProductParams{
		Name: stripe.String("Attribute Test"),
		Type: stripe.String("good"),
		Attributes: attributes,
	}

	product, err := sc.Products.New(&params)

	if err != nil {
		t.Errorf("Product creation failed: %v", err)
	}

	params = stripe.ProductParams{
		Attributes: []*string{},
	}
	product, err = sc.Products.Update(product.ID, &params)

	if err != nil {
		t.Errorf("Product update failed: %v", err)
	}

	log.Printf("%v", product)

	if len(product.Attributes) != 0 {
		t.Errorf("Attributes were not deleted: %v", product.Attributes)
	}
}
```

brandur:
Allows API arrays like `additional_owners` to be emptied by setting
their value to an explicit empty array. To accomplish this we need to
rework parts of the form-encoding library to detect this specific
condition.

I also had to rework a fair bit of the form package's test suite because
the existing code was making no distinction between an actual value
being set an empty in the generated form versus one that's not present.

Fixes #788.

r? @remi-stripe

### DIFF
--- a/form/form_test.go
+++ b/form/form_test.go
@@ -251,13 +251,16 @@ func TestAppendTo(t *testing.T) {
 		},
 
 		{"slice[2]", &testStruct{Slice: sliceVal}, stringPtr("3")},
-		{"slice", &testStruct{Slice: []string{}}, nil},
+		{"slice", &testStruct{Slice: []string{}}, stringPtr("")},
+		{"slice", &testStruct{Slice: nil}, nil},
 
 		{"slice_ptr[2]", &testStruct{SlicePtr: &sliceVal}, stringPtr("3")},
 
 		// A slice pointer given an explicit but empty slice should encode to
 		// an empty string (this tells the Stripe API to "zero" the array).
 		{"slice_ptr", &testStruct{SlicePtr: &sliceVal0}, stringPtr("")},
+
+		{"slice_ptr", &testStruct{SlicePtr: nil}, nil},
 
 		{"string", &testStruct{String: stringVal}, &stringVal},
 		{"string_ptr", &testStruct{StringPtr: &stringVal}, &stringVal},
--- a/form/form_test.go
+++ b/form/form_test.go
@@ -1,6 +1,7 @@
 package form
 
 import (
+	"fmt"
 	"net/url"
 	"sync"
 	"testing"
@@ -118,6 +119,7 @@ func BenchmarkAppendTo(b *testing.B) {
 
 func TestAppendTo(t *testing.T) {
 	var arrayVal = [3]string{"1", "2", "3"}
+	var arrayVal0 = [3]string{}
 
 	var boolValT = true
 	var boolValF = false
@@ -140,6 +142,7 @@ func TestAppendTo(t *testing.T) {
 	var int64Val0 int64
 
 	var sliceVal = []string{"1", "2", "3"}
+	var sliceVal0 = []string{}
 
 	var stringVal = "123"
 	var stringVal0 = ""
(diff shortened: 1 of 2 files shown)


form/form_test.go
```diff
@@ -164,57 +167,64 @@ func TestAppendTo(t *testing.T) {
...
-		{"int16_ptr", &testStruct{}, ""},
-		{"int32", &testStruct{Int32: int32Val}, "123"},
-		{"int32_ptr", &testStruct{Int32Ptr: &int32Val}, "123"},
-		{"int32_ptr", &testStruct{Int32Ptr: &int32Val0}, "0"},
-		{"int32_ptr", &testStruct{}, ""},
-		{"int64", &testStruct{Int64: int64Val}, "123"},
-		{"int64_ptr", &testStruct{Int64Ptr: &int64Val}, "123"},
-		{"int64_ptr", &testStruct{Int64Ptr: &int64Val0}, "0"},
-		{"int64_ptr", &testStruct{}, ""},
+		{"appender", &testStruct{Appender: &testAppender{String: "123"}}, stringPtr("123")},
```
brandur-stripe: As mentioned in the PR description, I had to rework these test cases so that we make a meaningful distinction between a string value that's present _but empty_ versus one that's not present at all.

This is important for testing cases like our zeroed slices where the resulting output should look like `additional_owners=`.

brandur-stripe:
Also note that I left in `AdditionalOwnersEmpty` for now. This should open up a new option for removing its values, but I figure that we may as well stay API incompatible for now.

form/form_test.go
remi-stripe: Since `additional_owners` is going away next week and already has a special case `AdditionalOwnersEmpty` we likely want to avoid changing that one. For that reason, I'd recommend shipping this in a major version (just in case we changed some behaviour we missed) since we're shipping one today.

form/form_test.go
```diff
@@ -414,3 +454,13 @@ func TestValues(t *testing.T) {
...
 	assert.Nil(t, values.Get("boguskey"))
 }
+
+//
+// Private functions
+//
+
+// Trivial helper to get is a string pointer from a string (because you're not
+// allowed to take the address of a literal directly).
+func stringPtr(s string) *string {
```
remi-stripe: Why no use `stripe.String()` instead? Is it because form is "deeper" and should rely on a Stripe method?

brandur-stripe: Yeah, my justification is that essentially the `form` package should be considered kind of a utility that's at a lower level than anything Stripe-related. I wouldn't mind using `stripe.String` that much, but I figure that this implementation keeps boundaries a little cleaner, and the function is so trivial that it's never going to change or anything.

brandur-stripe:
@remi-stripe 

> This looks good to me but I'm having trouble finding the explicit test about empty arrays in all the changes. I expected to see something similar to what @ob-stripe did in dotnet where you set a specific params struct and encode it and see the output has what you would expect such as `xxxx=` as the output for an empty, non-nil slice.

Should have pointed this out (it's hard to see amongst all the churn), but I did add some new test cases in. Here's the one in particular that checks what we care about:

``` go
		{"slice_ptr", &testStruct{SlicePtr: &sliceVal0}, stringPtr("")},
```

Basically we're saying that: given a slice pointer and an explicit but empty slice value, we expect to see an empty string encoded. I just added a comment on there because it's not super obvious from reading through the code that this case is particularly important.

> Since `additional_owners` is going away next week and already has a special case `AdditionalOwnersEmpty` we likely want to avoid changing that one. For that reason, I'd recommend shipping this in a major version (just in case we changed some behaviour we missed) since we're shipping one today.

Okay, no string opinion there.

clintonb-stripe:
My test at https://github.com/stripe/stripe-go/issues/788 still fails. Is this supposed to work with `[]*string{}`?

remi-stripe:
I can confirm it fails for me too and we don't seem to go in that `if` that handles the empty slices.

brandur-stripe:
Yeah, I confirmed that it doesn't fix this particular issue as well. I'll need to spend a little more time on this one to cover `[]*string` in addition to `*[]string`.

brandur-stripe:
Okay! I've update the code to also handle the case of a non-pointer slice. I ran Clinton's repro and confirmed that it now passes as expected.

@remi-stripe Would you mind reviewing again? I pushed the new changes as a separate commit so that it's easier to see what changed. Thanks!

brandur-stripe:
Released as 56.0.0.

PR #782, opened 2019-01-31
Add support for API version 2019-02-19

remi-stripe:
This PR contains the changes related to the new API version `2019-02-19`:
* Move stripe-go to API version `2019-02-19`
* Change the shape of `Account`
  * The `legal_entity` property on the Account API resource has been replaced with `individual`, `company`, and `business_type`.
  * `LegalEntityType` is now `BusinessType`.
  * The verification hash has been replaced with a requirements hash.
  * Multiple top-level properties were moved to the `settings` hash.
  * The `keys` property on `Account` has been removed. Platforms should authenticate as their connected accounts with their own key via the `Stripe-Account` [header](https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header).
* The `requested_capabilities` property on `Account` creation is now required for accounts in the US.
* Removed deprecated parameter `save_source_to_customer` on `PaymentIntent`. Use `save_payment_method` instead.
r? @brandur-stripe 
cc @stripe/api-libraries

PR #794, opened 2019-02-20
Fix API version which should be pinned to 2019-02-19 instead

remi-stripe:
Something went wrong on https://github.com/stripe/stripe-go/pull/782 and the API version was never upgraded, making that version broken unfortunately.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
Oops. Retroactive LGTM.

PR #796, opened 2019-02-22
Make invoice items a slice, as per API spec

EVODelavega:
Quick fix for the package, seems to be what is required for this particular call to work

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -99,18 +99,18 @@ type InvoiceParams struct {
 
 	// These are all for exclusive use by GetNext.
 
-	Coupon                         *string                           `form:"coupon"`
-	InvoiceItems                   *InvoiceUpcomingInvoiceItemParams `form:"invoice_items"`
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Oh shoot. Thank you @EVODelavega!

@remi-stripe Thoughts on bumping this as minor? It probably didn't work before anyway (and definitely didn't work correctly even if it technically ran).

remi-stripe:
@brandur-stripe Yes I think it makes sense here to bump as a minor and call it out in the changelog!

EVODelavega:
@brandur-stripe Yeah, would really appreciate this fix to land ASAP. Previewing an upcoming invoice and adding an invoice item (only 1, because that's all you could add) results in a 400 response. I've tried the same request with this change, everything worked like a charm

remi-stripe:
Can I let you release this one to unblock @EVODelavega?

brandur-stripe:
@remi-stripe Yep, I got it.

brandur-stripe:
Released as 57.1.0.

Issue #797, opened 2019-02-23
Stripe events have the object property serialized twice


sudhirj:
I'm serializing all the events coming in for audit purposes, and the `object` property is being represented in all serializations twice. 

In https://github.com/stripe/stripe-go/blob/master/event.go#L36
```
// EventData is the unmarshalled object as a map.
type EventData struct {
	Object             map[string]interface{}
	PreviousAttributes map[string]interface{} `json:"previous_attributes"`
	Raw                json.RawMessage        `json:"object"`
}
```
The `Raw` Go property is serialized once in the JSON as `object`, and the `Object` Go property is serialized again as `Object` in the JSON. Ideally, the `Object` property should have a `json:"-"` annotation.

remi-stripe:
@sudhirj Thanks a lot for the report! I opened a PR for this and we'll get this released likely next week!

EVODelavega:
@sudhirj A while back I started work on something to avoid having to manually map that `map[string]interface{}` type onto the correct object. I ended up hacking out something to generate code based on the docs. We're currently using that approach, but maybe a similar approach could be adopted in the stripe-go package as part of this change? I've included a link to gist in the last comment here: https://github.com/stripe/stripe-go/pull/729

Just add a cusom marshaller to the template/type check for nil pointers, and set the `object` field accordingly. Something like this would do:

```go
// add `json:"-"` to all fields, or `json:"fieldname,omitempty"`
func (e Event) MarshalJSON() ([]byte, error) {
    type mEvent struct {
        stripe.Event // embed the original event
        Object interface{} `json:"object"
    }
    me := mEvent{
        Event: e.Event, // set stripe.Event
     }
    if e.SKU != nil {
        me.Object = e.SKU
    } else if e.Order != nil {
        me.Object = e.Order
    } 
    // and so on
    return json.Marshal(me)
}
```

Just like the switch in the custom `UnmarshalJSON` func, you can generate the custom marshaller. Should be pretty straightforward

remi-stripe:
@sudhirj Thanks for the patience here. We just released a fix in 57.4.0

PR #798, opened 2019-02-26
Prevents `Object` from being serialized on the `Event` resource

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/797

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/event.go
+++ b/event.go
@@ -33,7 +33,10 @@ type EventRequest struct {
 
 // EventData is the unmarshalled object as a map.
 type EventData struct {
-	Object             map[string]interface{}
+	// Object is a raw mapping of the API resource contained in the event.
+	// Although marked with json:"-", it's still populated independently by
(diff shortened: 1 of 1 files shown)


event.go
```diff
@@ -33,7 +33,7 @@ type EventRequest struct {
 
 // EventData is the unmarshalled object as a map.
 type EventData struct {
-	Object             map[string]interface{}
+	Object             map[string]interface{} `json:"-"`
```
brandur-stripe: Do you think that we could add a comment here that explains that despite the `json:"-"`, this property will be populated independently? Maybe something like:

> Object is a raw mapping of the API resource contained in the event.
>
> Although marked with `json:"-"`, it's still populated independently by a custom `UnmarshalJSON` implementation.

remi-stripe:
@brandur thanks for the recommendation. Added your proposal. PTAL

Issue #804, opened 2019-03-05
Should we be setting MaxNetworkRetries in production?


matthewmueller:
Hi there,

I'm looking through the source and noticed that `MaxNetworkRetries` defaults to 0. If I'm not mistaken, that's also the default when you do `client.New(key, nil)`. I haven't run into an issue with this yet, but it seems like an issue waiting to happen.

Should we be setting this value in production to something other than 0? And if so, would it make sense for this library to have a non-zero default here?

Thanks!

remi-stripe:
@matthewmueller The idea with this feature is that it's off by default (we don't retry) but configurable if you want to do automatic retries in your project. We set the default to 0 in our other libraries too.

It seems we never documented this in the readme so I'll work on this but the idea is that you can configure this value so that the library automatically retries requests that fail because of a network issue.

brandur-stripe:
(Hah, Remi beat me to the punch, but I was 80% through writing this answer, so I'm just going to post it.)

In general you should be able to set this to a non-zero value in production with no harmful side effects.

The one thing you might notice is that with retries on, the latency around your Stripe API calls may become less deterministic as each call into the API library may actually be making multiple requests and sleeping between them. For most integrations I wouldn't expect this to be a problem though.

In terms of the zero default: I think this is mostly a historical artifact of the feature being relatively experimental when it was first implemented, so we wanted to have the safest possible default. Since then, it's been pretty widely vetted, but we just never revisited.

matthewmueller:
Awesome! Thanks guys for the information and the note in the readme is very helpful!

PR #805, opened 2019-03-05
Add README section on configurable retries

brandur:
Adds a new README subsection that describes how to enable automatic
retries for network or other non-deterministic problems. This reflects
information that's available in other client libraries like stripe-ruby.

As discussed in #804.

r? @remi-stripe


brandur-stripe:
Thanks Remi!

PR #806, opened 2019-03-06
Add support for PaymentMethod

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -21,6 +21,30 @@ const (
 	ChargeFraudStripeReportFraudulent ChargeFraudStripeReport = "fraudulent"
 )
 
+// ChargePaymentMethodDetailsType is the type of the PaymentMethod associated with the Charge's
+// payment method details.
+type ChargePaymentMethodDetailsType string
+
+// List of values that ChargeFraudUserReport can take.
+const (
+	ChargePaymentMethodDetailsTypeAchCreditTransfer ChargePaymentMethodDetailsType = "ach_credit_transfer"
+	ChargePaymentMethodDetailsTypeAchDebit          ChargePaymentMethodDetailsType = "ach_debit"
+	ChargePaymentMethodDetailsTypeAlipay            ChargePaymentMethodDetailsType = "alipay"
+	ChargePaymentMethodDetailsTypeBancontact        ChargePaymentMethodDetailsType = "bancontact"
+	ChargePaymentMethodDetailsTypeBitcoin           ChargePaymentMethodDetailsType = "bitcoin" // This is unsupported today and is here for legacy charges.
+	ChargePaymentMethodDetailsTypeCard              ChargePaymentMethodDetailsType = "card"
+	ChargePaymentMethodDetailsTypeCardPresent       ChargePaymentMethodDetailsType = "card_present"
+	ChargePaymentMethodDetailsTypeEps               ChargePaymentMethodDetailsType = "eps"
+	ChargePaymentMethodDetailsTypeGiropay           ChargePaymentMethodDetailsType = "giropay"
+	ChargePaymentMethodDetailsTypeIdeal             ChargePaymentMethodDetailsType = "ideal"
+	ChargePaymentMethodDetailsTypeMultibanco        ChargePaymentMethodDetailsType = "multibanco"
+	ChargePaymentMethodDetailsTypeP24               ChargePaymentMethodDetailsType = "p24"
+	ChargePaymentMethodDetailsTypeSepaDebit         ChargePaymentMethodDetailsType = "sepa_debit"
+	ChargePaymentMethodDetailsTypeSofort            ChargePaymentMethodDetailsType = "sofort"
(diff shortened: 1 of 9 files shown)


charge.go
```diff
@@ -144,6 +167,194 @@ type ChargeLevel3 struct {
...
+}
+
+// ChargePaymentMethodDetailsCardThreeDSecure represents details about 3DS associated with the
+// charge's PaymentMethod.
+type ChargePaymentMethodDetailsCardThreeDSecure struct {
+}
+
+// PaymentMethodDetailsCardWallet represents the details of the card wallet if this Card
+// PaymentMethod is part of a card wallet.
+type PaymentMethodDetailsCardWallet struct {
```
mickjermsurawong-stripe: @remi-stripe  should this be prefixed with `ChargeXXX` like other structs too?

remi-stripe: Yes it should, nice catch.

paymentmethod.go
```diff
@@ -0,0 +1,185 @@
...
+	PaymentMethodCardWalletTypeGooglePay           PaymentMethodCardWalletType = "google_pay"
+	PaymentMethodCardWalletTypeMasterpass          PaymentMethodCardWalletType = "masterpass"
+	PaymentMethodCardWalletTypeSamsungPay          PaymentMethodCardWalletType = "samsung_pay"
+	PaymentMethodCardWalletTypeVisaCheckout        PaymentMethodCardWalletType = "visa_checkout"
+)
+
+// PaymentMethodTransferDataParams is the set of parameters allowed for the transfer_data hash.
+type PaymentMethodTransferDataParams struct {
+	Destination *string `form:"destination"`
+}
```
garrett-stripe: I think this is unintentional—this struct doesn't appear to be used anywhere.

remi-stripe: ah yes bad copy paste, fixed (won't push while you review)

paymentintent.go
```diff
@@ -114,14 +114,16 @@ type PaymentIntentListParams struct {
 
 // PaymentIntentLastPaymentError represents the last error happening on a payment intent.
 type PaymentIntentLastPaymentError struct {
```
remi-stripe: Technically, this should be `Error` right? Is it worth changing this and releasing as a major version to be cleaner?

remi-stripe:
Okay, this should now be feature complete and ready for review.

There is one test failing, `TestTokenNew_SharedCustomerBankAccount` and it's because of a change in openapi which does not expect `bank_account` to be a string even though it can be. cc @mickjermsurawong-stripe as this is likely due to a change in our codebase for stripe-java autogeneration.

r? @brandur-stripe

remi-stripe:
@mickjermsurawong-stripe Thanks a lot for the thorough review. I can't believe after all the back and forth I did on those PRs I still missed the wallet fields on Charge. I fixed everything you found I think. PTAL!

mickjermsurawong-stripe:
thanks remi for the fix!
LGTM

mickjermsurawong-stripe:
ah the test is failing on `/v1/tokens` as it is trying to create bank account token with a token. 
My change on typed param work requires only the bank_account object instead of hash. 
But the failing test here is a valid use case we want to support? I can add this back to support string token as well. @remi-stripe 

```
func TestTokenNew_SharedCustomerBankAccount(t *testing.T) {
	params := &stripe.TokenParams{
		BankAccount: &stripe.BankAccountParams{
			ID: stripe.String("ba_123"),
		},
		Customer: stripe.String("cus_123"),
	}
	params.SetStripeAccount("acct_123")
	token, err := New(params)
	assert.Nil(t, err)
	assert.NotNil(t, token)
}
```

remi-stripe:
You can ignore that failing test, it's being fixed in another PR in parallel I just didn't want to duplicate the work

paymentmethod.go
```diff
@@ -0,0 +1,183 @@
...
+// List of values that PaymentMethodCardBrand can take.
+const (
+	PaymentMethodCardBrandAmex       PaymentMethodCardBrand = "amex"
+	PaymentMethodCardBrandDiners     PaymentMethodCardBrand = "diners"
+	PaymentMethodCardBrandDiscover   PaymentMethodCardBrand = "discover"
+	PaymentMethodCardBrandJCB        PaymentMethodCardBrand = "jcb"
+	PaymentMethodCardBrandMastercard PaymentMethodCardBrand = "mastercard"
+	PaymentMethodCardBrandUnionpay   PaymentMethodCardBrand = "unionpay"
+	PaymentMethodCardBrandVisa       PaymentMethodCardBrand = "visa"
+	PaymentMethodCardBrandUnknown    PaymentMethodCardBrand = "unknown"
```
brandur-stripe: Minor alphabetization problem here: unknown should be above visa.

remi-stripe: Nice catch, fixed

remi-stripe:
I had to remove a test because we don't support `bank_account: "ba_123"` when creating tokens anymore. cc @mickjermsurawong-stripe

mickjermsurawong-stripe:
LGTM

PR #808, opened 2019-03-07
Add new fields to Subscription

remi-stripe:
Add support for `backdate_start_date` and `cancel_at` on `Subscription`. This is gated but many users have access to it and I missed it while releasing `SubscriptionSchedule`.

r? @brandur-stripe
@stripe/api-libraries

### DIFF
--- a/sub.go
+++ b/sub.go
@@ -38,11 +38,13 @@ type SubscriptionTransferDataParams struct {
 type SubscriptionParams struct {
 	Params                      `form:"*"`
 	ApplicationFeePercent       *float64                             `form:"application_fee_percent"`
+	BackdateStartDate           *int64                               `form:"backdate_start_date"`
 	Billing                     *string                              `form:"billing"`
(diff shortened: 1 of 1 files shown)


sub.go
```diff
@@ -134,6 +136,8 @@ type Subscription struct {
 	Billing               SubscriptionBilling            `json:"billing"`
 	BillingCycleAnchor    int64                          `json:"billing_cycle_anchor"`
 	BillingThresholds     *SubscriptionBillingThresholds `json:"billing_thresholds"`
+	CancelAt              int64                          `json:"cancel_at"`
+	CancelAtPeriodEnd     bool                           `json:"cancel_at_period_end"`
```
remi-stripe: that one was not in the right alphabetical order :p

brandur-stripe: Haha, good catch!

remi-stripe: Truthfully, I did not catch it. I noticed the field was missing, added it, looked at the original PR and then figured out what happened :p

brandur-stripe:
LGTM.

PR #810, opened 2019-03-14
Add support for columns on Reporting resources

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

Related to https://github.com/stripe/stripe-java/pull/689


brandur-stripe:
Could find `columns` in either admin docs or even in `master` checkout as of this morning, but LGTM if you're pretty sure those are in the right place.

ptal @remi-stripe

remi-stripe:
They are not in the docs but it shipped today and it's in openapi! Thanks for the fast review.

PR #811, opened 2019-03-15
Add support for API version 2019-03-14

remi-stripe:
The changes in the library itself look minor. But this API version changes the behavior of Subscriptions entirely under the hood. If the charge fails, you still get a Subscription back instead of an exception and it's on you to check the Subscription status.

This is now fully ready to be reviewed. It also changes the `apiversion` internal-only constant to a public one.

r? @brandur-stripe 
cc @stripe/api-libraries 

Related to https://github.com/stripe/stripe-dotnet/pull/1552

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
       tar -zxf "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}_linux_amd64.tar.gz" -C "stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/"
     fi
   - |
-    stripe-mock/stripe-mock_${STRIPE_MOCK_VERSION}/stripe-mock -https-port 12112 > /dev/null &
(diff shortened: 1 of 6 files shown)


brandur-stripe:
> The changes in the library itself look minor. But this API version changes the behavior of Subscriptions entirely under the hood. If the charge fails, you still get a Subscription back instead of an exception and it's on you to check the Subscription status.

Oh, very good to know! Thanks.

LGTM.

remi-stripe:
@brandur-stripe I ended up finding a bug in the library which I fixed in https://github.com/stripe/stripe-go/pull/814. After that one is merged, we can properly force the new stripe-mock flag to test the API version in the library is the correct one.

Because of this, `apiVersion` now has to be exported so I renamed it. 

PTAL

brandur-stripe:
> Because of this, `apiVersion` now has to be exported so I renamed it. 

IMO, this is okay. It may be useful to print informational logs or something in client applications anyway. As long as it stays as a `const`, it can't be abused (by people overwriting it to a new value, etc.).

brandur-stripe:
Released as 58.0.0.

PR #812, opened 2019-03-16
Add support for Terminal Location and Reader deletion

remi-stripe:
cc @stripe/api-libraries

### DIFF
--- a/terminal/location/client.go
+++ b/terminal/location/client.go
@@ -52,6 +52,19 @@ func (c Client) Update(id string, params *stripe.TerminalLocationParams) (*strip
 	return location, err
 }
 
+// Del removes a location.
+func Del(id string, params *stripe.TerminalLocationParams) (*stripe.TerminalLocation, error) {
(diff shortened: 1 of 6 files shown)


terminal/location/client.go
```diff
@@ -52,6 +52,19 @@ func (c Client) Update(id string, params *stripe.TerminalLocationParams) (*strip
 	return location, err
 }
 
+// Del removes a location.
+func Del(id string, params *stripe.TerminalLocationParams) (*stripe.TerminalLocation, error) {
```
mickjermsurawong-stripe: I assume we are okay with sharing Params here having fields more than what apirefs now has..

remi-stripe: For now we're sharing in go for some of those. So I'd rather stay consistent

mickjermsurawong-stripe:
I"m not familiar with go, but I compared with SKU del, and this looks correct to me.
Schema is correct.
LGTM

PR #818, opened 2019-03-21
Implement leveled logging

### LINKED ISSUE #813: Feature request: Leveled logging
It would be useful to pipe stripe-go's internal log levels into our logging framework (e.g. [logrus](https://github.com/sirupsen/logrus/), e.g. [zap](https://go.uber.org/zap)).

This way, we could keep diagnostic level logs turned on in production, but monitor error-level logs more closely.

I think there is a backwards compatibility problem here that could be solved by adding a new, leveled log interface in addition to the existing `Printfer`. For example, it could be a new interface `Logfer`.

brandur:
Implements leveled logging for the Stripe library which has two main
benefits:

1. Greatly cleans up our logging call sites and makes them more
   readable.
2. Allows standard loggers seen in the Go ecosystem to be easily plugged
   into the library. [Logrus][logrus] for example.

See the additions in the README for usage information and other details.

Fixes #813.

r? @brandur (WIP.)

[logrus]: https://github.com/sirupsen/logrus/

### DIFF
--- a/log.go
+++ b/log.go
@@ -1,0 +1,221 @@
+package stripe
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"os"
+)
+
+//
+// Public constants
+//
+
+const (
+	// LevelError sets a logger to show error messages only.
+	LevelError Level = 1
+
+	// LevelWarn sets a logger to show warning messages or anything more
+	// severe.
+	LevelWarn Level = 2
+
+	// LevelInfo sets a logger to show informational messages or anything more
+	// severe.
+	LevelInfo Level = 3
+
+	// LevelDebug sets a logger to show informational messages or anything more
+	// severe.
+	LevelDebug Level = 4
+
+	// Older deprecated levels for Printfer-style logging.
+	printferLevelError = 1
+	printferLevelInfo  = 2
+	printferLevelDebug = 3
+)
+
+//
+// Public variables
+//
+
+// DefaultLeveledLogger is the default logger that the library will use to log
+// errors, warnings, and informational messages.
+//
+// LeveledLoggerInterface is implemented by LeveledLogger, and one can be
+// initialized at the desired level of logging.  LeveledLoggerInterface also
+// provides out-of-the-box compatibility with a Logrus Logger, but may require
+// a thin shim for use with other logging libraries that use less standard
+// conventions like Zap.
+//
+// This Logger will be inherited by any backends created by default, but will
+// be overridden if a backend is created with GetBackendWithConfig with a
+// custom LeveledLogger set.
+var DefaultLeveledLogger LeveledLoggerInterface
+
+// LogLevel is the logging level for this library.
+// 0: no logging
+// 1: errors only
+// 2: errors + informational (default)
+// 3: errors + informational + debug
+//
+// Deprecated: Logging should be configured with DefaultLeveledLogger instead.
+var LogLevel = 2
+
+// Logger controls how stripe performs logging at a package level. It is useful
+// to customise if you need it prefixed for your application to meet other
+// requirements.
+//
+// This Logger will be inherited by any backends created by default, but will
+// be overridden if a backend is created with GetBackendWithConfig with a
+// custom Logger set.
+//
+// Deprecated: Logging should be configured with DefaultLeveledLogger instead.
+var Logger Printfer
+
+//
+// Public types
+//
+
+// Level represents a logging level.
+type Level uint32
+
+// LeveledLogger is a leveled logger implementation.
+//
+// It prints warnings and errors to `os.Stderr` and other messages to
+// `os.Stdout`.
+type LeveledLogger struct {
+	// Level is the minimum logging level that will be emitted by this logger.
+	//
+	// For example, a Level set to LevelWarn will emit warnings and errors, but
+	// not informational or debug messages.
+	//
+	// Always set this with a constant like LevelWarn because the individual
+	// values are not guaranteed to be stable.
+	Level Level
+
+	// Internal testing use only.
+	stderrOverride io.Writer
+	stdoutOverride io.Writer
+}
+
+// Debugf logs a debug message using Printf conventions.
+func (l *LeveledLogger) Debugf(format string, v ...interface{}) {
+	if l.Level >= LevelDebug {
+		fmt.Fprintf(l.stdout(), format+"\n", v...)
+	}
+}
+
+// Errorf logs a warning message using Printf conventions.
+func (l *LeveledLogger) Errorf(format string, v ...interface{}) {
+	// Infof logs a debug message using Printf conventions.
+	if l.Level >= LevelError {
+		fmt.Fprintf(l.stderr(), format+"\n", v...)
+	}
+}
+
+// Infof logs an informational message using Printf conventions.
+func (l *LeveledLogger) Infof(format string, v ...interface{}) {
+	if l.Level >= LevelInfo {
+		fmt.Fprintf(l.stdout(), format+"\n", v...)
+	}
+}
+
+// Warnf logs a warning message using Printf conventions.
+func (l *LeveledLogger) Warnf(format string, v ...interface{}) {
+	if l.Level >= LevelWarn {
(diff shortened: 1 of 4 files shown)


log.go
```diff
@@ -0,0 +1,212 @@
+package stripe
```
brandur-stripe: I ended up moving all log-related constructs to a new file in the top-level package so that it's easier to find log-related things.

log.go
```diff
@@ -0,0 +1,212 @@
...
+//
+
+const (
+	LevelError Level = 1
+	LevelWarn Level = 2
+	LevelInfo Level = 3
+	LevelDebug Level = 4
+
+	// Older deprecated levels for Printfer-style logging.
+	printferLevelError = 1
```
brandur-stripe: These are new constants for a deprecated idea, but they're not exported, and make code much more readable.

stripe.go
```diff
@@ -167,9 +152,8 @@ type BackendImplementation struct {
 	Type              SupportedBackend
 	URL               string
 	HTTPClient        *http.Client
+	LeveledLogger     LeveledLoggerInterface
 	MaxNetworkRetries int
-	LogLevel          int
-	Logger            Printfer
```
brandur-stripe: This is technically a backwards-incompatible change, but I think I'm going to release as a minor anyway. There's no plausible reason to ever access a `BackendImplementation` directly (and especially not the `LogLevel` and `Logger` fields), and I've had a note in its docblock that we're going to export it soon forever now.

stripe.go
```diff
@@ -64,21 +62,6 @@ var EnableTelemetry = false
...
-var LogLevel = 2
-
-// Logger controls how stripe performs logging at a package level. It is useful
-// to customise if you need it prefixed for your application to meet other
-// requirements.
-//
-// This Logger will be inherited by any backends created by default, but will
-// be overridden if a backend is created with GetBackendWithConfig.
-var Logger Printfer
-
```
brandur-stripe: Almost all the red in this file is due to things moving to `log.go`.

brandur-stripe:
Thanks @mickeyreiss!

r? @remi-stripe Would you mind taking a pass on this as well? Thanks!

stripe.go
remi-stripe: Works for me, though the cost of doing a major is quite low and so when in doubt, maybe a major is better? We could also wait for the next breaking change to release this.

But ultimately fine with your approach.

stripe.go
```diff
@@ -394,38 +371,29 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
 		}
 
 		sleepDuration := s.sleepTime(retry)
 		retry++
 
-		if s.LogLevel > 1 {
-			s.Logger.Printf("Initiating retry %v for request %v %v%v after sleeping %v\n",
-				retry, req.Method, req.URL.Host, req.URL.Path, sleepDuration)
-		}
+		s.LeveledLogger.Infof("Initiating retry %v for request %v %v%v after sleeping %v",
```
remi-stripe: should this be Warnf?

stripe.go
```diff
@@ -450,19 +418,15 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
...
 	}
 
 	if res.StatusCode >= 400 {
 		return s.ResponseToError(res, resBody)
 	}
 
-	if s.LogLevel > 2 {
-		s.Logger.Printf("Response: %s\n", string(resBody))
-	}
+	s.LeveledLogger.Errorf("Response: %s\n", string(resBody))
```
remi-stripe: Should this be `Infof`?

stripe.go
brandur-stripe: Hmm, thinking about it more deeply, I could see it either way. It is probably not actually a concerning operation, so `Infof` might be fine, but it's also a little unusual, so I could see `Warnf` too.

I'll change to `Warnf` just to make it more distinctive from other `Infof` messages.

stripe.go
remi-stripe: I was mostly going from the fact that it was `s.LogLevel > 2` before

brandur-stripe: Err, yes .... :)

I'd say even `Debugf` is more appropriate even given that this is raw data. Going to change to that.

stripe.go
remi-stripe: I was mostly going from the fact that it used to be `if s.LogLevel > 1 {`

brandur-stripe: Oh I see. Yeah, I think I messed up the port.

brandur-stripe:
Thanks Remi!

> This looks great though I don't grasp every single change in this PR. Like I missed why we still seem to need the printfer logic for example even though we say it's deprecated.

Ah right. So in this case it's mostly to keep backwards compatibility around for a little bit longer just so we don't force people to migrate right away. `Printfer` is not compatible with the new `LeveledLoggerInterface`, so we can't reuse the old `stripe.Logger = ...` variable that probably a ton of people were using.

Instead, we convert `Printfer`s to `LeveledLoggerInterface` under the hood for now with a shim, recommend the use of `stripe.DefaultLeveledLogger = ...` in our docs, and then think about dropping `stripe.Logger` and `Printfer` a few major versions from now (and I think we could put a note in the CHANGELOG about how to port in a way that's relatively easy).

> Would love to see an example in the readme on how to do a more custom logger? Like what if I wanted to prefix messages with [E], [W] or [I]? But it's more for my own education I think because I don't fully understand the LeveledLogger versus LeveledLoggerInterface as you seem to use them interchangeably.

So `LeveledLoggerInterface` is an interface designed to support a variety of different loggers that we can plug in.

`LeveledLogger` is a really basic implementation of it that we'll bundle into the stripe-go package. It's usable, but the idea is that it won't do anything too sophisticated.

If you wanted to say add a prefix (and actually, we might want to commit a further change to do something like that by default in `LeveledLogger`), you'd plug in something like Logrus which supports [formatters](https://github.com/sirupsen/logrus/#formatters).

This example probably won't compile, but it might look something like this:

``` go
type MyFormatter struct {
}

func (f *MyFormatter) Format(entry *Entry) ([]byte, error) {
    var prefix
    switch entry.Level {
        case logrus.DebugLevel:
            prefix = "[r] "
        case logrus.ErrorLevel:
            prefix = "[e] "
        case logrus.InfoLevel:
            prefix = "[i] "
        case logrus.WarnLevel:
            prefix = "[w] "
    }
    return []byte(prefix + entry.Message), nil
}

func init() {
    stripe.DefaultLeveledLogger = &logrus.Logger{
        Formatter: &MyFormatter{},
    }
}
```

(Or something like that.)

We could provide similar functionality in `LeveledLogger`, but I'd probably leave it out for the time being and recommend a third party package that does it better. Logrus does colors, structured logging, and a bunch of other nice stuff that you'd want in a real app.

> Works for me, though the cost of doing a major is quite low and so when in doubt, maybe a major is better? We could also wait for the next breaking change to release this.

Yeah, true. If it works for you, I think I'll do a a major and mark it as a minor breaking change in the CHANGELOG, but I'll do it now.

mickeyreiss:
My perspective (an anecdote as a library user): I appreciate the thought
and effort put into compatibility. I would much prefer to have some
compatibility cruft than ever be forced to upgrade working integration
code, in order to adopt a new feature. For me, the trade-off is tied to the
sense that changing my sensitive payments code in production can increase
risk of bugs and regressions.

One idea, in addition to the CHANGELOG note: If you prefix the comments
with `deprecated:`, my IDE (Goland) will clearly mark the symbol with a
strikeout and include the comment at the callsite.

brandur-stripe:
Thanks for the note @mickeyreiss.

> One idea, in addition to the CHANGELOG note: If you prefix the comments
> with `deprecated:`, my IDE (Goland) will clearly mark the symbol with a
> strikeout and include the comment at the callsite.

Good idea. It turns out that this is actually [an official convention](https://rakyll.org/deprecated/) established by the standard library that I wasn't aware of. I've just pushed a change to reformat all the deprecated notes I added to use that instead.

brandur-stripe:
Released as 59.0.0.

PR #819, opened 2019-03-22
Add log prefixes + test suite

brandur:
Adds level prefixes to the default `LeveledLogger` which will look
something like:

    [DEBUG] A debug message
    [INFO] An informational message
    [WARN] A warning message
    [ERROR] An error message

I also included a log test suite which I actually wrote for the last
leveled logging patch, but apparently completely forgot to commit.

r? @remi-stripe

### DIFF
--- a/log.go
+++ b/log.go
@@ -100,7 +100,7 @@ type LeveledLogger struct {
 // Debugf logs a debug message using Printf conventions.
 func (l *LeveledLogger) Debugf(format string, v ...interface{}) {
 	if l.Level >= LevelDebug {
-		fmt.Fprintf(l.stdout(), format+"\n", v...)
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks Remi!

> I'm partial to shorter prefixes `[E]`, `[W]` and such but this looks great and I'm really glad we now have tests for this!

Haha, I'm not crazy about the one letter prefixes, but I could probably convinced to have shorter ones so that they're all the same length and line up nicely (`[INFO]`, `[WARN]`, `[ERRO]`, `[DEBU]`).

Anyway, it's definitely not breaking if we want to change it, so going to merge this as is for now and we can revisit later if someone feels strongly.

brandur-stripe:
Released as 59.1.0.

PR #820, opened 2019-03-27
Add a long list of missing params and properties

remi-stripe:
r? @brandur-stripe 
cc @mickjermsurawong-stripe who found all of those for us.
cc @stripe/api-libraries

### DIFF
--- a/refund.go
+++ b/refund.go
@@ -37,7 +37,10 @@ type RefundParams struct {
 // RefundListParams is the set of parameters that can be used when listing refunds.
 // For more details see https://stripe.com/docs/api#list_refunds.
 type RefundListParams struct {
-	ListParams `form:"*"`
+	ListParams   `form:"*"`
+	Charge       *string           `form:"charge"`
+	Created      *int64            `form:"created"`
+	CreatedRange *RangeQueryParams `form:"created"`
 }
 
 // Refund is the resource representing a Stripe refund.
--- a/token.go
+++ b/token.go
@@ -46,6 +46,6 @@ type Token struct {
 
 // PIIParams are parameters for personal identifiable information (PII).
 type PIIParams struct {
-	Params           `form:"*"`
-	PersonalIDNumber *string `form:"personal_id_number"`
+	Params   `form:"*"`
+	IDNumber *string `form:"id_number"`
 }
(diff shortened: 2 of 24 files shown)


token.go
```diff
@@ -46,6 +46,6 @@ type Token struct {
 
 // PIIParams are parameters for personal identifiable information (PII).
 type PIIParams struct {
-	Params           `form:"*"`
-	PersonalIDNumber *string `form:"personal_id_number"`
+	Params   `form:"*"`
+	IDNumber *string `form:"id_number"`
```
remi-stripe: This is a breaking change and likely requires a major version to play it safe even if it likely did not work.

refund.go
```diff
@@ -38,6 +38,7 @@ type RefundParams struct {
 // For more details see https://stripe.com/docs/api#list_refunds.
 type RefundListParams struct {
 	ListParams `form:"*"`
+	Charge     *string `form:"charge"`
```
mickjermsurawong-stripe: One more created range here pls Remi
https://stripe.com/docs/api/refunds/list#list_refunds-created

remi-stripe:
@mickjermsurawong-stripe IMO we should not keep such a PR open for a week. There's no real upside to "waiting until we find more", we would just do a brand new PR at that point. 

I also think the empty params should be resolved entirely separately and not in this PR. We should also talk about the best approach internally instead.

mickjermsurawong-stripe:
ah that's fair. I'm good with merging this then
LGTM

brandur-stripe:
> I also think the empty params should be resolved entirely separately and not in this PR. We should also talk about the best approach internally instead.

+1. I think we're going to need to add a new property (e.g. `Empty`) and custom encode function to support this properly unfortunately.

mickjermsurawong-stripe:
The build was stuck but seems working now. I can take the release

remi-stripe:
@brandur-stripe yeah I canceled and restarted. Leaving the release to you, thanks a lot! Reminder that it's a major version to be safe

brandur-stripe:
Released as 60.0.0.

mickjermsurawong-stripe:
ah thank you @brandur! i'm just about reading instructions and preparing the release message!

brandur-stripe:
> ah thank you @brandur-stripe! i'm just about reading instructions and preparing the release message!

Oops, sorry @mickjermsurawong-stripe — I didn't mean to step on your toes. Somehow I read the "you" in as meaning me, hah (which reading through again makes no sense I see).

PR #824, opened 2019-04-02
Add support for payment_intent on invoice and default_payment_method

remi-stripe:
This is now ready to be released. Those fields will be public later this week and are stable. Having this out ahead of time is likely the best approach.

r? @mickjermsurawong-stripe 
cc @stripe/api-libraries

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -53,8 +53,9 @@ type CustomerInvoiceCustomFieldParams struct {
 // CustomerInvoiceSettingsParams is the structure containing the default settings for invoices
 // associated with this customer.
 type CustomerInvoiceSettingsParams struct {
-	CustomFields []*CustomerInvoiceCustomFieldParams `form:"custom_fields"`
-	Footer       *string                             `form:"footer"`
+	CustomFields         []*CustomerInvoiceCustomFieldParams `form:"custom_fields"`
+	DefaultPaymentMethod *string                             `form:"default_payment_method"`
+	Footer               *string                             `form:"footer"`
 }
 
 // CustomerShippingDetailsParams is the structure containing shipping information.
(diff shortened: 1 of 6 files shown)


customer.go
```diff
@@ -53,8 +53,9 @@ type CustomerInvoiceCustomFieldParams struct {
 // CustomerInvoiceSettingsParams is the structure containing the default settings for invoices
 // associated with this customer.
 type CustomerInvoiceSettingsParams struct {
-	CustomFields []*CustomerInvoiceCustomFieldParams `form:"custom_fields"`
-	Footer       *string                             `form:"footer"`
+	CustomFields         []*CustomerInvoiceCustomFieldParams `form:"custom_fields"`
+	DefaultPaymentMethod *string                             `form:"default_payment_method"`
+	Footer               *string                             `form:"footer"`
```
mickjermsurawong-stripe: Ah just flagging that `default_payment_method` is only present in update and not create.

invoice.go
```diff
@@ -87,6 +87,7 @@ type InvoiceParams struct {
 	CustomFields         []*InvoiceCustomFieldParams `form:"custom_fields"`
 	Customer             *string                     `form:"customer"`
 	DaysUntilDue         *int64                      `form:"days_until_due"`
+	DefaultPaymentMethod *string                     `form:"default_payment_method"`
```
mickjermsurawong-stripe: @remi-stripe 
We miss `payment_method` in  `InvoicePayParams`

customer.go
```diff
@@ -110,6 +111,20 @@ type Customer struct {
...
+}
+
+// CustomerInvoiceSettings is the structure containing the default settings for invoices associated
+// with this customer.
+type CustomerInvoiceSettings struct {
+	CustomFields         []*CustomerInvoiceCustomField `json:"custom_fields"`
+	DefaultPaymentMethod *PaymentMethod                `json:"default_payment_method"`
+	Footer               string                        `json:"footer"`
+}
+
```
mickjermsurawong-stripe: We miss expandable `default_payment_method` on Customer struct here.
Or do you leave it out intentionally?

mickjermsurawong-stripe:
ptal @remi-stripe

customer.go
mickjermsurawong-stripe: Oh whoops my bad. It is correct here!

customer.go
remi-stripe: Fairly sure this is a bug in the backend so will flag.

remi-stripe:
@mickjermsurawong-stripe Fixed the Pay one, PTAL

mickjermsurawong-stripe:
LGTM
thanks @remi-stripe

PR #827, opened 2019-04-10
Add support for `confirmation_method` on PaymentIntent

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries @garrett-stripe

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -28,6 +28,9 @@ type PaymentIntentConfirmationMethod string
 
 // List of values that PaymentIntentConfirmationMethod can take.
 const (
+	PaymentIntentConfirmationMethodAutomatic PaymentIntentConfirmationMethod = "automatic"
+	PaymentIntentConfirmationMethodManual    PaymentIntentConfirmationMethod = "manual"
+	// The following constants are considered deprecated
 	PaymentIntentConfirmationMethodPublishable PaymentIntentConfirmationMethod = "publishable"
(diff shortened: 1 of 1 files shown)


paymentintent.go
```diff
@@ -90,8 +93,9 @@ type PaymentIntentParams struct {
 	Params               `form:"*"`
 	Amount               *int64                           `form:"amount"`
 	ApplicationFeeAmount *int64                           `form:"application_fee_amount"`
-	Confirm              *bool                            `form:"confirm"`
 	CaptureMethod        *string                          `form:"capture_method"`
+	Confirm              *bool                            `form:"confirm"`
+	ConfirmationMethod   *string                          `form:"confirmation_method"`
```
mickjermsurawong-stripe: @remi, I think `confirmation_method` is on `PaymentIntentCreateParams` instead?
It's currently gated in `enable_payment_intent_manual_confirmation`

mickjermsurawong-stripe:
ptal @remi-stripe

paymentintent.go
remi-stripe: @mickjermsurawong-stripe That's where I put it though https://github.com/stripe/stripe-go/blob/5335e5f4161ab7422d39b7fe1f9983121584bbe8/paymentintent.go#L98

We don't have CreateParams and UpdateParams in go. Just Params. So it works, it just only works on Create like CaptureMethod

mickjermsurawong-stripe: ah ack! sorry I forgot the convention we have here.

mickjermsurawong-stripe:
LGTM

PR #837, opened 2019-04-17
Add helper to go from []T to []*T for string, int64, float64, bool

sobel-stripe:
I ran into a [common mistake in golang](https://developmentality.wordpress.com/2014/02/25/go-gotcha-0-why-taking-the-address-of-an-iterated-variable-is-wrong/) when writing my own code to go from a `[]string` to `[]*string` for use in the `PaymentMethodTypes` field.

This adds a helper function to do it correctly. If we want to merge this PR, I also think that we should update the examples in the documentation to use it.

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -52,9 +52,9 @@ func TestAccountNew(t *testing.T) {
 		ExternalAccount: &stripe.AccountExternalAccountParams{
 			Token: stripe.String("tok_123"),
 		},
-		RequestedCapabilities: []*string{
-			stripe.String("card_payments"),
-		},
+		RequestedCapabilities: stripe.StringSlice([]string{
+			"card_payments",
+		}),
 		Settings: &stripe.AccountSettingsParams{
 			Branding: &stripe.AccountSettingsBrandingParams{
 				Icon: stripe.String("file_123"),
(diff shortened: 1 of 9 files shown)


brandur-stripe:
Yeah, this makes sense to me — the AWS SDK has it already, and it'll make a lot of our test code cleaner as it becomes possible to leave out all the individual elements' own `stripe.String`.

@sobel-stripe Thanks! Would you mind updating this so that there's one for all the types that we support? (e.g. `Int64Slice`) It shouldn't just be a string-specific construct.

sobel-stripe:
@brandur-stripe done, and updated the test call sites [could only find string] as well

brandur-stripe:
Thanks @sobel-stripe (and wow, thanks a lot for updating all those other tests).

Sorry to be a pain, but one more small-ish ask: would you mind putting tests in `stripe_test.go` for all the different new functions for non-string types as well? I know the type system pretty much guarantees that we're good here, but just seems good for general hygiene.

ptal @sobel-stripe

louissobel:
@brandur-stripe no problem — added tests

sobel-stripe:
@brandur-stripe  no problem — added the tests

brandur-stripe:
Thanks @sobel-stripe! LGTM.

brandur-stripe:
Released as 60.6.0.

PR #839, opened 2019-04-30
Add ACSS Debit to PaymentMethodDetails

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries


mickjermsurawong-stripe:
Just to double-check that Go:
- uses typed data source type
- doesn't provide type params for source create params.
Otherwise,
LGTM

PR #843, opened 2019-05-03
Lock before changing field of backends

chrsmith:
Just a minor thing I noticed while trying to update a Stripe client to specify `MaxNetworkRetries`, since the default is 0. Apologies if this is somehow intentional.

---

The `SetBackend` method updates one of the members of the package-level variable `backends`. Since it's a pointer, it's probably safe on most architectures. But the `Backends` struct already has a `sync.RWMutex` field, and earlier in the file the `GetBackend` function goes through the trouble of explicitly acquiring a read and possibly a write lock. See:
https://github.com/stripe/stripe-go/blob/d94ce0ed857fb1571b9b30bd24430bdac3d50c0f/stripe.go#L666

So to be safe, and consistent with the rest of the code, it makes sense to acquire a write lock before updating the `backends` value.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -816,7 +816,7 @@ func SetAppInfo(info *AppInfo) {
 func SetBackend(backend SupportedBackend, b Backend) {
 	backends.mu.Lock()
 	defer backends.mu.Unlock()
-	
+
 	switch backend {
 	case APIBackend:
 		backends.API = b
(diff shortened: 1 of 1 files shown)


brandur-stripe:
@chrsmith Yep, sounds good to me!

The build is failing because Gofmt needs to be run on your changes. It's not obvious to me from your diff exactly what the problem is though (probably leading or trailing whitespace?).

chrsmith:
@brandur-stripe Sorry about that. I made the change in GitHub's code editor, so it's probably some whitespace issue. Fixed.

brandur-stripe:
Awesome, thanks!

brandur-stripe:
Released as 60.12.2.

PR #845, opened 2019-05-04
Add support for Customer Balance Transaction resource and APIs

remi-stripe:
r? @mickjermsurawong-stripe 
cc @stripe/api-libraries

### DIFF
--- a/customerbalancetransaction.go
+++ b/customerbalancetransaction.go
@@ -1,0 +1,81 @@
+package stripe
+
+import "encoding/json"
+
+// CustomerBalanceTransactionType is the list of allowed values for the customer's balance
+// transaction type.
+type CustomerBalanceTransactionType string
+
+// List of values that CustomerBalanceTransactionDuration can take.
+const (
+	CustomerBalanceTransactionTypeAdjustment            CustomerBalanceTransactionType = "adjustment"
+	CustomerBalanceTransactionTypeAppliedToInvoice      CustomerBalanceTransactionType = "applied_to_invoice"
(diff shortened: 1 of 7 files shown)


customerbalancetransaction.go
```diff
@@ -0,0 +1,86 @@
...
+type CustomerBalanceTransaction struct {
+	Amount        int64                          `json:"amount"`
+	Created       int64                          `json:"created"`
+	Currency      Currency                       `json:"currency"`
+	Customer      *Customer                      `json:"customer"`
+	Description   string                         `json:"description"`
+	EndingBalance int64                          `json:"ending_balance"`
+	ID            string                         `json:"id"`
+	Invoice       *Invoice                       `json:"invoice"`
+	Livemode      bool                           `json:"livemode"`
```
mickjermsurawong-stripe: I think `livemode` field shouldn't be present here. Both api refs, and OpenAPI doesn't have it

customerbalancetransaction.go
```diff
@@ -0,0 +1,86 @@
+package stripe
+
+import "encoding/json"
+
+// CustomerBalanceTransactionType is the list of allowed values for the customer's balance
+// transaction type..
+type CustomerBalanceTransactionType string
```
mickjermsurawong-stripe: Extra "." in the comment

customerbalancetransaction.go
remi-stripe: Yep I put it because I think they made a mistake, all objects should have this field, and I've asked them to fix it already. Will report back!

remi-stripe:
PTAL @ob-stripe

ob-stripe:
Released as 61.4.0.

Issue #849, opened 2019-05-07
Source doesn't have Customer data?


danqing:
According to the [documentation](https://stripe.com/docs/api/sources/attach), source objects have a string key called `customer` that is supposed to hold the customer ID. The [Source type](https://godoc.org/github.com/stripe/stripe-go#Source) doesn't have such a key though. Am I missing anything? Thanks!

remi-stripe:
@danqing This is now fixed in 60.13.0

PR #851, opened 2019-05-08
Add Customer property on the Source object

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/849

r? @mickjermsurawong-stripe 
cc @stripe/api-libraries

### DIFF
--- a/source.go
+++ b/source.go
@@ -230,6 +230,7 @@ type Source struct {
 	CodeVerification    *CodeVerificationFlow `json:"code_verification,omitempty"`
 	Created             int64                 `json:"created"`
 	Currency            Currency              `json:"currency"`
+	Customer            string                `json:"customer"`
 	Flow                SourceFlow            `json:"flow"`
 	ID                  string                `json:"id"`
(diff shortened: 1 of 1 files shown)


mickjermsurawong-stripe:
@remi-stripe 
I think the field is not expandable
https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/Sources/Source.cs#L51

source.go
```diff
@@ -230,6 +230,7 @@ type Source struct {
 	CodeVerification    *CodeVerificationFlow `json:"code_verification,omitempty"`
 	Created             int64                 `json:"created"`
 	Currency            Currency              `json:"currency"`
+	Customer            string                `json:"customer"`
```
remi-stripe: It is not technically expandable, and we're unlikely to change this so I choose a string property this time. If it became expandable, it would be a breaking change but I think it's fine.

mickjermsurawong-stripe:
ah you fixed it! 
LGTM

PR #854, opened 2019-05-08
Add test script to check that all clients are included in `client.API`

brandur:
When adding a new package with a `Client` struct, it's really easy to
forget to add it to the big list of clients in `client.API`, and we
often have to rely on user reports to tell us that we forgot about it.

Here we add a script to run with CI which looks for packages with a
`Client` and compares that list against what's in `client.API`, then
errors if any are missing from the latter.

Also adds a few clients that were missing from the list.

r? @remi-stripe
cc @stripe/api-libraries

---

A note on design: I wrote the script in Go because (1) it doesn't make a
bad scripting language, (2) we get a lot of typing safety that some Ruby
thing wouldn't give us, and (3) it allowed me to leverage Go's built in
AST parser for its own syntax so we don't have to do as much fuzzy
matching against strings.

By the end of writing it I think it was the right decision because there
are a few cases involved here that make the check non-trivial. For
example, packages can be aliased when imported in `client/api.go`, and
we need to be able to translate those aliases back to real package paths
that we detected earlier.

The final script is a little on the long side (~300 LOC), but it handles
all the corners we've developed, and I think it'll be relatively stable.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -4,6 +4,7 @@ package client
 import (
 	stripe "github.com/stripe/stripe-go"
 	"github.com/stripe/stripe-go/account"
+	"github.com/stripe/stripe-go/accountlink"
 	"github.com/stripe/stripe-go/applepaydomain"
 	"github.com/stripe/stripe-go/balance"
 	"github.com/stripe/stripe-go/bankaccount"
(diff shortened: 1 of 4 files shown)


client/api.go
```diff
@@ -190,6 +200,8 @@ type API struct {
 	TerminalLocations *terminallocation.Client
 	// TerminalReaders is the client used to invoke /terminal/readers related APIs.
 	TerminalReaders *terminalreader.Client
+	// ThreeDSecures is the client used to invoke /3d_secure related APIs.
+	ThreeDSecures *threedsecure.Client
```
brandur-stripe: @remi-stripe Haha, I imagine that you're not super excited about exposing some of these. I kind of figured we should probably just do it for consistency as long as the subpackages are still present in the project. The only alternative option would be to add a blacklist in the check script, and it'd be nice if we could avoid that. Thoughts?

remi-stripe: Yep I started typing and then realized it's cleaner to have it and remove later.

brandur-stripe:
Thanks Remi!

PR #857, opened 2019-05-10
Fix invoice's `PaymentIntent` to its JSON tag uses snakecase

### LINKED ISSUE #856: Bug in invoice.go: `json:"paymentIntent"` (camel case instead of snake case)
Looks like a bug on this line:

https://github.com/stripe/stripe-go/blob/24c6ea237b6270ac71be612718a495e3a2dff740/invoice.go#L229

Should use snake case instead of camel case, e.g.:
```
PaymentIntent                *PaymentIntent           `json:"payment_intent"`
```

As in API: https://stripe.com/docs/api/invoices/object#invoice_object-payment_intent

brandur:
Just through bad copy/paste this ended up using camelcase instead of
snakecase.

Fixes #856.

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -226,7 +226,7 @@ type Invoice struct {
 	NextPaymentAttempt           int64                    `json:"next_payment_attempt"`
 	Number                       string                   `json:"number"`
 	Paid                         bool                     `json:"paid"`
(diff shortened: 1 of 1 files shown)


brandur-stripe:
> Thanks for fixing @brandur-stripe ! I also confirm there are no other property or parameter with that bug right now

+1. I looked as well and couldn't find any.

PR #860, opened 2019-05-16
Add OffSession parameter to PaymentIntent

danwang-stripe:
cc @stripe/api-libraries 
r? @brandur @remi-stripe 

Adds `off_session` parameter [0] to PaymentIntent create and confirm.

[0] https://stripe.com/docs/api/payment_intents/confirm#confirm_payment_intent-off_session

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -43,6 +43,15 @@ const (
 	PaymentIntentNextActionTypeRedirectToURL PaymentIntentNextActionType = "redirect_to_url"
 )
 
+// PaymentIntentOffSession is the list of allowed values for types of off-session.
+type PaymentIntentOffSession string
+
+// List of values that PaymentIntentOffSession can take.
+const (
+	PaymentIntentOffSessionOneOff    PaymentIntentOffSession = "one_off"
+	PaymentIntentOffSessionRecurring PaymentIntentOffSession = "recurring"
+)
+
 // PaymentIntentStatus is the list of allowed values for the payment intent's status.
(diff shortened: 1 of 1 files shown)


brandur-stripe:
@danwang-stripe Thanks for the pull!

For enum-like fields, we also normally include some constants that people can use in case they'd prefer to do that. Can you add something like this to the file (try to insert where it's alphabetically ordered):

``` patch
diff --git a/paymentintent.go b/paymentintent.go
index bb9948b..834fc20 100644
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -43,6 +43,15 @@ const (
 	PaymentIntentNextActionTypeRedirectToURL PaymentIntentNextActionType = "redirect_to_url"
 )
 
+// PaymentIntentOffSession is the list of allowed values for types of off-session.
+type PaymentIntentOffSession string
+
+// List of values that PaymentIntentOffSession can take.
+const (
+	PaymentIntentOffSessionOneOff    PaymentIntentOffSession = "one_off"
+	PaymentIntentOffSessionRecurring PaymentIntentOffSession = "recurring"
+)
+
 // PaymentIntentStatus is the list of allowed values for the payment intent's status.
 type PaymentIntentStatus string
```

Thanks!

ptal @danwang-stripe

danwang-stripe:
👍thanks, added the enums in 1bb7754!

ptal @brandur-stripe

brandur-stripe:
Thanks! LGTM.

brandur-stripe:
Released as 60.17.0.

PR #865, opened 2019-05-24
Fix earlyfraudwarning client

ob-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries 

My bad, you added a check specifically for this but I missed it because of the API version mismatch issue :/

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -45,6 +45,7 @@ import (
 	"github.com/stripe/stripe-go/person"
 	"github.com/stripe/stripe-go/plan"
 	"github.com/stripe/stripe-go/product"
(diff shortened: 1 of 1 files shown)


ob-stripe:
Fixed, ptal @brandur-stripe 

> Damn, did the build not catch this?! If not, something in my new script isn't working :/

I think the build in #862 didn't catch it because it failed earlier (because of the API version mismatch in stripe-mock).

The build in #864 didn't catch it either, I must have forgotten to rebase after merging #862 :/

ob-stripe:
Thanks Brandur!

Released as 61.0.1.

PR #874, opened 2019-06-18
Log only to info on 402 errors from Stripe

### LINKED ISSUE #869: Feature request: Downgrade log level for `card_error`s
Following up on #813, we've noticed error level logging for error responses of type `card_error`.

My understanding is that these errors require user input to resolve them, so they are normal processing and not possible to resolve with backend code.

Is this log level intentional, or would it make sense to downgrade these types of error logs?

brandur:
As discussed in #869, breaks apart how Stripe errors are logged so that
status 402 errors go to info instead of error. This is meant to signal
that those errors are probably prompted by user data (as opposed to
sending invalid parameters or something), and handling them differently
allows integrations to avoid filling their error logs with messages that
they can't really prevent.

Fixes #869.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -486,7 +486,25 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
 	}
 	raw.E.Err = typedError
 
-	s.LeveledLogger.Errorf("Error encountered from Stripe: %v", raw.E.Error)
+	// The Stripe API makes a distinction between errors that were caused by
+	// invalid parameters or something else versus those that occurred
+	// *despite* valid parameters, the latter coming back with status 402.
+	//
+	// On a 402, log to info so as to not make an integration's log noisy with
+	// error messages that they don't have much control over.
+	//
+	// Note I use the constant 402 here instead of an `http.Status*` constant
+	// because technically 402 is "Payment required". The Stripe API doesn't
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks OB!

stripe.go
```diff
@@ -486,7 +486,25 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
...
+	//
+	// On a 402, log to info so as to not make an integration's log noisy with
+	// error messages that they don't have much control over.
+	//
+	// Note I use the constant 402 here instead of an `http.Status*` constant
+	// because technically 402 is "Payment required". The Stripe API doesn't
+	// comply to the letter of the specification and uses it in a broader
+	// sense.
+	if res.StatusCode == 402 {
+		s.LeveledLogger.Infof("User-compelled request error from Stripe: %v",
```
mickeyreiss: I find the wording here a bit confusing. Maybe a standard format could help, along the lines of:

    “Stripe request GET /blah returned 402 User Error: Invalid Card Expiration”

With method, path, status, status text and error message being placeholders?

mickeyreiss:
LGTM otherwise, solving the problem from the issue!

PR #879, opened 2019-06-21
Add OAuth

cjavilla-stripe:
r? @remi-stripe 

Looking for some feedback:

1) Is this PR too big?
2) Is this a decent param passing approach? (I'm about 95% confident that I understand the reason for using pointers)
3) Is stubbing out the server like this ok to test the oauth flow since oauth isnt supported by stripe-mock?

Here's a sample echo server that uses this successfully:

```go
package main
import (
  // "bytes"
  // "encoding/json"
  "github.com/foolin/goview/supports/echoview"
  "github.com/labstack/echo"
  "github.com/labstack/echo/middleware"
  "log"
  "net/http"
  "github.com/stripe/stripe-go"
  "github.com/stripe/stripe-go/oauth"
)

func main() {
  stripe.ClientID = "ca_123"
  stripe.Key = "sk_123"

  e := echo.New()
  e.Use(middleware.Logger())
  e.Use(middleware.Recover())
  e.Renderer = echoview.Default()

  e.GET("/", func(c echo.Context) error {
    return c.Render(http.StatusOK, "index", echo.Map{})
  })

  e.GET("/start", func(c echo.Context) error {
    // THEN...
    // client_id := "ca_123"
    // base := "https://connect.stripe.com"
    // path := "/oauth/authorize"
    // response_type := "code"
    // redirect_uri := "http://localhost:1323/callback"
    // query := fmt.Sprintf(
    //   "client_id=%s&response_type=%s&redirect_uri=%s",
    //   client_id,
    //   response_type,
    //   redirect_uri,
    // )
    // url := fmt.Sprintf("%s%s?%s", base, path, query)
    // NOW...
    url := oauth.AuthorizeURL(&stripe.AuthorizeURLParams{
      RedirectURI: stripe.String("http://localhost:1323/callback"),
    })
    return c.Redirect(http.StatusTemporaryRedirect, url)
  })

  e.GET("/callback", func(c echo.Context) error {
    // THEN...
    // client_secret := "sk_test_123"
    // code := c.QueryParam("code")
    // scope := c.QueryParam("scope")
    // state := c.QueryParam("state")
    // message := map[string]interface{}{
    //   "client_secret": client_secret,
    //   "grant_type": "authorization_code",
    //   "code": code,
    // }
    // bytesRepresentation, err := json.Marshal(message)
    // if err != nil {
    //   log.Fatalln(err)
    // }
    // resp, err := http.Post(
    //   "https://connect.stripe.com/oauth/token",
    //   "application/json",
    //   bytes.NewBuffer(bytesRepresentation),
    // )
    // if err != nil {
    //   log.Fatalln(err)
    // }
    // var result map[string]interface{}
    // json.NewDecoder(resp.Body).Decode(&result)
    // log.Println(result)

    token, err := oauth.New(&stripe.OAuthTokenParams{
      Code:      stripe.String(c.QueryParam("code")),
    })
    if err != nil {
      log.Fatalln(err)
    }
    return c.HTML(http.StatusOK, token.StripeUserID)
  })
  e.Logger.Fatal(e.Start(":1323"))
}
```

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -37,6 +37,7 @@ import (
 	issuingdispute "github.com/stripe/stripe-go/issuing/dispute"
 	"github.com/stripe/stripe-go/issuing/transaction"
 	"github.com/stripe/stripe-go/loginlink"
+	"github.com/stripe/stripe-go/oauth"
 	"github.com/stripe/stripe-go/order"
 	"github.com/stripe/stripe-go/orderreturn"
 	"github.com/stripe/stripe-go/paymentintent"
@@ -148,6 +149,8 @@ type API struct {
 	IssuingTransactions *transaction.Client
 	// LoginLinks is the client used to invoke login link related APIs.
 	LoginLinks *loginlink.Client
+	// OAuth is the client used to invoke /oauth APIs.
+	OAuth *oauth.Client
 	// Orders is the client used to invoke /orders APIs.
 	Orders *order.Client
 	// OrderReturns is the client used to invoke /order_returns APIs.
@@ -234,6 +237,7 @@ func (a *API) Init(key string, backends *stripe.Backends) {
 	if backends == nil {
 		backends = &stripe.Backends{
 			API:     stripe.GetBackend(stripe.APIBackend),
+			Connect: stripe.GetBackend(stripe.ConnectBackend),
 			Uploads: stripe.GetBackend(stripe.UploadsBackend),
 		}
 	}
@@ -272,8 +276,9 @@ func (a *API) Init(key string, backends *stripe.Backends) {
 	a.IssuingDisputes = &issuingdispute.Client{B: backends.API, Key: key}
 	a.IssuingTransactions = &transaction.Client{B: backends.API, Key: key}
 	a.LoginLinks = &loginlink.Client{B: backends.API, Key: key}
-	a.Orders = &order.Client{B: backends.API, Key: key}
+	a.OAuth = &oauth.Client{B: backends.Connect, Key: key}
 	a.OrderReturns = &orderreturn.Client{B: backends.API, Key: key}
+	a.Orders = &order.Client{B: backends.API, Key: key}
 	a.PaymentIntents = &paymentintent.Client{B: backends.API, Key: key}
 	a.PaymentMethods = &paymentmethod.Client{B: backends.API, Key: key}
 	a.PaymentSource = &paymentsource.Client{B: backends.API, Key: key}
--- a/error.go
+++ b/error.go
@@ -136,6 +136,10 @@ type Error struct {
 	SetupIntent    *SetupIntent   `json:"setup_intent,omitempty"`
 	Source         *PaymentSource `json:"source,omitempty"`
 	Type           ErrorType      `json:"type"`
+
+	// OAuth specific Error properties. Named OAuthError because of name conflict.
+	OAuthError            string `json:"error,omitempty"`
+	OAuthErrorDescription string `json:"error_description,omitempty"`
 }
 
 // Error serializes the error object to JSON and returns it as a string.
--- a/form/form.go
+++ b/form/form.go
@@ -538,14 +538,18 @@ func (f *Values) Add(key, val string) {
 	f.values = append(f.values, formValue{key, val})
 }
 
-// Encode encodes the values into “URL encoded” form ("bar=baz&foo=quux").
+// Encode encodes the keys and values into “URL encoded” form
+// ("bar=baz&foo=quux").
 func (f *Values) Encode() string {
 	var buf bytes.Buffer
 	for _, v := range f.values {
 		if buf.Len() > 0 {
 			buf.WriteByte('&')
 		}
-		buf.WriteString(url.QueryEscape(v.Key))
+		key := url.QueryEscape(v.Key)
+		key = strings.Replace(key, "%5B", "[", -1)
+		key = strings.Replace(key, "%5D", "]", -1)
+		buf.WriteString(key)
 		buf.WriteString("=")
 		buf.WriteString(url.QueryEscape(v.Value))
 	}
--- a/oauth.go
+++ b/oauth.go
@@ -1,0 +1,129 @@
+package stripe
+
+// OAuthScopeType is the type of OAuth scope.
+type OAuthScopeType string
+
+// List of possible values for OAuth scopes.
+const (
+	OAuthScopeTypeReadOnly  OAuthScopeType = "read_only"
+	OAuthScopeTypeReadWrite OAuthScopeType = "read_write"
+)
+
+// OAuthTokenType is the type of token. This will always be "bearer."
+type OAuthTokenType string
+
+// List of possible OAuthTokenType values.
+const (
+	OAuthTokenTypeBearer OAuthTokenType = "bearer"
+)
+
+// OAuthStripeUserBusinessType is the business type for the Stripe oauth user.
+type OAuthStripeUserBusinessType string
+
+// List of supported values for business type.
+const (
+	OAuthStripeUserBusinessTypeCorporation OAuthStripeUserBusinessType = "corporation"
+	OAuthStripeUserBusinessTypeLLC         OAuthStripeUserBusinessType = "llc"
+	OAuthStripeUserBusinessTypeNonProfit   OAuthStripeUserBusinessType = "non_profit"
+	OAuthStripeUserBusinessTypePartnership OAuthStripeUserBusinessType = "partnership"
+	OAuthStripeUserBusinessTypeSoleProp    OAuthStripeUserBusinessType = "sole_prop"
+)
+
+// OAuthStripeUserGender of the person who will be filling out a Stripe
+// application. (International regulations require either male or female.)
+type OAuthStripeUserGender string
+
+// The gender of the person who  will be filling out a Stripe application.
+// (International regulations require either male or female.)
+const (
+	OAuthStripeUserGenderFemale OAuthStripeUserGender = "female"
+	OAuthStripeUserGenderMale   OAuthStripeUserGender = "male"
+)
+
+// OAuthStripeUserParams for the stripe_user OAuth Authorize params.
+type OAuthStripeUserParams struct {
+	BlockKana          *string `form:"block_kana"`
+	BlockKanji         *string `form:"block_kanji"`
+	BuildingKana       *string `form:"building_kana"`
+	BuildingKanji      *string `form:"building_kanji"`
+	BusinessName       *string `form:"business_name"`
+	BusinessType       *string `form:"business_type"`
+	City               *string `form:"city"`
+	Country            *string `form:"country"`
+	Currency           *string `form:"currency"`
+	DOBDay             *int64  `form:"dob_day"`
+	DOBMonth           *int64  `form:"dob_month"`
+	DOBYear            *int64  `form:"dob_year"`
+	Email              *string `form:"email"`
+	FirstName          *string `form:"first_name"`
+	FirstNameKana      *string `form:"first_name_kana"`
+	FirstNameKanji     *string `form:"first_name_kanji"`
+	Gender             *string `form:"gender"`
+	LastName           *string `form:"last_name"`
+	LastNameKana       *string `form:"last_name_kana"`
+	LastNameKanji      *string `form:"last_name_kanji"`
+	PhoneNumber        *string `form:"phone_number"`
+	PhysicalProduct    *bool   `form:"physical_product"`
+	ProductDescription *string `form:"product_description"`
+	State              *string `form:"state"`
+	StreetAddress      *string `form:"street_address"`
+	URL                *string `form:"url"`
+	Zip                *string `form:"zip"`
+}
+
+// AuthorizeURLParams for creating OAuth AuthorizeURLs.
+type AuthorizeURLParams struct {
+	Params                `form:"*"`
+	AlwaysPrompt          *bool                  `form:"always_prompt"`
+	ClientID              *string                `form:"client_id"`
+	RedirectURI           *string                `form:"redirect_uri"`
+	ResponseType          *string                `form:"response_type"`
+	Scope                 *string                `form:"scope"`
+	State                 *string                `form:"state"`
+	StripeLanding         *string                `form:"stripe_landing"`
+	StripeUser            *OAuthStripeUserParams `form:"stripe_user"`
+	SuggestedCapabilities []*string              `form:"suggested_capabilities"`
+
+	// Express is not sent as a parameter, but is used to modify the authorize URL
+	// path to use the express OAuth path.
+	Express *bool `form:"-"`
+}
+
+// DeauthorizeParams for deauthorizing an account.
+type DeauthorizeParams struct {
+	Params       `form:"*"`
+	ClientID     *string `form:"client_id"`
+	StripeUserID *string `form:"stripe_user_id"`
+}
+
+// OAuthTokenParams is the set of paramaters that can be used to request
+// OAuthTokens.
+type OAuthTokenParams struct {
+	Params             `form:"*"`
+	AssertCapabilities []*string `form:"assert_capabilities"`
+	ClientSecret       *string   `form:"client_secret"`
+	Code               *string   `form:"code"`
+	GrantType          *string   `form:"grant_type"`
+	RefreshToken       *string   `form:"refresh_token"`
+	Scope              *string   `form:"scope"`
+}
+
+// OAuthToken is the value of the OAuthToken from OAuth flow.
+// https://stripe.com/docs/connect/oauth-reference#post-token
+type OAuthToken struct {
+	Livemode     bool           `json:"livemode"`
+	Scope        OAuthScopeType `json:"scope"`
+	StripeUserID string         `json:"stripe_user_id"`
+	TokenType    OAuthTokenType `json:"token_type"`
+
+	// Deprecated, please use StripeUserID
+	AccessToken          string `json:"access_token"`
+	RefreshToken         string `json:"refresh_token"`
+	StripePublishableKey string `json:"stripe_publishable_key"`
+}
+
+// Deauthorize is the value of the return from deauthorizing.
+// https://stripe.com/docs/connect/oauth-reference#post-deauthorize
+type Deauthorize struct {
+	StripeUserID string `json:"stripe_user_id"`
+}
--- a/oauth/client.go
+++ b/oauth/client.go
@@ -1,0 +1,77 @@
+// Package oauth provides the OAuth APIs
+package oauth
+
+import (
+	"fmt"
+	"net/http"
+
+	stripe "github.com/stripe/stripe-go"
+	"github.com/stripe/stripe-go/form"
+)
+
+// Client is used to invoke /oauth and related APIs.
+type Client struct {
+	B   stripe.Backend
+	Key string
+}
+
+// AuthorizeURL builds an OAuth authorize URL.
+func AuthorizeURL(params *stripe.AuthorizeURLParams) string {
+	return getC().AuthorizeURL(params)
+}
+
+// AuthorizeURL builds an OAuth authorize URL.
+func (c Client) AuthorizeURL(params *stripe.AuthorizeURLParams) string {
+	express := ""
+	if stripe.BoolValue(params.Express) {
+		express = "/express"
+	}
+	qs := &form.Values{}
+	form.AppendTo(qs, params)
+	return fmt.Sprintf(
+		"%s%s/oauth/authorize?%s",
+		stripe.ConnectURL,
+		express,
+		qs.Encode(),
+	)
+}
+
+// New creates an OAuth token using a code after successful redirection back.
+func New(params *stripe.OAuthTokenParams) (*stripe.OAuthToken, error) {
+	return getC().New(params)
+}
+
+// New creates an OAuth token using a code after successful redirection back.
+func (c Client) New(params *stripe.OAuthTokenParams) (*stripe.OAuthToken, error) {
+	// client_secret is sent in the post body for this endpoint.
+	if stripe.StringValue(params.ClientSecret) == "" {
+		params.ClientSecret = stripe.String(stripe.Key)
+	}
+
+	oauthToken := &stripe.OAuthToken{}
+	err := c.B.Call(http.MethodPost, "/oauth/token", c.Key, params, oauthToken)
+
+	return oauthToken, err
+}
+
+// Del deauthorizes a connected account.
+func Del(params *stripe.DeauthorizeParams) (*stripe.Deauthorize, error) {
+	return getC().Del(params)
+}
+
+// Del deauthorizes a connected account.
+func (c Client) Del(params *stripe.DeauthorizeParams) (*stripe.Deauthorize, error) {
+	deauthorization := &stripe.Deauthorize{}
+	err := c.B.Call(
+		http.MethodPost,
+		"/oauth/deauthorize",
+		c.Key,
+		params,
+		deauthorization,
+	)
+	return deauthorization, err
+}
+
+func getC() Client {
+	return Client{stripe.GetBackend(stripe.ConnectBackend), stripe.Key}
+}
--- a/oauth/client_test.go
+++ b/oauth/client_test.go
@@ -1,0 +1,263 @@
+package oauth
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	stripe "github.com/stripe/stripe-go"
+	_ "github.com/stripe/stripe-go/testing"
+)
+
+func TestAuthorizeURL(t *testing.T) {
+	url := AuthorizeURL(&stripe.AuthorizeURLParams{})
+
+	assert.Equal(t, url, "https://connect.stripe.com/oauth/authorize?")
+}
+
+func TestAuthorizeURLWithOptionalArgs(t *testing.T) {
+	url := AuthorizeURL(&stripe.AuthorizeURLParams{
+		ClientID:              stripe.String("ca_123"),
+		State:                 stripe.String("test-state"),
+		Scope:                 stripe.String("read_only"),
+		RedirectURI:           stripe.String("https://t.example.com"),
+		ResponseType:          stripe.String("test-code"),
+		StripeLanding:         stripe.String("register"),
+		AlwaysPrompt:          stripe.Bool(true),
+		Express:               stripe.Bool(true),
+		SuggestedCapabilities: stripe.StringSlice([]string{"card_payments"}),
+	})
+
+	assert.Contains(t, url, "https://connect.stripe.com/express/oauth/authorize?")
+	assert.Contains(t, url, "client_id=ca_123")
+	assert.Contains(t, url, "redirect_uri=https%3A%2F%2Ft.example.com")
+	assert.Contains(t, url, "response_type=test-code")
+	assert.Contains(t, url, "scope=read_only")
+	assert.Contains(t, url, "state=test-state")
+	assert.Contains(t, url, "stripe_landing=register")
+	assert.Contains(t, url, "always_prompt=true")
+	assert.Contains(t, url, "suggested_capabilities[0]=card_payments")
+}
+
+func TestAuthorizeURLWithStripeUser(t *testing.T) {
+	url := AuthorizeURL(&stripe.AuthorizeURLParams{
+		ResponseType: stripe.String("test-code"),
+		StripeUser: &stripe.OAuthStripeUserParams{
+			BlockKana:          stripe.String("block-kana"),
+			BlockKanji:         stripe.String("block-kanji"),
+			BuildingKana:       stripe.String("building-kana"),
+			BuildingKanji:      stripe.String("building-kanji"),
+			BusinessName:       stripe.String("b-name"),
+			BusinessType:       stripe.String("llc"),
+			City:               stripe.String("Elko"),
+			Country:            stripe.String("US"),
+			Currency:           stripe.String("USD"),
+			DOBDay:             stripe.Int64(15),
+			DOBMonth:           stripe.Int64(10),
+			DOBYear:            stripe.Int64(2019),
+			Email:              stripe.String("test@example.com"),
+			FirstName:          stripe.String("first-name"),
+			FirstNameKana:      stripe.String("first-name-kana"),
+			FirstNameKanji:     stripe.String("first-name-kanji"),
+			Gender:             stripe.String("female"),
+			LastName:           stripe.String("last-name"),
+			LastNameKana:       stripe.String("last-name-kana"),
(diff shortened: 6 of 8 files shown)


oauth/client.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (c Client) AuthorizeURL(params *stripe.AuthorizeURLParams) string {
+  var buf bytes.Buffer
+  buf.WriteString(stripe.CONNECTURL)
+
+  if stripe.BoolValue(params.Express) {
+    buf.WriteString("/express/oauth/authorize")
+  } else {
+    buf.WriteString("/oauth/authorize")
```
remi-stripe: I don't think we use `WriteString` in the rest of the code. We use `stripe.FormatURLPath` instead: https://github.com/stripe/stripe-go/blob/master/card/client.go#31 which prepends `api.stripe.com` but we should find a way to re-use this logic. This avoids escaping issues and such, especially for the GET params, which is a new concept.

oauth.go
```diff
@@ -0,0 +1,52 @@
...
+// Params for creating OAuth AuthorizeURL's.
+type AuthorizeURLParams struct {
+  State           *string
+  ResponseType    *string
+  Scope           *string
+  RedirectURI     *string
+  StripeLanding   *string
+  AlwaysPrompt    *bool
+  Express         *bool
+}
```
remi-stripe: You're missing `stripe_user`. I am not a fan of implementing this, but @ob-stripe added this to .NET and so we likely will need to do the same in go for consistency
Params: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Services/OAuth/OAuthAuthorizeUrlStripeUserOptions.cs
PR: https://github.com/stripe/stripe-dotnet/pull/1542

oauth.go
```diff
@@ -0,0 +1,52 @@
...
+  Code           *string  `form:"code"`
+  RefreshToken   *string  `form:"refresh_token"`
+  Scope          *string  `form:"scope"`
+  ClientSecret   *string  `form:"client_secret"`
+}
+
+// OAuthToken is the value of the OAuthToken from OAuth flow.
+// https://stripe.com/docs/connect/oauth-reference#post-token
+type OAuthToken struct {
+  AccessToken            string          `json:"access_token"`
```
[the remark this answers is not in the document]
cjavilla-stripe: I didn't see which of these were actually deprecated in the docs. Are they officially deprecated, yet?

oauth/client.go
cjavilla-stripe: Switched to use FormatURLPath, but hacked it a little to get the connect domain and fit with `/express` because `/express` url encodes to `%2fexpress`

oauth.go
cjavilla-stripe: Will do :)

oauth/client.go
cjavilla-stripe: Switched to using out form module to build the URL query string from a params struct.

nirajjayantbolt:
Excited to get to use this soon! Is there a rough timeline for when this can get in?

oauth.go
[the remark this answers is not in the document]
cjavilla-stripe: Just wanted to bump this. I was looking here: https://stripe.com/docs/connect/oauth-reference#post-token-response and didn't see any of these deprecated. Thinking we should hide the `refresh_token`, `access_token`, `stripe_publishable_key` ? or mark as deprecated?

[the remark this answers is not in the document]
ob-stripe: `refresh_token` and `access_token` are part of the OAuth specification so it would be awkward to not have them at all, but we can mark them as "deprecated" in favor of `stripe_user_id`.

We should probably be consistent with the .NET implementation: https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/OAuth/OAuthToken.cs

cjavilla-stripe:
Hmm. tests are passing locally... Investigating

cjavilla-stripe:
Odd. I didn't have `golint` installed and `make test` didn't yell or anything. Once I installed `golint` it was clear what needed to be fixed. I should've looked at the Travis output before that last comment :)

brandur-stripe:
> Odd. I didn't have `golint` installed and `make test` didn't yell or anything.

Lint's in Make, but it's a separate step (`make lint`). If you want to run it, I'd suggest just doing a `make`, which will run `all`, and run all the checks that we've bundled in there.

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+	URL                *string                     `form:"url"`
+	Zip                *string                     `form:"zip"`
+}
+
+// AuthorizeURLParams for creating OAuth AuthorizeURL's.
+type AuthorizeURLParams struct {
+	Params        `form:"*"`
+	AlwaysPrompt  *bool                  `form:"always_prompt"`
+	ClientID      *string                `form:"client_id"`
+	Express       *bool                  `form:"-"`
```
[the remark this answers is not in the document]
brandur-stripe: Ah, I see from reading the code below now. Do you think you could move this to a different "section" below the other parameters here and put a comment on top of it to something of the effect "Express is not sent as a parameter, but may modify the target URL of an authorize request".

(Just when reading this it's not obvious that it's even going to be used without reading the accompanying client code which is in a different file.)

stripe.go
```diff
@@ -35,6 +35,13 @@ const (
 	// APIURL is the URL of the API service backend.
 	APIURL string = "https://api.stripe.com"
 
+	// CONNECTURL is the URL for OAuth.
+	CONNECTURL string = "https://connect.stripe.com"
```
brandur-stripe: Haha, so admittedly this is not obvious at all. But I think `APIURL` was actually trying to follow normal naming conventions, but just ended up awkward because both parts were an acronym. (Maybe it should've been `ApiURL` instead in this case? Not sure, but whatever.)

So in this case let's go with `ConnectURL` here instead.

stripe.go
```diff
@@ -62,6 +69,9 @@ var EnableTelemetry = true
 // Key is the Stripe API key used globally in the binding.
 var Key string
 
+// ClientID is the Stripe Client ID used globally in the binding.
```
brandur-stripe: Could you elaborate on this comment a bit by saying that: all OAuth requests will use this `ClientID` by default unless the client ID specific request's parameter is overridden.

Also, I'm not 100% convinced that we need this one — it wouldn't that much worse for users to just keep `ClientID` in their own global config somewhere and just pass it in whenever they need a Stripe OAuth requests (of which there hopefully won't be that many).

cc @remi-stripe For thoughts as well.

oauth/client.go
```diff
@@ -0,0 +1,84 @@
...
+
+// New creates the OAuthToken using a code after successful redirection back.
+func (c Client) New(params *stripe.OAuthTokenParams) (*stripe.OAuthToken, error) {
+	// client_secret is sent in the post body for this endpoint.
+	if stripe.StringValue(params.ClientSecret) == "" {
+		params.ClientSecret = stripe.String(stripe.Key)
+	}
+	if stripe.StringValue(params.GrantType) == "" {
+		params.GrantType = stripe.String("authorization_code")
+	}
```
brandur-stripe: OOC, is there some particular rationale behind providing a parameter default here?

It may be a little more convenient I suppose, but I think our normal convention is just allowing the API request to fail the first time and signal back that `grant_type=authorization_code` should have been passed, at which point the user just corrects their implementation. It's one extra step, but they probably won't ever make the mistake again.

oauth/client.go
```diff
@@ -0,0 +1,84 @@
...
+		qs.EncodeValues(),
+	)
+}
+
+// New creates the OAuthToken using a code after successful redirection back.
+func New(params *stripe.OAuthTokenParams) (*stripe.OAuthToken, error) {
+	return getC().New(params)
+}
+
+// New creates the OAuthToken using a code after successful redirection back.
```
brandur-stripe: Could we use the "word style" name of the resource here? Like:

``` go
// New creates an OAuth token ...
```

oauth/client.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+// Client is used to invoke /oauth and related APIs.
+type Client struct {
+	B   stripe.Backend
+	Key string
+}
+
+// AuthorizeURL builds the OAuth authorize URL.
+func AuthorizeURL(params *stripe.AuthorizeURLParams) string {
```
brandur-stripe: Will `AuthorizeURL` ever be needed by a user integration?

If not, my vote is that we keep it unexported (i.e. rename to `authorizeURL`). The tests will still be able to get at it because they're in the same package.

oauth/client_test.go
```diff
@@ -0,0 +1,234 @@
+package oauth
+
+import (
+	"bytes"
+	assert "github.com/stretchr/testify/require"
+	stripe "github.com/stripe/stripe-go"
+	_ "github.com/stripe/stripe-go/testing"
+	"io/ioutil"
+	"net/http"
+	"testing"
```
brandur-stripe: This may be a local stylistic convention, but would you mind breaking the imports into two groups? That's generally how we've done it here.

1. Al the "built-ins" like `bytes`, `io/ioutil`, etc.
2. All the GOPATH packages like `assert "github.com/stretchr/testify/require"`, `stripe "github.com/stripe/stripe-go"`, `_ "github.com/stripe/stripe-go/testing"`.

(You can find examples in most other files in the project.)

Gofmt is smart enough to support this fine.

oauth/client.go
```diff
@@ -0,0 +1,84 @@
+// Package oauth provides the OAuth APIs
+package oauth
+
+import (
+	"fmt"
+	stripe "github.com/stripe/stripe-go"
+	"github.com/stripe/stripe-go/form"
+	"net/http"
```
brandur-stripe: And same comment as in `client_test.go` about reorganizing imports into two groups.

oauth/client.go
cjavilla-stripe: Yeah this is a tool for users to create their AuthorizeURL that they will redirect to.

brandur-stripe: Ah yes, now that I think about how OAuth works, that makes sense :)

cjavilla-stripe:
Thanks for the review and your patience, @brandur-stripe :)

client/api.go
```diff
@@ -148,6 +149,8 @@ type API struct {
 	IssuingTransactions *transaction.Client
 	// LoginLinks is the client used to invoke login link related APIs.
 	LoginLinks *loginlink.Client
+	// OAuth is the client used to invoke /oauth APIs.
+	OAuth *oauth.Client
```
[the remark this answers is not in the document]
cjavilla-stripe: I believe I've initialized correctly below.

form/form.go
```diff
@@ -538,14 +538,15 @@ func (f *Values) Add(key, val string) {
...
+// Encode encodes the keys and values into “URL encoded” form
+// ("bar=baz&foo=quux").
 func (f *Values) Encode() string {
 	var buf bytes.Buffer
 	for _, v := range f.values {
 		if buf.Len() > 0 {
 			buf.WriteByte('&')
 		}
-		buf.WriteString(url.QueryEscape(v.Key))
+		buf.WriteString(v.Key)
```
cjavilla-stripe: Switching to this form without an escape worked fine. I thought this might be a breaking change, which is why I initially started with the second encoding method. I think we control the keys tho so this _should_ be fine.

oauth/client.go
cjavilla-stripe: Yeah it was all about convenience. Trying to make it so that users of this can pass some basics and we send sensible defaults.

oauth/client_test.go
cjavilla-stripe: Got it. Thanks for the heads up :)

form/form.go
```diff
@@ -498,6 +499,21 @@ func (f *Values) Encode() string {
 	return buf.String()
 }
 
+// EncodeValues encodes only the values into “URL encoded” form
+// ("bar=baz&foo=quux") so that we can have keys like `stripe_user[email]=test`
+func (f *Values) EncodeValues() string {
```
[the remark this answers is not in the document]
cjavilla-stripe: I thought that not escaping the key would be considered a breaking change, so opted initially for a second encode method, but I think we control the keys here so not escaping the key seems fine.

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+const (
+	OAuthStripeUserBusinessTypeCorporation OAuthStripeUserBusinessType = "corporation"
+	OAuthStripeUserBusinessTypeLLC         OAuthStripeUserBusinessType = "llc"
+	OAuthStripeUserBusinessTypeNonProfit   OAuthStripeUserBusinessType = "non_profit"
+	OAuthStripeUserBusinessTypePartnership OAuthStripeUserBusinessType = "partnership"
+	OAuthStripeUserBusinessTypeSoleProp    OAuthStripeUserBusinessType = "sole_prop"
+)
+
+// OAuthStripeUserGender of the person who will be filling out a Stripe
+// application.  (International regulations require either male or female.)
```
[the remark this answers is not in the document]
cjavilla-stripe: Done

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+	BuildingKana       *string                     `form:"building_kana"`
+	BuildingKanji      *string                     `form:"building_kanji"`
+	BusinessName       *string                     `form:"business_name"`
+	BusinessType       OAuthStripeUserBusinessType `form:"business_type"`
+	City               *string                     `form:"city"`
+	Country            *string                     `form:"country"`
+	Currency           *string                     `form:"currency"`
+	DobDay             uint64                      `form:"dob_day"`
+	DobMonth           uint64                      `form:"dob_month"`
+	DobYear            uint64                      `form:"dob_year"`
```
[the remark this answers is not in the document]
cjavilla-stripe: I think I got them all updated to pointers.

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+	Params        `form:"*"`
+	AlwaysPrompt  *bool                  `form:"always_prompt"`
+	ClientID      *string                `form:"client_id"`
+	Express       *bool                  `form:"-"`
+	RedirectURI   *string                `form:"redirect_uri"`
+	ResponseType  *string                `form:"response_type"`
+	Scope         *string                `form:"scope"`
+	State         *string                `form:"state"`
+	StripeLanding *string                `form:"stripe_landing"`
+	StripeUser    *OAuthStripeUserParams `form:"stripe_user"`
```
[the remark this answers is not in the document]
cjavilla-stripe: Yeah good call! I missed that on the first pass through the docs.

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+	Params       `form:"*"`
+	ClientID     *string `form:"client_id"`
+	StripeUserID *string `form:"stripe_user_id"`
+}
+
+// OAuthTokenParams is the set of paramaters that can be used to request
+// OAuthTokens.
+type OAuthTokenParams struct {
+	Params       `form:"*"`
+	ClientSecret *string `form:"client_secret"`
```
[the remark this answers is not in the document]
cjavilla-stripe: ✔️

stripe.go
cjavilla-stripe: When reviewing this, I don't think this is actually used anywhere and would be a convenient reference for library users, but probably not that valuable. Think I should just kill it?

stripe.go
cjavilla-stripe: Updated the comment. Again, keeping this here and in params was for convenience.

cjavilla-stripe:
> Excited to get to use this soon! Is there a rough timeline for when this can get in?

@nirajjayantbolt  hopefully soon! sorry for the delay 😄

oauth/client_test.go
cjavilla-stripe: I think I got all of these cleaned up.

client/api.go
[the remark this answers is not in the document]
brandur-stripe: Yep looks good.

form/form.go
brandur-stripe: Hm, actually now that I think about it, we probably do want to keep the escape — most keys should be safe, but there are a couple free-form areas in the API like `metadata` where we could have arbitrary input passed in.

Would you mind bringing back the `QueryEscape` and adopt the solution I suggested from stripe-ruby (replace encoded square brackets with real square brackets)? I believe that should be pretty safe regardless of use in form body or query string, and allows us to keep the API simple by having only one implementation. Sorry for the churn.

stripe.go
```diff
@@ -35,6 +35,13 @@ const (
 	// APIURL is the URL of the API service backend.
 	APIURL string = "https://api.stripe.com"
 
+	// ConnectURL is the URL for OAuth.
+	ConnectURL string = "https://connect.stripe.com"
```
brandur-stripe: So now that I look at this again, I think the duplicate constant between `APIURL` and `apiURL` was basically a refactoring mistake, and really doesn't help anything particularly. I think it's somewhat convenient that these URLs are exported in case users want them, so would you mind just dropping `connectURL` in favor `ConnectURL`?

oauth/client.go
```diff
@@ -0,0 +1,85 @@
...
+	if params.ClientID == nil {
+		params.ClientID = stripe.String(stripe.ClientID)
+	}
+	if params.ResponseType == nil {
+		params.ResponseType = stripe.String("code")
+	}
+	qs := &form.Values{}
+	form.AppendTo(qs, params)
+	return fmt.Sprintf(
+		"https://connect.stripe.com%s/oauth/authorize?%s",
```
brandur-stripe: And can we can change this to reuse `stripe.ConnectURL`?

stripe.go
```diff
@@ -62,6 +69,11 @@ var EnableTelemetry = true
 // Key is the Stripe API key used globally in the binding.
 var Key string
 
+// ClientID is the Stripe Client ID used by default for OAuth requests.
+// Relevant OAuth parameter types can also be initialized with a specific
+// ClientID that will take precidence over this global ClientID.
```
brandur-stripe: Minor spelling error: "precedence".

stripe.go
```diff
@@ -62,6 +69,11 @@ var EnableTelemetry = true
...
+var ClientID string
```
brandur-stripe: @remi-stripe Any opinion on having a global `ClientID` versus just requiring that one be passed with parameters? I don't feel too strongly about it, but in general would prefer to avoid globals where possible, so I'd lean toward the latter.

brandur-stripe:
Oops, and realized I forgot to answer your original questions:

> Is this PR too big?

Definitely big-ish, but most of our OAuth implementations in other languages came in as part of one big PR, so it's not that unusual.

> Is this a decent param passing approach? (I'm about 95% confident that I understand the reason for using pointers)

Yeah, it was best to just follow the convention and use pointers, so you did the right thing.

The reason for pointer is that because variables in Go by default get their zero value (so a string defaults to `""`, a boolean defaults to `false`, etc.), without the pointer it's not possible to distinguish between a field that hasn't been set versus one that's been set explicitly to the zero value.

This doesn't make that much of a difference a lot of the time, but there are a couple places where the zero values are important. For example, we might want to explicitly set a string in the API to an empty string to empty it, or a boolean in the API to a `false` value, or an integer in the API to `0`. Before in stripe-go, this was only possible through some pretty gross workarounds, so we ended up switching to pointers everywhere.

> Is stubbing out the server like this ok to test the oauth flow since oauth isnt supported by stripe-mock?

Yeah, what you did there seems sane to me. Thanks for testing!

oauth.go
```diff
@@ -0,0 +1,154 @@
...
+	BuildingKana       *string `form:"building_kana"`
+	BuildingKanji      *string `form:"building_kanji"`
+	BusinessName       *string `form:"business_name"`
+	BusinessType       *string `form:"business_type"`
+	City               *string `form:"city"`
+	Country            *string `form:"country"`
+	Currency           *string `form:"currency"`
+	DOBDay             *uint64 `form:"dob_day"`
+	DOBMonth           *uint64 `form:"dob_month"`
+	DOBYear            *uint64 `form:"dob_year"`
```
brandur-stripe: And oops, actually if you wouldn't mind, these should be `*int64` actually. They can't be negative so technically `uint64` would be a better fit, but at one point we just decided to just go with `int64` everywhere by convention — in practice we'll never get to numbers high enough that we need the extra range `uint64` gets us, and using `int64` gives us some forward compatibility in case a field that currently doesn't allow negative numbers allows them in the future.

oauth/client_test.go
```diff
@@ -0,0 +1,242 @@
...
+			BuildingKana:       stripe.String("building-kana"),
+			BuildingKanji:      stripe.String("building-kanji"),
+			BusinessName:       stripe.String("b-name"),
+			BusinessType:       stripe.String("llc"),
+			City:               stripe.String("Elko"),
+			Country:            stripe.String("US"),
+			Currency:           stripe.String("USD"),
+			DOBDay:             &dobDay,
+			DOBMonth:           &dobMonth,
+			DOBYear:            &dobYear,
```
brandur-stripe: And switching those to `int64` will also allow you to use the `string.Int64` helper down here.

oauth.go
```diff
@@ -0,0 +1,149 @@
...
+	GrantType    *string `form:"grant_type"`
+	RefreshToken *string `form:"refresh_token"`
+	Scope        *string `form:"scope"`
+}
+
+// OAuthToken is the value of the OAuthToken from OAuth flow.
+// https://stripe.com/docs/connect/oauth-reference#post-token
+type OAuthToken struct {
+	Error            OAuthError     `json:"error"`
+	ErrorDescription string         `json:"error_description"`
```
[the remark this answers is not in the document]
brandur-stripe: (And just flagging this comment once more in case it was missed originally.)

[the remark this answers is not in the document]
cjavilla-stripe: Yeah I've been trying to work through this one... The shape of the errors from OAuth are different from the rest of the API. Our error handling / wrapping is failing to unmarshal because the key "error" in the OAuth errors is the type of error and the key "error" in the rest of the API seems to be the json object with error data.

Feels like the right approach is to make the connect backend handle the error with a different error type?

oauth/client_test.go
cjavilla-stripe: Sweet. I was wondering why that type didn't have a helper.

form/form.go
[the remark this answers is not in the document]
cjavilla-stripe: Solved it this way. seems to work well :)

form/form.go
cjavilla-stripe: Sounds good :)

cjavilla-stripe:
Thanks for that explanation about our pointer use, @brandur-stripe. Remi also shared that with me and seeing those examples really drives it home. 

I think there are only two major things that need review:

- [ ] error handling - My current approach is to (1) attempt to deserialize json with the current struct object (top level error key pointing at error data, which is the common error shape), if that fails (2) attempt to deserialize with the top level Error struct type which has 2 new fields for OAuth specific errors. (https://github.com/stripe/stripe-go/pull/879/files#diff-faeee970b41138d8bdc51e7aa3601141)
- [ ] global client ID - @remi-stripe thoughts on global vs requiring that at call time? (https://github.com/stripe/stripe-go/pull/879/files#diff-ef7561d9a44d35bccb615dd5adb151be)

brandur-stripe:
> My current approach is to (1) attempt to deserialize json with the current struct object (top level error key pointing at error data, which is the common error shape), if that fails (2) attempt to deserialize with the top level Error struct type which has 2 new fields for OAuth specific errors.

Thanks for changing! I was thinking about it some more, and I think this is definitely closer to the path we want. Having a single unified error object for API or OAuth is how it works in stripe-dotnet already [1], and how it's being proposed for stripe-ruby as well [2].

I think it's a little janky that we have to do that two-phase fallback for deserializing an error object (although definitely not the worse thing, and the comment helps a lot). Did you consider alternative a custom `UnmarshalJSON` for the `Error` type? We might be able to have it try to detect an API error versus OAuth error, and then pick one path or the other based off of type.

Anyway, sorry that this is turning into more work than you probably expected :) The good news is that I think we're pretty close at this point.

[1] https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/StripeError.cs#L88-L90
[2] https://github.com/stripe/stripe-ruby/pull/811/files#diff-30e2a416d16a9b8714284f021b0a68e0R51

brandur-stripe:
[ ] And maybe the one thing we might still want to review is the use of default values for unspecified parameters in OAuth requests. I don't think we do this for any other object types, so my preference would be to lean against the pattern even if there may be some usability benefit.

One pitfall that I can see possibly happening is that the user doesn't have control anymore over whether the parameter is sent. Say they purposely didn't want to send a particular parameter, they now no longer have a great way to do that (they could do `stripe.String("")`, but I don't think we should make them resort to that).

cjavilla-stripe:
Removed the remaining defaults for unspecified values, removed the global ClientID, and took another swag at unmarshaling the error data by forking based on the BackendImplementation type.

r? @brandur-stripe

stripe.go
cjavilla-stripe: We chatted via slack and he agreed with you, so I've removed :)

client/api.go
```diff
@@ -272,8 +276,9 @@ func (a *API) Init(key string, backends *stripe.Backends) {
 	a.IssuingDisputes = &issuingdispute.Client{B: backends.API, Key: key}
 	a.IssuingTransactions = &transaction.Client{B: backends.API, Key: key}
 	a.LoginLinks = &loginlink.Client{B: backends.API, Key: key}
-	a.Orders = &order.Client{B: backends.API, Key: key}
+	a.OAuth = &oauth.Client{B: backends.Connect, Key: key}
 	a.OrderReturns = &orderreturn.Client{B: backends.API, Key: key}
+	a.Orders = &order.Client{B: backends.API, Key: key}
```
brandur-stripe: Thanks for fixing :)

error.go
```diff
@@ -136,6 +136,10 @@ type Error struct {
 	SetupIntent    *SetupIntent   `json:"setup_intent,omitempty"`
 	Source         *PaymentSource `json:"source,omitempty"`
 	Type           ErrorType      `json:"type"`
+
+	// OAuth specific Error properties. Named OAuthError because of name conflict.
+	OAuthError            string `json:"error,omitempty"`
+	OAuthErrorDescription string `json:"error_description,omitempty"`
```
brandur-stripe: I actually like the "OAuth" prefix anyway — it makes it harder to use these outside of an OAuth context by accident (it's clear they're specifically for OAuth and nothing else).

oauth/client_test.go
```diff
@@ -0,0 +1,267 @@
...
+			LastName:           stripe.String("last-name"),
+			LastNameKana:       stripe.String("last-name-kana"),
+			LastNameKanji:      stripe.String("last-name-kanji"),
+			PhoneNumber:        stripe.String("999-999-9999"),
+			PhysicalProduct:    stripe.Bool(false),
+			ProductDescription: stripe.String("product-description"),
+			State:              stripe.String("NV"),
+			StreetAddress:      stripe.String("123 main"),
+			URL:                stripe.String("http://example.com"),
+			Zip:                stripe.String("12345"),
```
brandur-stripe: I just realized that "Zip" is actually an acronym (Zone Improvement Plan) and probably should've been "ZIP" instead. Don't change it though because we've already got the "Zip" convention in some fields elsewhere.

stripe.go
```diff
@@ -450,9 +457,20 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 
 // ResponseToError converts a stripe response to an Error.
 func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []byte) error {
+
 	var raw rawError
-	if err := s.UnmarshalJSONVerbose(res.StatusCode, resBody, &raw); err != nil {
-		return err
+	if s.Type == ConnectBackend {
```
brandur-stripe: Yeah, I like this new deserialization logic better — feels cleaner. Thanks for changing.

stripe.go
```diff
@@ -450,9 +457,20 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 
 // ResponseToError converts a stripe response to an Error.
 func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []byte) error {
+
```
brandur-stripe: Ultra minor, but would you mind taking out this extraneously added newline?

brandur-stripe:
@cjavilla-stripe New changes look great! Would you mind squashing this down? After that I think we're pretty much good to go.

@remi-stripe This looks ready to me. Do you want to take a pass for yourself before we go to release?

error.go
cjavilla-stripe: This was actually required because we use Error as a method on this object elsewhere.

stripe.go
cjavilla-stripe: Done.

oauth/client_test.go
cjavilla-stripe: Got it. Yeah tried to follow convention :)

oauth/client_test.go
```diff
@@ -0,0 +1,267 @@
...
+	assert.Contains(t, url, "stripe_user[phone_number]=999-999-9999")
+	assert.Contains(t, url, "stripe_user[physical_product]=false")
+	assert.Contains(t, url, "stripe_user[product_description]=product-description")
+	assert.Contains(t, url, "stripe_user[state]=NV")
+	assert.Contains(t, url, "stripe_user[street_address]=123+main")
+	assert.Contains(t, url, "stripe_user[url]=http%3A%2F%2Fexample.com")
+	assert.Contains(t, url, "stripe_user[zip]=12345")
+}
+
+// RoundTripFunc.
```
brandur-stripe: Can you either unexport this or put together a more complete comment?

oauth/client_test.go
```diff
@@ -0,0 +1,267 @@
...
+type RoundTripFunc func(req *http.Request) *http.Response
+
+// RoundTrip.
```
brandur-stripe: And same with this one.

oauth/client_test.go
```diff
@@ -0,0 +1,267 @@
...
+			HTTPClient: httpClient,
+		},
+	)
+	stripe.SetBackend(stripe.ConnectBackend, mockBackend)
+}
+
+func TestNewOAuthToken(t *testing.T) {
+	stripe.Key = "sk_123"
+
+	// stripe-mock doesn't support connect URL's so this stubs out the server.
```
brandur-stripe: Maybe consider changing all instances of "URL's" to "URLs"? I Googled and there's [no clear answer for the correct way to handle these](https://english.stackexchange.com/a/921) (sources vary), but my feeling is that the apostrophe-less version is generally preferred.

cjavilla-stripe: Updated these instances.

oauth/client_test.go
cjavilla-stripe: Unexported.

cjavilla-stripe:
Thanks for the thorough review, @brandur-stripe! :)

brandur-stripe:
Excellent!! Thank you again CJ!

brandur-stripe:
Released as 61.23.0.

cjavilla-stripe:
🎉

PR #880, opened 2019-06-24
Add support for `CollectionMethod` on Billing resources

remi-stripe:
* Add support for `CollectionMethod` on `Invoice`, `Subscription` and `SubscriptionSchedule`
* Add support for `UnifiedProration` on `InvoiceLine`

r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -12,6 +12,7 @@ const (
 )
 
 // InvoiceBilling is the type of billing method for this invoice.
+// This is considered deprecated. Use InvoiceCollectionMethod instead.
 type InvoiceBilling string
 
 // List of values that InvoiceBilling can take.
@@ -44,6 +45,15 @@ const (
 	InvoiceBillingStatusPaid          InvoiceBillingStatus = "paid"
 	InvoiceBillingStatusUncollectible InvoiceBillingStatus = "uncollectible"
 	InvoiceBillingStatusVoid          InvoiceBillingStatus = "void"
+)
+
+// InvoiceCollectionMethod is the type of billing method for this invoice.
+type InvoiceCollectionMethod string
+
+// List of values that InvoiceBilling can take.
+const (
+	InvoiceCollectionMethodChargeAutomatically InvoiceCollectionMethod = "charge_automatically"
+	InvoiceCollectionMethodSendInvoice         InvoiceCollectionMethod = "send_invoice"
 )
 
 // InvoiceUpcomingInvoiceItemPeriodParams represents the period associated with that invoice item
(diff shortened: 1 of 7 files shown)


brandur-stripe:
LGTM Remi.

You're probably aware, but the build is failing because the target version of stripe-mock isn't released yet.

ptal @remi-stripe

remi-stripe:
> You're probably aware, but the build is failing because the target version of stripe-mock isn't released yet.

Yeah I was releasing stripe-mock in parallel but hit a snag and I'm now waiting on a review before I can merge. Should be fixed in a few hours!

invoice.go
```diff
@@ -195,9 +211,9 @@ type Invoice struct {
 	AttemptCount                 int64                    `json:"attempt_count"`
 	Attempted                    bool                     `json:"attempted"`
 	AutoAdvance                  bool                     `json:"auto_advance"`
-	Billing                      InvoiceBilling           `json:"billing"`
 	BillingReason                InvoiceBillingReason     `json:"billing_reason"`
 	Charge                       *Charge                  `json:"charge"`
+	CollectionMethod             *InvoiceCollectionMethod `json:"collection_method"`
```
tucksaun: Any reason to move to a pointer here?
I believe the API is always providing this field and this makes the comparison with `InvoiceCollectionMethodSendInvoice` or  `InvoiceCollectionMethodChargeAutomatically` less smooth.
If you confirm this should have stayed a value I can send a PR.

tucksaun: @remi-stripe makes sense.
Though, what about the change on `Billing`? It was a BC break for people upgrading (that's actually how I spot this issue in the first place).
Do you want to revert this in current version or wait the same way?

remi-stripe: @tucksaun which change on `Billing`? We follow semantic versioning so we make all breaking changes in major versions.

tucksaun: @remi-stripe https://github.com/stripe/stripe-go/pull/880/files#diff-7d99b0ab199591cd498ef67d8e82c4e1L198 became a pointer (https://github.com/stripe/stripe-go/pull/880/files#diff-7d99b0ab199591cd498ef67d8e82c4e1R266) in this PR and released as of `v61.7.0`

remi-stripe: @tucksaun Oh wow, I'm sorry about that. This shouldn't have happened. I was just marking the property as deprecated for a future deprecation and the type should definitely not have changed at that point. That also explains why the other type is a pointer when it should not, because I must have set both incorrectly at the same time.

PR #883, opened 2019-06-27
Add support for the `SetupIntent` resource and APIs

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -7,7 +7,7 @@ import (
 // PaymentIntentCancellationReason is the list of allowed values for the cancelation reason.
 type PaymentIntentCancellationReason string
 
-// List of values that PaymentIntentCaptureMethod can take.
+// List of values that PaymentIntentCancellationReason can take.
 const (
 	PaymentIntentCancellationReasonAbandoned           PaymentIntentCancellationReason = "abandoned"
 	PaymentIntentCancellationReasonAutomatic           PaymentIntentCancellationReason = "automatic"
@@ -54,6 +54,15 @@ type PaymentIntentOffSession string
 const (
 	PaymentIntentOffSessionOneOff    PaymentIntentOffSession = "one_off"
 	PaymentIntentOffSessionRecurring PaymentIntentOffSession = "recurring"
+)
+
+// PaymentIntentSetupFutureUsage is the list of allowed values for SetupFutureUsage.
+type PaymentIntentSetupFutureUsage string
+
+// List of values that PaymentIntentSetupFutureUsage can take.
+const (
+	PaymentIntentSetupFutureUsageOffSession PaymentIntentSetupFutureUsage = "off_session"
+	PaymentIntentSetupFutureUsageOnSession  PaymentIntentSetupFutureUsage = "on_session"
 )
 
 // PaymentIntentStatus is the list of allowed values for the payment intent's status.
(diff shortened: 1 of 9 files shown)


paymentintent.go
```diff
@@ -56,6 +56,15 @@ const (
 	PaymentIntentOffSessionRecurring PaymentIntentOffSession = "recurring"
 )
 
+// PaymentIntentSetupFutureUsage is the list of allowed values for SetupFutureUsage.
+type PaymentIntentSetupFutureUsage string
+
+// List of values that PaymentIntentSetupFutureUsage can take.
+const (
+	PaymentIntentSetupFutureUsageOffSession PaymentIntentSetupFutureUsage = "off_session"
+	PaymentIntentSetupFutureUsageOnSession  PaymentIntentSetupFutureUsage = "on_session"
```
brandur-stripe: I'm sure you're well aware, but: It looks like this is still correct according to current docs, but I thought we'd decided to rename in the end.

setupintent.go
```diff
@@ -0,0 +1,117 @@
...
+	LastSetupError     *Error                   `json:"last_setup_error"`
+	Livemode           bool                     `json:"livemode"`
+	Metadata           map[string]string        `json:"metadata"`
+	NextAction         *PaymentIntentNextAction `json:"next_action"`
+	Object             string                   `json:"object"`
+	OnBehalfOf         *Account                 `json:"on_behalf_of"`
+	PaymentMethod      *PaymentMethod           `json:"payment_method"`
+	PaymentMethodTypes []string                 `json:"payment_method_types"`
+	Status             SetupIntentStatus        `json:"status"`
+	Usage              SetupIntentUsage         `json:"usage"`
```
brandur-stripe: Just flagging that you included `usage` which is marked as a preview feature, but omitted `requested_credentials` and `single_use`. Not sure if that was intentional or not.

setupintent.go
```diff
@@ -0,0 +1,117 @@
...
+// UnmarshalJSON handles deserialization of a SetupIntent.
+// This custom unmarshaling is needed because the resulting
+// property may be an id or the full struct if it was expanded.
+func (p *SetupIntent) UnmarshalJSON(data []byte) error {
+	if id, ok := ParseID(data); ok {
+		p.ID = id
+		return nil
+	}
+
+	type paymentintent SetupIntent
```
brandur-stripe: Maybe rename this to `setupIntent` (I'd use a capitalized "I" in "Intent" probably).

paymentintent.go
remi-stripe: Yes, I thought so too but after all we did not. They gavel-ed the original proposal!

brandur-stripe: Ah, weird. SG though.

setupintent.go
remi-stripe: Confirmed internally that this one is gated by mistake and will ship.

setupintent.go
remi-stripe: fixed

remi-stripe:
@brandur-stripe thanks for catching the issues. PTAL

remi-stripe:
self-approving as I just added a new field `cancellation_reason`.

brandur-stripe:
We found some bad logic in this one — reverted in ec2279d and removed from the 61.8.0 release.

PR #894, opened 2019-07-16
Introduce encoding for high precision decimal fields

brandur:
Introduces a strategy for encoding forthcoming "high precision" decimal fields which will soon be seen in the API. These differ from standards floats in the way that the API cares about their fractional values, and that we'll be encoding them as strings in JSON to try and make as faithfully accurate to the original value as possible.

I reused almost all of OB's prototype branch for this, but moved away from manually specified precision as a form option like `precision=10` and over to a simple tag for `high_precision`. The thesis behind the change is that the float formatting precision of `-1`, which attemps to use as few encoded numbers as is possible to produce a string that's parsable back as the original number, is going to be appropriate for all cases that we're likely to see, and taking another dimension of configuration away is more likely to help us maintain uniform conventions.

If it turns out that we need configurable precision at some point down the line, we can bring back the original `precision=` without much trouble. (I'm even leaving OB's commit in Git history.)

r? @ob-stripe Mind taking a look at this?
cc @stripe/api-libraries

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -54,10 +54,21 @@ type formOptions struct {
 	// empty string is a string's zero value and wouldn't normally be encoded.
 	Empty bool
 
-	// Precision controls the number of digits printed for the field's value.
-	// It can only be used for floating-point numbers. The special value -1
-	// uses the smallest number of digits necessary.
-	Precision *int
+	// HighPrecision indicates that this field should be treated as a high
+	// precision decimal, a decimal whose precision is important to the API and
+	// which we want to encode as accurately as possible.
+	//
+	// All parameters are encoded using form encoding, so this of course
+	// encodes a value to a string, but notably, these high precision fields
+	// are sent back as strings in JSON, even though they might be surface as
+	// floats in this library.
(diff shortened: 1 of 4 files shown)


brandur-stripe:
I also removed the changes on plan from OB's original branch in 84728ca. It should be easy to use that as a reference commit for when we want them back.

brandur-stripe:
Also, on using primitive floats versus `big.Float`: I kind of think we might be okay getting away with float for now just given that I suspect few of the numbers going in or out of stripe-go specifically will be calculated. It'll be more like `AmountPrecise: 0.01` or something along those lines.

Would you be okay with a later migration to `big.Float` after it's more clear that the primitive types are causing problems?

ob-stripe:
> Would you be okay with a later migration to `big.Float` after it's more clear that the primitive types are causing problems?

Yeah, that sounds fine.

brandur-stripe:
Excellent! Thanks Olivier.

brandur-stripe:
Released as 61.16.0.

Issue #896, opened 2019-07-18
ErrorCode doesn't cover all of codes listed in official doc


wawoon:
# Background

I'm creating GraphQL API which uses Stripe API to provide payment feature with our client application, and I want to provide ErrorCode as enum in GraphQL API.

Then I investigated the error codes api [document](https://stripe.com/docs/error-codes). But I found some problems.

# What to be fixed

- The `ErrorCode`  doesn't cover all types listed in [docs](https://stripe.com/docs/error-codes)
- The code of "invalid_swipe_data" is not listed in [docs](https://stripe.com/docs/error-codes)

# Solution

I think this problem can be fixed by just adding missing error codes, and removing `invalid_swipe_data` value.

remi-stripe:
@wawoon The list of errors predates the time we shipped that list of error codes. Originally, the API only had a handful of unique error codes mostly tailored for declines and high level errors. That's why the list seems different. The `invalid_swipe_data` is a real error code though it's gated and not accessible by all integrations.

I've opened a PR https://github.com/stripe/stripe-go/pull/897 to add all existing error codes from that documentation

PR #897, opened 2019-07-18
Add all missing error codes

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/896

r? @ob-stripe 
cc @stripe/api-libraries


error.go
```diff
@@ -21,20 +21,100 @@ type ErrorCode string
...
+	ErrorCodeTLSVersionUnsupported                  ErrorCode = "tls_version_unsupported"
+	ErrorCodeTokenAlreadyUsed                       ErrorCode = "token_already_used"
+	ErrorCodeTokenInUse                             ErrorCode = "token_in_use"
+	ErrorCodeTransfersNotAllowed                    ErrorCode = "transfers_not_allowed"
+	ErrorCodeUpstreamOrderCreationFailed            ErrorCode = "upstream_order_creation_failed"
+	ErrorCodeURLInvalid                             ErrorCode = "url_invalid"
+
+	// The following error codes can be returned though are undocumented
+	ErrorCodeIncorrectCVC     ErrorCode = "incorrect_cvc"
+	ErrorCodeInvalidCVC       ErrorCode = "invalid_cvc"
```
ob-stripe: I have no idea how this is even working, but both `ErrorCodeIncorrectCVC` and `ErrorCodeInvalidCVC` are already defined (with the same values) above.

remi-stripe:
Re-assigning to @brandur-stripe as we never finished that one

error.go
```diff
@@ -21,20 +21,98 @@ type ErrorCode string
...
+	ErrorCodePlatformAPIKeyExpired                  ErrorCode = "platform_api_key_expired"
+	ErrorCodePostalCodeInvalid                      ErrorCode = "postal_code_invalid"
+	ErrorCodeProcessingError                        ErrorCode = "processing_error"
+	ErrorCodeProductInactive                        ErrorCode = "product_inactive"
+	ErrorCodeRateLimit                              ErrorCode = "rate_limit"
+	ErrorCodeResourceAlreadyExists                  ErrorCode = "resource_already_exists"
+	ErrorCodeResourceMissing                        ErrorCode = "resource_missing"
+	ErrorCodeRoutingNumberInvalid                   ErrorCode = "routing_number_invalid"
+	ErrorCodeSecretKeyRequired                      ErrorCode = "secret_key_required"
+	ErrorCodeSepaUnsupportedAccount                 ErrorCode = "sepa_unsupported_account"
```
brandur-stripe: Last one Remi: I think we want to go with "SEPA" instead of "Sepa" here.

brandur-stripe:
One more casing issue, but other than that LGTM! Love how comprehensive this list is.

ptal @remi-stripe

error.go
remi-stripe: We use `Sepa` everywhere else in the library though for SepaDebit. I don't think we should change that one but happy to make the change if you prefer on that one

brandur-stripe: Ah, k. I looked around online and I think "SEPA" is definitely the overall preferred casing (and use in Stripe's docs, etc. too).

That said, given we have pre-existing `Sepa` in stripe-go, I'm good just keeping it like this for consistency.

remi-stripe: Yeah, I'm unsure on that one, I think both would make sense so deferring to you but happy to make the change

brandur-stripe: K cool. Easiest to go with "Sepa" for now then, so just leave it. Thanks!

brandur-stripe:
Released as 61.20.0.

layby42:
`invoice_payment_intent_requires_action`  (?)

remi-stripe:
@layby42 Would you have a bit more details than just this code? It's not listed in https://stripe.com/docs/error-codes today which is why it wasn't added. I'd recommend flagging to our support team if it's a code you rely on: https://support.stripe.com/contact

layby42:
@remi-stripe happens when you try to pay invoice (Invoices.Pay) for "incomplete" subscription with card payment method (stripe.InvoicePayParams -> PaymentMethod) that requires 3D Secure authorization on attempt to charge. 

Also (not directly related) would be nice to get Invoice from stripe.Error (the way it is possible to get PaymentIntent or PaymentMethod) to not re-request if from Stripe.

remi-stripe:
> Also (not directly related) would be nice to get Invoice from stripe.Error (the way it is possible to get PaymentIntent or PaymentMethod) to not re-request if from Stripe.
Yeah we've been thinking about this internally too but no short term plan to support this unfortunately.

PR #899, opened 2019-07-19
Add OffSession to SubscriptionItem update

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


subitem.go
```diff
@@ -15,6 +15,7 @@ type SubscriptionItemParams struct {
 	TaxRates          []*string                                `form:"tax_rates"`
 
 	// The following parameter is only supported on updates
```
ob-stripe:
```suggestion
	// The following parameters are only supported on updates
```

remi-stripe: Nice catch, fixed.

Issue #866, opened 2019-05-27
Charges sometimes requested with no POST body


craigandrews:
Go version: 1.12.5
Stripe Go version: v61.0.1

Receive Stripe error code `parameter_missing` with failure `Must provide source or customer.` This happens sporadically, about 0.5% of the time (18 out of 3150 charges this morning). On looking at the Stripe developer logs dashboard I see that the request had no POST body at all.

![Screenshot at 2019-05-27 15-25-04](https://user-images.githubusercontent.com/44933829/58426945-fd51f500-8095-11e9-96bf-9287d0f1b7ba.png)

Retrying the transaction with the exact same code works fine.

Client code is pretty standard:

```go
	// Create the charge in Stripe
	chargeParams := &stripe.ChargeParams{
		Amount:      stripe.Int64(amount),
		Currency:    stripe.String(string(currency)),
		Description: stripe.String(description),
		Customer:    stripe.String(customerToken),
	}
	chargeParams.AddExpand("balance_transaction")

	if idempotencyKey != "" {
		chargeParams.SetIdempotencyKey(idempotencyKey)
	}

	ch, err := charge.New(chargeParams)
```

Can confirm that all values are set properly as they are logged just before creating the charge. And again, retrying seems to work.

brandur-stripe:
Hi @candrewsmtn, sorry to hear about the trouble.

This is definitely a weird one. This probably isn't what you want to hear, but this is most likely a bug in your code somewhere — if you pass a `nil` params into `charge.New`, the underlying `Call` will silently skip over that input and allow the API itself to complain, producing exactly the same error that you're seeing:

```
2019/05/28 15:21:33 Error encountered from Stripe: {"code":"parameter_missing","status":400,"message":"Must provide source or customer.","request_id":"req_Xd6WU9vN1HEwA9","type":"invalid_request_error"}
```

Can you further audit your code to make sure that `chargeParams` couldn't possibly be getting set to `nil` somewhere?

Even if you're sure that it's not possible, I think we're still going to need to come up with a way to reproduce this. I took a look through the code in question and there's nothing obvious to me that would cause it, so we'd need to try and localize the problem.

Do you have any ideas on how we might be able to repro? Have you noticed any patterns around the failed requests? (Do they all happen near the beginning of the app's lifecycle? Are they roughly distributed throughout its lifetime?) Thanks!

layby42:
[...]

PR #903, opened 2019-07-24
Disable HTTP/2 by default

brandur:
See #866, but once in a while when using HTTP/2, stripe-go sends an
empty request body which then results in a 400 back from the server.

We tracked the issue down to this one open on the Go language's HTTP/2
implementation:

https://github.com/golang/go/issues/32441

For now, the best thing we can do is disable HTTP/2 by default for new
integrations. When the bug is fixed, we can turn it back on again.

r? @dcarney-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -3,6 +3,7 @@ package stripe
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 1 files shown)


stripe.go
```diff
@@ -944,7 +945,43 @@ var appInfo *AppInfo
...
+// `HTTPClient` value when using `BackendConfig`.
+var httpClient = &http.Client{
+	Timeout: defaultHTTPTimeout,
+
+	// There is a bug in Go's HTTP/2 implementation that occasionally causes it
+	// to send an empty body when it receives a `GOAWAY` message from a server:
+	//
+	//     https://github.com/golang/go/issues/32441
+	//
+	// This is particularly problematic for this library because the empty body
```
dcarney-stripe: Great comment; thanks for capturing all the context in here.  It's my sincere hope that we can re-enable H/2 support soon, when that Go issue gets addressed.

brandur-stripe:
Thanks Dylan!

stripe.go
brandur-stripe: Thanks! And yep, totally +1.

brandur-stripe: Thanks! And yep, totally +1.

brandur-stripe:
Released as 61.20.0.

PR #904, opened 2019-07-24
Add support for Klarna and SourceOrders

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Putting up for naming review, this was kinda a gnarly one for naming.

I saw another reference of `...ItemsParams` so I used `SourceOrderItemsParams` not sure if that's better as `SourceOrderItemParams` (single item param, technically.)

Also, for the deserialized json object having SourceSourceOrder... everywhere seems a little odd :)

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -359,31 +359,6 @@ type ChargePaymentMethodDetailsIdeal struct {
 // ChargePaymentMethodDetailsKlarna represents details for the Klarna
 // PaymentMethod.
 type ChargePaymentMethodDetailsKlarna struct {
-	BackgroundImageURL              string `json:"background_image_url"`
-	ClientToken                     string `json:"client_token"`
-	FirstName                       string `json:"first_name"`
-	LastName                        string `json:"last_name"`
-	Locale                          string `json:"locale"`
-	LogoURL                         string `json:"logo_url"`
-	PageTitle                       string `json:"page_title"`
-	PayLaterAssetURLsDescriptive    string `json:"pay_later_asset_urls_descriptive"`
-	PayLaterAssetURLsStandard       string `json:"pay_later_asset_urls_standard"`
-	PayLaterName                    string `json:"pay_later_name"`
-	PayLaterRedirectURL             string `json:"pay_later_redirect_url"`
-	PayNowAssetURLsDescriptive      string `json:"pay_now_asset_urls_descriptive"`
-	PayNowAssetURLsStandard         string `json:"pay_now_asset_urls_standard"`
-	PayNowName                      string `json:"pay_now_name"`
-	PayNowRedirectURL               string `json:"pay_now_redirect_url"`
-	PayOverTimeAssetURLsDescriptive string `json:"pay_over_time_asset_urls_descriptive"`
-	PayOverTimeAssetURLsStandard    string `json:"pay_over_time_asset_urls_standard"`
-	PayOverTimeName                 string `json:"pay_over_time_name"`
-	PayOverTimeRedirectURL          string `json:"pay_over_time_redirect_url"`
-	PaymentMethodCategories         string `json:"payment_method_categories"`
-	PurchaseCountry                 string `json:"purchase_country"`
-	PurchaseType                    string `json:"purchase_type"`
-	RedirectURL                     string `json:"redirect_url"`
-	ShippingFirstName               string `json:"shipping_first_name"`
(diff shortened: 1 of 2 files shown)


charge.go
```diff
@@ -355,6 +356,36 @@ type ChargePaymentMethodDetailsIdeal struct {
 	VerifiedName string `json:"verified_name"`
 }
 
+// ChargePaymentMethodDetailsKlarna represents details for the Klarna
+// PaymentMethod.
+type ChargePaymentMethodDetailsKlarna struct {
```
remi-stripe: I think there's a confusion here, there's nothing in PMD on Charge. In stripe-java it's currently empty: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Charge.java#L1570

source.go
```diff
@@ -46,6 +46,17 @@ const (
 	SourceMandateNotificationMethodNone   SourceMandateNotificationMethod = "none"
 )
 
+// SourceSourceOrderItemType describes the type of source order items on source
+// orders for sources.
+type SourceSourceOrderItemType string
+
+// The list of possible values for source order item types.
+const (
+	SourceSourceOrderItemTypeSKU      SourceSourceOrderItemType = "sku"
```
remi-stripe: it looks like you're missing discount. Or the team made discount in the spec but shouldn't have: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/param/SourceCreateParams.java#L1070

source.go
```diff
@@ -222,6 +273,45 @@ type SourceMandate struct {
...
+	Amount      int64                     `json:"amount"`
+	Currency    Currency                  `json:"currency"`
+	Description string                    `json:"description"`
+	Quantity    int64                     `json:"quantity"`
+	Type        SourceSourceOrderItemType `json:"type"`
+}
+
+// SourceSourceOrderShipping describes the shipping options for source orders on
+// sources.
+type SourceSourceOrderShipping struct {
```
remi-stripe: This should re-use the normal address fields that exist elsewhere. That's what we do in Java where it uses `ShippingDetails`: https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Source.java#L877

So here in go you want to use `Shipping` and `ShippingParams` I thiiiiink

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries 

Yep. I was confused by the data on the charge.source.klarna, which you've shared is roughly typed by the type_data.

I must have missed the discount source order item type because it's not documented here: https://stripe.com/docs/api/sources/object#source_object-source_order-items-type

but I _do_ see discount in the spec.

remi-stripe:
This looks good to me. Re-assigning to @brandur-stripe to be safe.

source.go
```diff
@@ -222,6 +253,24 @@ type SourceMandate struct {
...
+	Quantity    int64                     `json:"quantity"`
+	Type        SourceSourceOrderItemType `json:"type"`
+}
+
+// SourceSourceOrder describes a source order for a source.
+type SourceSourceOrder struct {
+	Amount   int64                  `json:"amount"`
+	Currency Currency               `json:"currency"`
+	Email    string                 `json:"email"`
+	Items    SourceSourceOrderItems `json:"items"`
```
brandur-stripe: Mind making this one a pointer to `SourceSourceOrderItems`? It probably doesn't matter that much, but it's basically the overwhelming convention we have.

cjavilla-stripe: Got it. Yeah I'd like to eventually get spun up on releasing. I have a few other projects I'm trying to wrap up before I'll be ready to learn that :)

cjavilla-stripe:
Happy to squash, didn't want to kill the comments again :)

brandur-stripe:
> Happy to squash, didn't want to kill the comments again :)

Yeah, actually would you mind going for the squash? It's not too big of a deal, but I think all regular contributors are in this habit now, and it does help to keep history relatively clean.

> Got it. Yeah I'd like to eventually get spun up on releasing. I have a few other projects I'm trying to wrap up before I'll be ready to learn that :)

Cool! Alright, I'll take this one then.

brandur-stripe:
@cjavilla-stripe Just FYI, I just ended up squashing from the GitHub interface (I don't think it makes a huge amount of difference in practice).

brandur-stripe:
Released as 61.21.0.

PR #905, opened 2019-07-25
Add missing Authenticated field for 3DS charges

remi-stripe:
Add support for the missing `Authenticated` field for 3DS `Charge`.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
Released as 61.20.0.

PR #906, opened 2019-07-26
Add decline code constants

daisy1754:
Adding constants for code in https://stripe.com/docs/declines/codes

### DIFF
--- a/error.go
+++ b/error.go
@@ -19,6 +19,9 @@ const (
 // ErrorCode is the list of allowed values for the error's code.
 type ErrorCode string
 
+// DeclineCode is the list of reasons provided by card issuers for decline of payment.
+type DeclineCode string
+
 // List of values that ErrorCode can take.
 const (
 	ErrorCodeAccountAlreadyExists                   ErrorCode = "account_already_exists"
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -115,6 +117,53 @@ const (
 	ErrorCodeInvalidSwipeData ErrorCode = "invalid_swipe_data"
 )
 
+// List of DeclineCode values.
+const (
+	DeclineCodeApprovalWithID                 DeclineCode = "approve_with_id"
```
remi-stripe: This should be DeclineCodeApproveWithID

error.go
```diff
@@ -115,6 +117,53 @@ const (
 	ErrorCodeInvalidSwipeData ErrorCode = "invalid_swipe_data"
 )
 
+// List of DeclineCode values.
+const (
```
remi-stripe: You're missing `authentication_required`

error.go
```diff
@@ -115,6 +117,53 @@ const (
...
+	DeclineCodeDoNotHonor                     DeclineCode = "do_not_honor"
+	DeclineCodeDoNotTryAgain                  DeclineCode = "do_not_try_again"
+	DeclineCodeDuplicateTransaction           DeclineCode = "duplicate_transaction"
+	DeclineCodeExpiredCard                    DeclineCode = "expired_card"
+	DeclineCodeFraudulent                     DeclineCode = "fraudulent"
+	DeclineCodeGenericDecline                 DeclineCode = "generic_decline"
+	DeclineCodeIncorrectNumber                DeclineCode = "incorrect_number"
+	DeclineCodeIncorrectCVC                   DeclineCode = "incorrect_cvc"
+	DeclineCodeIncorrectPIN                   DeclineCode = "incorrect_pin"
+	DeclineCodeIncorrectZIP                   DeclineCode = "incorrect_zip"
```
remi-stripe: We use `Zip` and not `ZIP` in the library

error.go
```diff
@@ -115,6 +117,53 @@ const (
...
+	DeclineCodeCurrencyNotSupported           DeclineCode = "currency_not_supported"
+	DeclineCodeDoNotHonor                     DeclineCode = "do_not_honor"
+	DeclineCodeDoNotTryAgain                  DeclineCode = "do_not_try_again"
+	DeclineCodeDuplicateTransaction           DeclineCode = "duplicate_transaction"
+	DeclineCodeExpiredCard                    DeclineCode = "expired_card"
+	DeclineCodeFraudulent                     DeclineCode = "fraudulent"
+	DeclineCodeGenericDecline                 DeclineCode = "generic_decline"
+	DeclineCodeIncorrectNumber                DeclineCode = "incorrect_number"
+	DeclineCodeIncorrectCVC                   DeclineCode = "incorrect_cvc"
+	DeclineCodeIncorrectPIN                   DeclineCode = "incorrect_pin"
```
remi-stripe: @brandur-stripe Should we use `PIN` or `Pin`?

error.go
```diff
@@ -115,6 +117,53 @@ const (
...
+	DeclineCodeIncorrectZIP                   DeclineCode = "incorrect_zip"
+	DeclineCodeInsufficientFunds              DeclineCode = "insufficient_funds"
+	DeclineCodeInvalidAccount                 DeclineCode = "invalid_account"
+	DeclineCodeInvalidAmount                  DeclineCode = "invalid_amount"
+	DeclineCodeInvalidCvc                     DeclineCode = "invalid_cvc"
```
remi-stripe: This should be `DeclineCodeInvalidCVC`

remi-stripe:
@daisy1754 Thanks a lot for doing the work on that one! I left some comments but this is already looking great!

remi-stripe:
@daisy1754 Thanks for making all the changes! Can I ask you to squash all your commits into one? That's what we usually do on the library before merging. Once it's done, you can assign to @brandur-stripe for review!

daisy1754:
@remi-stripe  thanks for quick turnaround and detailed review. I squashed all commits into one. It seems I don't have permission to assign PR to another person

error.go
remi-stripe: let's use PIN, it's an acronym

error.go
```diff
@@ -115,6 +118,54 @@ const (
...
+	DeclineCodeInvalidNumber                  DeclineCode = "invalid_number"
+	DeclineCodeInvalidPIN                     DeclineCode = "invalid_pin"
+	DeclineCodeIssuerNotAvailable             DeclineCode = "issuer_not_available"
+	DeclineCodeLostCard                       DeclineCode = "lost_card"
+	DeclineCodeMerchantBlacklist              DeclineCode = "merchant_blacklist"
+	DeclineCodeNewAccountInformationAvailable DeclineCode = "new_account_information_available"
+	DeclineCodeNoActionTaken                  DeclineCode = "no_action_taken"
+	DeclineCodeNotPermitted                   DeclineCode = "not_permitted"
+	DeclineCodePickupCard                     DeclineCode = "pickup_card"
+	DeclineCodePinTryExceeded                 DeclineCode = "pin_try_exceeded"
```
remi-stripe: @daisy1754 Can you rename `Pin` to `PIN` to be consistent?

daisy1754: @remi-stripe  done

remi-stripe:
@daisy1754 Thanks a lot! Re-assigning to @brandur-stripe to do a final review and then a release!

brandur-stripe:
LGTM! Thank you @daisy1754 for the very close attention to detail here. Great stuff :)

@remi-stripe You may have noted elsewhere otherwise, but I think technically this is a breaking change. Now that `DeclineCode` on `CardError` has transition from `string` to the new `DeclineCode` type, users that were putting it in a string field may now have to convert it back to a string to do so:

``` go
// Before
log.DeclineCode = stripeErr.DeclineCode

// After
log.DeclineCode = string(stripeErr.DeclineCode)
```

Any thoughts on what to do about this? We could just release the major. Another alternative would be to not change the type of `CardError.DeclineCode` for the time being (leave it as a `string` for now), and then change it over on the next major as a separate step. It wouldn't be as convenient or type safe, but having access to the new constants would still be useful for people. Thoughts?

remi-stripe:
Ah yes nice catch, we likely should leave as a string for now with a TODO for the next major version (there are a few other things to clean up!)

error.go
```diff
@@ -191,7 +242,7 @@ func (e *PermissionError) Error() string {
 // reason.
 type CardError struct {
 	stripeErr   *Error
-	DeclineCode string `json:"decline_code,omitempty"`
+	DeclineCode DeclineCode `json:"decline_code,omitempty"`
```
brandur-stripe: Would you mind changing this line to something like:

``` go
	// DeclineCode is a code indicating a card issuer's reason for declining a
	// card (if they provided one).
	//
	// TODO: This field should be of type DeclineCode (as defined above), but
	// we've let it as a string for now for backwards compatibility. Change its
	// type on the library's next major version.
	DeclineCode string `json:"decline_code,omitempty"`
}
```

error.go
```diff
@@ -232,5 +283,5 @@ type rawError struct {
 // type to help with deserialization. (e.g. DeclineCode).
 type rawErrorInternal struct {
 	*Error
-	DeclineCode *string `json:"decline_code,omitempty"`
+	DeclineCode *DeclineCode `json:"decline_code,omitempty"`
```
brandur-stripe: And similar to the above, probably this one too.

brandur-stripe:
Cool, thanks Remi!

@daisy1754 As noted above, we're probably going to have to ship this out in two phases just so we can send the first one out as a minor version instead of having to cut a major. I left some suggested wording for a comment to add to the code above. Let me know if that all makes sense.

After that change is finished, I think this is good to go! Thanks.

error.go
daisy1754: hm good catch. what do you think about using type alias (ie., `type DeclineCode = string`)

brandur-stripe: Good question. I'd still say we should just put a `string`: the long term intent is that `DeclineCode` not be a type alias, so I'd say we just make it the right thing now, and then keep it that way.

Admittedly, we have to change something either way on the next major, but omitting the type alias in favor of a simple `string` makes looking at what the field's definition more clear in my mind.

daisy1754:
@brahn-stripe  updated

brandur-stripe:
This looks good to me. Thank you again for the persistent effort on this one @daisy1754!

brandur-stripe:
Released as 61.24.0.

PR #908, opened 2019-07-29
fix spelling in payment intent parameters struct

yep:
change "PayentMethodOptions" to "PaymentMethodOptions"

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -112,16 +112,16 @@ type PaymentIntentConfirmParams struct {
 	Params `form:"*"`
 	// This parameter expects a boolean but used to take an enum so we're adding support for both
 	// until the next major version (TODO).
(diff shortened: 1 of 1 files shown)


remi-stripe:
@yep Nice catch, sorry about that one.

Since the field is deserialized properly, multiple devs likely just used the name as is. Removing it would be a breaking change. Would you mind duplicate the field temporarily in the struct? The idea would be to add the old one (with the incorrect spelling) at the bottom of the struct with a `// TODO: remove` or similar comment.

All this field is used in the params and also in SetupIntent so it would need to be fixed on all those.

Let me know if you'd rather we handle for you!

yep:
Just go ahead and handle it yourself.

remi-stripe:
@yep I just released 61.22.0 with the fix!

PR #909, opened 2019-07-29
Fix typo in property name for PaymentMethodOptions

remi-stripe:
Fix the issue that @yep found in https://github.com/stripe/stripe-go/pull/908 but across both resources and in a non-breaking way.

r? @brandur-stripe 
cc @stripe/api-libraries


brandur-stripe:
Arg! Oops.

LGTM.

PR #910, opened 2019-07-30
Move Balance History to /v1/balance_transactions

remi-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries


balance/client.go
```diff
@@ -33,7 +33,7 @@ func GetBalanceTransaction(id string, params *stripe.BalanceTransactionParams) (
 
 // GetBalanceTransaction retrieves a balance transaction
 func (c Client) GetBalanceTransaction(id string, params *stripe.BalanceTransactionParams) (*stripe.BalanceTransaction, error) {
```
remi-stripe: @brandur-stripe Should we move to a balancetransaction package now that it has its own endpoint? If so would we call it `bt` or `balancetransaction`? And would we do a major for good or keep the current version for a while?

brandur-stripe: Yeah, good question. I'd keep what you have here already, but yeah, ideally it'd get its own package for consistency — balance and balance transaction are related concepts, but it's too weird how they interact like this and really almost nothng else does. We already have `customerbalancetransaction`, so call it `balancetransaction`.

Personally, I'd vote we just keep this one for a while and mark it deprecated (in comment using Go's magic `Deprecated:` string). BTs have been around forever, so I'm sure a lot of people are using this functions and it'd be a pain to require them to move off immediately (especially since it doesn't cost us much to support them).

brandur-stripe:
I love this cleanup. The weird edges in this library are getting increasingly rare ... (Well, my impression anyway.)

ptal @remi-stripe

remi-stripe:
@brandur-stripe okay I added the `balancetransaction` package and marked the old methods as deprecated. PTAL

balance/client.go
remi-stripe: Only saw the mention of `Deprecated:` after pushing so just force pushed again with the `Deprecated:` message!

PR #914, opened 2019-08-12
Update webhook handler example to use `http.MaxBytesReader`

brandur:
Updates the webhook handler example to use `http.MaxBytesReader` to
protect against a malicious client streaming an endless request body.

We're making a similar change in our server side documentation examples,
so I'm updating this spot as well for consistency.

r? @rattrayalex-stripe

### DIFF
--- a/webhook/client_handler_test.go
+++ b/webhook/client_handler_test.go
@@ -11,6 +11,10 @@ import (
 
 func Example() {
 	http.HandleFunc("/webhook", func(w http.ResponseWriter, req *http.Request) {
+		// Protects against a malicious client streaming us an endless requst
+		// body
+		const MaxBodyBytes = int64(65536)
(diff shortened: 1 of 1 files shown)


brandur-stripe:
> Still not the biggest fan (verbose, distracting, low likelihood of causing problems, 64k could possibly be too low and be a footgun) but +1 on consistency and I defer to you on all of these questions anyway.

Yeah, definitely fair enough. I thought about adding another `if` conditional as well where we check whether the number of bytes read equals max size and throws a specialized error in that case, because the trouble is that if it was, then the current error will be some JSON decoding thing like `missing }` or something terribly confusing like that. But of course that adds even more code to the sample which makes it even harder to parse.

Anyway, going to merge this for now to keep it consistent with the other docs, but definitely up for revisiting if we find it was the wrong decision longer term.

Thanks!

PR #915, opened 2019-08-14
Move to API version 2019-08-14 and other changes

remi-stripe:
* Pin to API version 2019-08-14
* Rename AccountCapabilityPlatformPayments to AccountCapabilityTransfers
* Add `Executive` in `PersonRelationship`
* Remove `PayentMethodOptions` as there was a typo which was fixed.
* Make `OffSession` only support booleans on `PaymentIntent`
* Remove `PaymentIntentLastPaymentError` and use `Error` instead
* Move `DeclineCode` on `Error` to the `DeclineCode` instead of `string`

r? @ob-stripe @brandur-stripe (tagging you both are there are multiple changes
cc @stripe/api-libraries

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -110,10 +110,8 @@ type PaymentIntentCaptureParams struct {
 
 // PaymentIntentConfirmParams is the set of parameters that can be used when confirming a payment intent.
 type PaymentIntentConfirmParams struct {
-	Params `form:"*"`
-	// This parameter expects a boolean but used to take an enum so we're adding support for both
-	// until the next major version (TODO).
-	OffSession           interface{}                              `form:"off_session"`
+	Params               `form:"*"`
+	OffSession           *bool                                    `form:"off_session"`
 	PaymentMethod        *string                                  `form:"payment_method"`
 	PaymentMethodOptions *PaymentIntentPaymentMethodOptionsParams `form:"payment_method_options"`
 	PaymentMethodTypes   []*string                                `form:"payment_method_types"`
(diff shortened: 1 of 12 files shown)


person.go
```diff
@@ -126,6 +126,7 @@ type DOB struct {
 type Relationship struct {
 	AccountOpener    bool    `json:"account_opener"`
 	Director         bool    `json:"director"`
+	Director         bool    `json:"executive"`
```
ob-stripe:
```suggestion
	Executive        bool    `json:"executive"`
```

remi-stripe: Fixed, PTAL and sorry for the churn

ob-stripe:
LGTM! Love the various cleanups here.

Reassigning to @brandur-stripe for a second look just in case.

paymentintent.go
```diff
@@ -242,7 +219,7 @@ type PaymentIntent struct {
 	Customer                  *Customer                       `json:"customer"`
 	Description               string                          `json:"description"`
 	Invoice                   *Invoice                        `json:"invoice"`
-	LastPaymentError          *PaymentIntentLastPaymentError  `json:"last_payment_error"`
+	LastPaymentError          *Error                          `json:"last_payment_error"`
```
brandur-stripe: Just flagging that it looks like `DocURL` will go away with this change. I think it may just be an oversight that it's not on the standard `Error` struct?

remi-stripe: Damn nice catch, yes looks like an oversight, let me add this

remi-stripe: Fixed! 🦅 👁

ob-stripe:
Released as 62.0.0.

PR #927, opened 2019-08-22
Put `stripe_test.go` into main `stripe` package

brandur:
Moves `stripe_test.go` out of the `stripe_test` package and into
`stripe`. This makes it consistent with all other `*_test.go` files in
the top-level package. The upside of the change is that it will allow
us to test certain unexported functions without having to export them.
The downside is that the hygiene isn't quite as good because we're no
longer just testing against the package's public interface which may
reduce flexibility in future changes.

Go standard library convention seems to use both methods. Just looking
through the `net/http` source code for example, you can find tests in
the `http` package _and_ in the `http_test` package depending on the
file.

I've been wanting to making this change for a while because it'd give us
a route to making internal-ish structures like `BackendImplementation`
unexported, but the direct impetus for today is that I was making some
changes to the `shouldRetry` function and noticed that it was difficult
to test them without having to go through the expensive song and dance
of spinning up a test HTTP server and making the test really
complicated. (I could also just export the function as `ShouldRetry`,
but I'd prefer not to do that if possible.)

r? @ob-stripe @remi-stripe Any thoughts/opinions on this change? I
personally like the idea of only testing public interfaces, but it can
be impractical in some cases.

### DIFF
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -1,4 +1,4 @@
-package stripe_test
+package stripe
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks @ob-stripe!

> Appreciate the explanation! Maybe we should have both "internal" and "external" tests like the `http` package.

+1. I'll see if I can try a clean up pass on `stripe.go`/`stripe_test.go` — they could use it, and we could probably break the tests pretty cleanly into ones that test the package interface versus ones that are more internal.

PR #928, opened 2019-08-22
Bring retry code in-line with current best practices

brandur:
Brings stripe-go's retry code in-line with what we currently understand
to be the best practices as already implemented in stripe-node and
stripe-ruby. Basically has the effect of adding some 500s and 503s to
results that will be retried.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -386,7 +386,7 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 
 		// If the response was okay, we're done, and it's safe to break out of
 		// the retry loop.
-		if !s.shouldRetry(err, res, retry) {
+		if !s.shouldRetry(err, req, res, retry) {
 			break
 		}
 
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -571,18 +571,35 @@ func (s *BackendImplementation) UnmarshalJSONVerbose(statusCode int, body []byte
...
 // socket errors that may represent an intermittent problem and some special
 // HTTP statuses.
-func (s *BackendImplementation) shouldRetry(err error, resp *http.Response, numRetries int) bool {
+func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *http.Response, numRetries int) bool {
 	if numRetries >= s.MaxNetworkRetries {
 		return false
 	}
 
+	// TODO: There are many errors that should not be retried. Try to make this
+	// more granular by including only connection errors, timeout errors, etc.
```
brandur-stripe: Added this TODO here because honestly what we're doing right now isn't great.

Unfortunately, routing around Go source looking for errors that we can compare against is slightly non-trivial, so deferring this to a separate improvement project.

brandur-stripe:
Thanks!

brandur-stripe:
Released as 62.3.0.

PR #938, opened 2019-08-29
Rearrange error logging so that 402 doesn't log an error

### LINKED ISSUE #936: #935 (merged yesterday) logs a error level for every 402
Using stripe-go 62.7.0
Golang 1.11.x and 1.12.x

This commit https://github.com/stripe/stripe-go/commit/53c32f1222679822a5b9ebb90c7c6abdaa09ad97 moves `s.LeveledLogger.Errorf("Request failed with body: %s (status: %v)" ...` before the `!s.shouldRetry(err, req, res, retry)` check.

Previously `shouldRetry()` returned `false` for 402 responses causing the logging to be skipped.  Now 402s are always logged as Errors.  I presume this change is unintentional given that 402 does not represent an unexpected error?

brandur:
While looking in #936 I accidentally realized that in #935 I'd
inadvertently duplicated error logging -- I add a log statement in `Do`
for API errors, but `ResponseToError` was already logging them.

Here I've resolved #936 and re-arranged logging somewhat so that we're
more consistent _where_ the logging happens. Most logs are now produced
directly in `Do` so it's easier to follow what's getting logged.

I haven't actually changed our previous policy around 402s -- just moved
it up a function. The previous idea (which I'd forgotten about) was that
because 402s are often induced by the _users of integrations_ instead of
integrations themselves, we just do an `Infof` log for it. Other errors
get an `Errorf` because they're more likely to be the integration's
fault, and actionable.

Fixes #936.

r? @ob-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -393,8 +393,29 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			s.LeveledLogger.Errorf("Request failed with error: %v", err)
 		} else if res.StatusCode >= 400 {
 			err = s.ResponseToError(res, resBody)
-			s.LeveledLogger.Errorf("Request failed with body: %s (status: %v)",
-				string(resBody), res.StatusCode)
+
+			if stripeErr, ok := err.(*Error); ok {
+				// The Stripe API makes a distinction between errors that were
+				// caused by invalid parameters or something else versus those
+				// that occurred *despite* valid parameters, the latter coming
+				// back with status 402.
+				//
+				// On a 402, log to info so as to not make an integration's log
+				// noisy with error messages that they don't have much control
+				// over.
+				//
+				// Note I use the constant 402 instead of an `http.Status*`
+				// constant because technically 402 is "Payment required". The
+				// Stripe API doesn't comply to the letter of the specification
+				// and uses it in a broader sense.
+				if res.StatusCode == 402 {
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Thanks OB!

brandur-stripe:
Released as 62.8.1.

stripe.go
```diff
@@ -393,8 +393,29 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v inte
 			s.LeveledLogger.Errorf("Request failed with error: %v", err)
 		} else if res.StatusCode >= 400 {
 			err = s.ResponseToError(res, resBody)
-			s.LeveledLogger.Errorf("Request failed with body: %s (status: %v)",
-				string(resBody), res.StatusCode)
+
+			if stripeErr, ok := err.(*Error); ok {
```
yareid: @brandur-stripe -- `err` is now not logged if the type assertion fails.  Not a problem if err always has type `*Error`, but might be a problem otherwise.

brandur-stripe: Thanks. To be fair, this should basically ~never happen (only in the case where we got back some mangled error body from the Stripe API, but where it somehow managed to close the connection properly, etc.), but yeah, we should make sure that it gets logged.

Created a small tweak on the last patch in #939.

PR #939, opened 2019-08-29
Also log error in case of non-`stripe.Error`

brandur:
In a few cases it's possible for `ResponseToError` to return an error
that's not a `*stripe.Error`. This should never happen, but could occur
if say the Stripe API returned a mangled JSON blob, but somehow managed
to close the connection properly. It's an unlikely scenario, but even
so, it's better not to let it slip through the cracks.

I also moved one more logging statement out of `ResponseToError` so we
can log everything of this type in one place.

r? @ob-stripe (Sorry, one more tweak on the last patch.)
cc @stripe/api-libraries


brandur-stripe:
Thank you again :)

brandur-stripe:
Released as 62.8.2.

PR #950, opened 2019-09-16
Remove lossy "MarshalJSON" implementations

### LINKED ISSUE #946: PayoutDestination Marshaling only returns ID
**go1.12.6 darwin/amd64
stripe-go v62.10.0**

When retrieving a PayoutList with expanded Destination ('data.destination') and returning as JSON in our orchestrator, destination is coded to only marshal the ID regardless of expansion.

**Line 190+ in payout.go**
```
func (d *PayoutDestination) MarshalJSON() ([]byte, error) {
      return json.Marshal(d.ID)
}
```

Expected it to return data much like the CURL does, with properly expanded detail.

brandur:
See #946 for context, but we're removing some `MarshalJSON`
implementations that are "lossy" in that they throw out information that
might be useful to somebody.

Note that we don't actually marshal API resources as part of normal
operations in stripe-go, so the only users that this will affect are
those that are manually re-marshaling resources to JSON in their
integrations.

Fixes #946.

r? @remi-stripe
cc @stripe/api-libraries

---

Targets the V64 integration branch in # instead of `master`.

### DIFF
--- a/transfer_test.go
+++ b/transfer_test.go
@@ -39,14 +39,11 @@ func TestTransferDestination_UnmarshalJSON(t *testing.T) {
 
 	// Unmarshals from a JSON object
 	{
-		// We build the JSON object manually here because TransferDestination
-		// has a custom MarshalJSON implementation as well, and it'll turn into
-		// a string if we marshaled a struct instance. This ensures that we're
-		// working with a JSON objects.
-		data := []byte(`{"id":"acct_123"}`)
+		v := TransferDestination{ID: "acct_123"}
+		data, err := json.Marshal(&v)
+		assert.NoError(t, err)
 
-		var v TransferDestination
-		err := json.Unmarshal(data, &v)
+		err = json.Unmarshal(data, &v)
 		assert.NoError(t, err)
 		assert.Equal(t, "acct_123", v.ID)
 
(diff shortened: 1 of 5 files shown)


transfer_test.go
```diff
@@ -39,14 +39,11 @@ func TestTransferDestination_UnmarshalJSON(t *testing.T) {
 
 	// Unmarshals from a JSON object
 	{
-		// We build the JSON object manually here because TransferDestination
-		// has a custom MarshalJSON implementation as well, and it'll turn into
-		// a string if we marshaled a struct instance. This ensures that we're
-		// working with a JSON objects.
-		data := []byte(`{"id":"acct_123"}`)
+		v := TransferDestination{ID: "acct_123"}
```
brandur-stripe: This change makes it more similar to some of the other test cases in the same file, and manually crafting JSON by hand is usually not that good of an idea anyway.

brandur-stripe:
Thanks Remi.

> Can I ask you to do a branch for this so that we can merge other breaking changes in it to prepare the next major? Similarly to what @ob-stripe does in .net: stripe/stripe-dotnet#1694

Yeah, sure. I'll open a new PR in just a sec.

PR #966, opened 2019-10-02
Add the Account field to BankAccount

pauldemarco:
This change adds the already present (network-wise) Account field to the BankAccount struct, and allows the BankAccount to become self contained --no longer needing to tuple an Event struct alongside the BankAccount when handling webhooks.

Tested working for `sc.BankAccounts.New` and when responding to a `account.external_account.updated` event in a webhook.

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -141,6 +141,7 @@ func (p *BankAccountListParams) AppendTo(body *form.Values, keyParts []string) {
 
 // BankAccount represents a Stripe bank account.
 type BankAccount struct {
+	Account            string                       `json:"account"`
 	AccountHolderName  string                       `json:"account_holder_name"`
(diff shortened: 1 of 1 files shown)


bankaccount.go
```diff
@@ -141,6 +141,7 @@ func (p *BankAccountListParams) AppendTo(body *form.Values, keyParts []string) {
 
 // BankAccount represents a Stripe bank account.
 type BankAccount struct {
+	Account            string                       `json:"account"`
```
remi-stripe: This should be `Account *Account `json:"account"` instead. This property is expandable in the API and so we need to build support for the expanded version even though it can not be expanded in your case because it comes from an event

pauldemarco: All set! (and tested working for me)

PR #967, opened 2019-10-08
orderreturn.Get

richardm-stripe:
This function was referenced in the apiref but lacked an implementation.
Could use some direction, on how to make consistently follow our standards.

I had to deviate from the naming convention for `OrderReturnResourceParams`, because `OrderReturnParams` is already occupied by the custom action on `order` called `return`.

r? @ob-stripe

### DIFF
--- a/order.go
+++ b/order.go
@@ -91,12 +91,6 @@ type OrderUpdateShippingParams struct {
 	TrackingNumber *string `form:"tracking_number"`
 }
 
-// OrderReturnParams is the set of parameters that can be used when returning orders.
-type OrderReturnParams struct {
-	Params `form:"*"`
-	Items  []*OrderItemParams `form:"items"`
-}
-
 // Shipping describes the shipping hash on an order.
 type Shipping struct {
 	Address        *Address `json:"address"`
--- a/orderreturn.go
+++ b/orderreturn.go
@@ -2,10 +2,11 @@ package stripe
 
 import "encoding/json"
 
-// OrderReturnResourceParams is the set of parameters that can be used when retrieving a transaction.
-// It is not to be confused with OrderReturnParams, which is used to return an order.
-type OrderReturnResourceParams struct {
-  Params   `form:"*"`
+// OrderReturnParams is the set of parameters that can be used when returning orders.
+type OrderReturnParams struct {
+	Params `form:"*"`
+	Items  []*OrderItemParams `form:"items"`
+	Order  *string            `form:"-"` // Included in the URL
 }
 
 // OrderReturn is the resource representing an order return.
--- a/orderreturn.go
+++ b/orderreturn.go
@@ -2,6 +2,12 @@ package stripe
 
 import "encoding/json"
 
+// OrderReturnResourceParams is the set of parameters that can be used when retrieving a transaction.
+// It is not to be confused with OrderReturnParams, which is used to return an order.
+type OrderReturnResourceParams struct {
+  Params   `form:"*"`
+}
+
 // OrderReturn is the resource representing an order return.
 // For more details see https://stripe.com/docs/api#order_returns.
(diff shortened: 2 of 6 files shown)


ob-stripe:
Looks generally sane to me, but I'm not terribly familiar with this library so reassigning. Also, it looks like you have a linting error.

brandur-stripe:
I think it'd be more conventional for this one to not introduce a new params class, and instead move the existing `OrderReturnParams` over to `orderreturn.go` and reuse it — I'm not sure there's a good reason for the current state of things here, but if you look at other resources the `New` and `Get` functions share a parameters. Order return's `New` is instead called `order.Return`, but it's the same idea.

While we're at it, we might also want to consider "normalizing" order return by also giving it a `New` method. It'd be the same as `order.Return`, but it'd have the `Order` ID as a member of the `OrderReturnParams` like so:

``` go
type OrderReturnParams struct {
	Params `form:"*"`
	Order  *string            `form:"-"` // Included in URL
...
```

(See `person.New` or `taxid.New` as relatively modern examples of how a "nested resource"'s new function looks.)

@remi-stripe Do you agree?

orderreturn/client.go
```diff
@@ -13,6 +13,18 @@ type Client struct {
...
 }
 
+// Get returns the details of an order return
+func Get(id string, params *stripe.OrderReturnResourceParams) (*stripe.OrderReturn, error) {
+  return getC().Get(id, params)
+}
+
+func (c Client) Get(id string, params *stripe.OrderReturnResourceParams) (*stripe.OrderReturn, error) {
+  path := stripe.FormatURLPath("/v1/order_returns/%s", id)
+  orderreturn := &stripe.OrderReturn{}
```
brandur-stripe: Package names in Go are all lower case like `orderreturn`, but camel case for variable names please :)

orderreturn/client.go
```diff
@@ -13,6 +13,18 @@ type Client struct {
 	Key string
 }
 
+// Get returns the details of an order return
+func Get(id string, params *stripe.OrderReturnResourceParams) (*stripe.OrderReturn, error) {
+  return getC().Get(id, params)
+}
+
+func (c Client) Get(id string, params *stripe.OrderReturnResourceParams) (*stripe.OrderReturn, error) {
```
brandur-stripe: You'll need a comment on this function to pass the lint check.

orderreturn.go
```diff
@@ -2,6 +2,12 @@ package stripe
 
 import "encoding/json"
 
+// OrderReturnResourceParams is the set of parameters that can be used when retrieving a transaction.
+// It is not to be confused with OrderReturnParams, which is used to return an order.
```
remi-stripe: You should use `OrderReturnParams` instead of making a new one, like we do in other clients I think

richardm-stripe:
like this?

order.go
```diff
@@ -91,12 +91,6 @@ type OrderUpdateShippingParams struct {
 	TrackingNumber *string `form:"tracking_number"`
 }
 
-// OrderReturnParams is the set of parameters that can be used when returning orders.
-type OrderReturnParams struct {
-	Params `form:"*"`
-	Items  []*OrderItemParams `form:"items"`
-}
-
```
remi-stripe: Why was this removed?

richardm-stripe: It's in the same package, different file FWIW

remi-stripe:
I don't fully understand why we made the change since order.Return worked fine and our other libraries don't have such a method either

order.go
brandur-stripe: It's fine to move it to a different file in the same package (the user can't even tell).

richardm-stripe:
Ok ready again
r? @brandur-stripe

PR #970, opened 2019-10-11
JSON tag spelling mistake

chebyrash:
The field Fulfilled had a misspelled JSON tag


remi-stripe:
@chebyrash Released as 65.1.1 thanks again for the fix!

PR #981, opened 2019-10-30
Add support for `InvoiceCustomerBalanceSettings` on `Subscription`

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries


remi-stripe:
We've decided not to ship this. Those parameters will disappear in the next API version and it doesn't make sense to have them in static client libraries.

PR #992, opened 2019-11-20
Add GoPkg.toml manifest declaring CI dependencies

richardm-stripe:
r? @rattrayalex-stripe 
cc @ob-stripe @brandur-stripe 

This is the result of
1. dep init
2. adding `required = ["github.com/jstemmer/go-junit-report"]` to Gopkg.toml.
3. dep ensure

Why? This is the most straightforward (and precedented) way of ensuring access to `go-junit-report` when running the stripe-go tests in pay-server CI.

I don't *think* committing the manifest should have an effect on users of the library.


rattrayalex-stripe:
I'm not familiar enough with these things to say
r? @brandur-stripe || @ob-stripe

ob-stripe:
I'm not familiar with `dep` at all (I've only used the newer Go 1.11 modules for dependency management) so deferring to Brandur here.

richardm-stripe:
Actually I'm not sure this will be sufficient to accomplish what I want. Going to punt on this for now.

PR #1001, opened 2019-12-04
Add support for `Network` on `Charge`

remi-stripe:
Similar to the java one but not touching `SourceTransaction` as stripe-go doesn't support the hash specific for now and since this is low usage I want to avoid adding features that will go away in the future.

r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/paymentmethod.go
+++ b/paymentmethod.go
@@ -39,6 +39,23 @@ const (
 	PaymentMethodCardBrandVisa       PaymentMethodCardBrand = "visa"
 )
 
+// PaymentMethodCardNetwork is the list of allowed values to represent the network
+// used for a card-like transaction.
+type PaymentMethodCardNetwork string
+
+// List of values that PaymentMethodCardNetwork can take.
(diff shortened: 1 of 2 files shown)


paymentmethod.go
```diff
@@ -39,6 +39,23 @@ const (
 	PaymentMethodCardBrandVisa       PaymentMethodCardBrand = "visa"
 )
 
+// PaymentMethodCardNetwork is the list of allowed values to represent the network
+// used for a card-like transaction.
+type PaymentMethodCardNetwork string
```
remi-stripe: I made a call here on where to set the enum. I am not sure this is the right decision in the library. In a "normal" world we'd name if `ChargePaymentMethodDetailsCardNetwork` and it'd be specific, and duplicated for CardPresent, but in the future the network will be chosen for a PI and that will make us rename it.

Long story short: we need to figure out clear rules on how to name enums.

PR #1008, opened 2020-01-06
Add ReportingCategory to BalanceTransaction

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries


cjavilla-stripe:
updated with these reporting categories: https://stripe.com/docs/reports/reporting-categories

PR #1012, opened 2020-01-10
Adds PendingUpdate to Subscription

cjavilla-stripe:
Think I should re-use the existing SubscriptionItem for subscription_items like this? or create a whole new `SubscriptionPendingUpdateSubscriptionItem` etc.

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/invoice.go
+++ b/invoice.go
@@ -119,6 +119,7 @@ type InvoiceParams struct {
 	SubscriptionItems                       []*SubscriptionItemsParams          `form:"subscription_items"`
 	SubscriptionPlan                        *string                             `form:"subscription_plan"`
 	SubscriptionProrate                     *bool                               `form:"subscription_prorate"`
+	SubscriptionProrationBehavior           *string                             `form:"subscription_proration_behavior"`
 	SubscriptionProrationDate               *int64                              `form:"subscription_proration_date"`
 	SubscriptionQuantity                    *int64                              `form:"subscription_quantity"`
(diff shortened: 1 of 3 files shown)


invoice.go
```diff
@@ -119,6 +119,7 @@ type InvoiceParams struct {
 	SubscriptionItems                       []*SubscriptionItemsParams          `form:"subscription_items"`
 	SubscriptionPlan                        *string                             `form:"subscription_plan"`
 	SubscriptionProrate                     *bool                               `form:"subscription_prorate"`
+	SubscriptionProrationBehavior           *string                             `form:"subscription_proration_behavior"`
```
remi-stripe: You're missing a list of constants for all those values. In go, we define them even if it's just used as a parameter and we pass `*string`

cjavilla-stripe: Thanks, I couldn't remember that rule.

cjavilla-stripe:
Added `SubscriptionProrationBehavior` to sub.go

ptal @remi-stripe

cjavilla-stripe:
Hmm. Maybe I'm missing something. I thought the upcoming invoice params were controlled by InvoiceParams? Did I miss it somewhere else?

remi-stripe:
> Hmm. Maybe I'm missing something. I thought the upcoming invoice params were controlled by InvoiceParams? Did I miss it somewhere else?

Ah my bad I always forget that we mix them. All good then. Merging

PR #1013, opened 2020-01-12
Add support for `CreditNoteLineItem`

remi-stripe:
r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/creditnote.go
+++ b/creditnote.go
@@ -31,18 +31,28 @@ const (
 	CreditNoteTypePrePayment  CreditNoteType = "pre_payment"
 )
 
+// CreditNoteLineItemType is the list of allowed values for the credit note line item's type.
+type CreditNoteLineItemType string
+
+// List of values that CreditNoteType can take.
+const (
+	CreditNoteLineItemTypeCustomLineItem  CreditNoteLineItemType = "custom_line_item"
+	CreditNoteLineItemTypeInvoiceLineItem CreditNoteLineItemType = "invoice_line_item"
+)
+
 // CreditNoteParams is the set of parameters that can be used when creating or updating a credit note.
 // For more details see https://stripe.com/docs/api/credit_notes/create, https://stripe.com/docs/api/credit_notes/update.
 type CreditNoteParams struct {
 	Params          `form:"*"`
-	Amount          *int64  `form:"amount"`
-	CreditAmount    *int64  `form:"credit_amount"`
-	Invoice         *string `form:"invoice"`
-	Memo            *string `form:"memo"`
-	OutOfBandAmount *int64  `form:"out_of_band_amount"`
-	Reason          *string `form:"reason"`
-	Refund          *string `form:"refund"`
-	RefundAmount    *int64  `form:"refund_amount"`
+	Amount          *int64                  `form:"amount"`
+	CreditAmount    *int64                  `form:"credit_amount"`
+	Invoice         *string                 `form:"invoice"`
+	Lines           []*CreditNoteLineParams `form:"lines"`
+	Memo            *string                 `form:"memo"`
+	OutOfBandAmount *int64                  `form:"out_of_band_amount"`
+	Reason          *string                 `form:"reason"`
+	Refund          *string                 `form:"refund"`
+	RefundAmount    *int64                  `form:"refund_amount"`
 }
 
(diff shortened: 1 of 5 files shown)


creditnote.go
```diff
@@ -31,6 +31,15 @@ const (
...
+
+// List of values that CreditNoteType can take.
+const (
+	CreditNoteLineItemTypeCustomLineItem  CreditNoteLineItemType = "custom_line_item"
+	CreditNoteLineItemTypeInvoiceLineItem CreditNoteLineItemType = "invoice_line_item"
+)
+
 // CreditNoteParams is the set of parameters that can be used when creating or updating a credit note.
 // For more details see https://stripe.com/docs/api/credit_notes/create, https://stripe.com/docs/api/credit_notes/update.
 type CreditNoteParams struct {
```
cjavilla-stripe: Should this one have `Lines`?  seems undocumented, but also the api-ref shows Lines in the CreditNote create call for dotnet

creditnote.go
```diff
@@ -53,6 +62,27 @@ type CreditNoteListParams struct {
 	Invoice    *string `form:"invoice"`
 }
 
+// CreditNoteLineItemListParams is the set of parameters that can be used when listing credit note line items.
+type CreditNoteLineItemListParams struct {
+	ListParams `form:"*"`
+
+	// ID is the credit note ID to list line items for.
+	ID *string `form:"-"` // Goes in the URL
```
cjavilla-stripe: Hmm. I would've expected this to be called `CreditNote` ? 

Or like CreditNoteID or something?

creditnote.go
```diff
@@ -53,6 +62,27 @@ type CreditNoteListParams struct {
...
+}
+
+// CreditNoteLineItemListPreviewParams is the set of parameters that can be used when previewing a credit note's line items
+type CreditNoteLineItemListPreviewParams struct {
+	ListParams      `form:"*"`
+	Amount          *int64  `form:"amount"`
+	CreditAmount    *int64  `form:"credit_amount"`
+	Invoice         *string `form:"invoice"`
+	Memo            *string `form:"memo"`
```
cjavilla-stripe: Should this one have `Lines` too?

creditnote.go
remi-stripe: Ack nice catch, yes it should, fixing.

creditnote.go
```diff
@@ -80,27 +117,57 @@ type CreditNote struct {
...
+	ID                string                 `json:"id"`
+	InvoiceLineItem   string                 `json:"invoice_line_item"`
+	Livemode          bool                   `json:"livemode"`
+	Object            string                 `json:"object"`
+	Quantity          int64                  `json:"quantity"`
+	TaxAmounts        []*CreditNoteTaxAmount `json:"tax_amounts"`
+	TaxRates          []*TaxRate             `json:"tax_rates"`
+	Type              CreditNoteLineItemType `json:"type"`
+	UnitAmount        int64                  `json:"unit_amount"`
+	UnitAmountDecimal float64                `json:"unit_amount_decimal,string"`
```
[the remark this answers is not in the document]
remi-stripe: Yes, that's mostly cargo culted from other resources

creditnote.go
```diff
@@ -53,6 +62,27 @@ type CreditNoteListParams struct {
...
+type CreditNoteLineItemListParams struct {
+	ListParams `form:"*"`
+
+	// ID is the credit note ID to list line items for.
+	ID *string `form:"-"` // Goes in the URL
+}
+
+// CreditNoteLineItemListPreviewParams is the set of parameters that can be used when previewing a credit note's line items
+type CreditNoteLineItemListPreviewParams struct {
+	ListParams      `form:"*"`
```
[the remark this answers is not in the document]
remi-stripe: It should, but that's not a pattern the library supports today. `Metadata` in go is handled via a magic nested `Params` struct and that doesn't work with list. Something we'll need to fix in the future.

creditnote.go
remi-stripe: Yeah the name is a bit random, it depends what the team puts in their url path. In a way it doesn't fully matter. I'm happy to go with `CreditNote` instead but usually we put `ID` unless a resource has multiple parent types (rare).

cjavilla-stripe: ack

PR #1025, opened 2020-02-04
[wip] Support source_types from /v1/balances and add fpx source_type string constants

sherbondy-stripe:
Hi folks,

Go novice here, but a merchant recently requested adding support to the Go language bindings for the `source_types` hash on the `pending`/`available` top-level balance object:

https://stripe.com/docs/api/balance/balance_object#balance_object-available-source_types

This is necessary for any markets that involve split balances for compliance/regulatory purposes. This also applies to payment methods like ACH debits in the US (where `source_type=bank_account`).

I also introduce support for the recently-added `fpx` `source_type`  ([currently undocumented as it is in private beta, pending broader FPX rollout](https://support.stripe.com/questions/eligibility-for-fpx-beta)). Could split that out into a separate PR.

One outstanding question I have is whether these `type` declarations are actually tenable in the long term. E.g. if Stripe introduces a new `source_type`, will the Go bindings fail to parse it because  it is not included in the `const` declaration listing? If so, then maybe using raw strings is actually preferable here?

Also assuming I should add tests to help validate?

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -1,5 +1,17 @@
 package stripe
 
+// BalanceSourceType is the list of allowed values for the balance amount's source_type field keys.
+type BalanceSourceType string
+
+// List of values that BalanceSourceType can take.
+const (
+	BalanceSourceTypeAlipayAccount   BalanceSourceType = "alipay_account"
+	BalanceSourceTypeBankAccount     BalanceSourceType = "bank_account"
+	BalanceSourceTypeBitcoinReceiver BalanceSourceType = "bitcoin_receiver"
+	BalanceSourceTypeCard            BalanceSourceType = "card"
+	BalanceSourceTypeFPX             BalanceSourceType = "fpx"
+)
+
 // BalanceTransactionStatus is the list of allowed values for the balance transaction's status.
 type BalanceTransactionStatus string
 
@@ -19,6 +31,7 @@ type Balance struct {
 
 // Amount is a structure wrapping an amount value and its currency.
 type Amount struct {
-	Value    int64    `json:"amount"`
-	Currency Currency `json:"currency"`
+	Value       int64                       `json:"amount"`
+	Currency    Currency                    `json:"currency"`
+	SourceTypes map[BalanceSourceType]int64 `json:"source_types"`
 }
--- a/balance/client.go
+++ b/balance/client.go
@@ -4,7 +4,7 @@ package balance
 import (
 	"net/http"
 
-	stripe "github.com/stripe/stripe-go"
+	stripe "github.com/sherbondy-stripe/stripe-go"
 )
 
 // Client is used to invoke /balance and transaction-related APIs.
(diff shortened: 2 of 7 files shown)


balance.go
```diff
@@ -19,6 +31,7 @@ type Balance struct {
 
 // Amount is a structure wrapping an amount value and its currency.
 type Amount struct {
-	Value    int64    `json:"amount"`
-	Currency Currency `json:"currency"`
+	Value       int64                       `json:"amount"`
+	Currency    Currency                    `json:"currency"`
+	SourceTypes map[BalanceSourceType]int64 `json:"source_types"`
```
remi-stripe: @brandur-stripe What do you think of doing this versus using a sub-class to represent the mapping?

In Java, we have a class named `Money` for the `Amount` class ([here](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Balance.java#L104)) and that has a sub-class named `SourceType`([here](https://github.com/stripe/stripe-java/blob/master/src/main/java/com/stripe/model/Balance.java#L122) (which really should be named `SourceType` instead.).

In dotnet we have `BalanceAmount` ([here](https://github.com/stripe/stripe-dotnet/blob/master/src/Stripe.net/Entities/Balance/BalanceAmount.cs)) (which is what `Amount` in Go should be named really if we were consistent). And that class doesn't have a sub-class for `SourceTypes` and instead uses a `Dictionary` like is introduced in this PR

brandur-stripe: @remi-stripe I'd personally say that this approach is okay.

I don't like Java's because it makes accessing information less convenient — you have to search the list instead of just being able to index on a map.

I could see an argument for maybe making the `int64` another substruct in case it ever becomes a subobject in the API, but IMO it'd be okay to leave it like this and just change it later if necessary.

brandur-stripe:
@sherbondy-stripe Thanks! Just to answer your question:

> One outstanding question I have is whether these `type` declarations are actually tenable in the long term. E.g. if Stripe introduces a new `source_type`, will the Go bindings fail to parse it because  it is not included in the `const` declaration listing? If so, then maybe using raw strings is actually preferable here?

Everything would still work because `BalanceSourceType` is actually just a string, as you've declared here:

```
type BalanceSourceType string
```

If a new `const` isn't introduced in time, it will be less convenient for users, but they can still fall back to using a string with type conversation like:

``` go
amount.SourceTypes[BalanceSourceType("pfx")]
```

remi-stripe:
@sherbondy-stripe Thanks again for doing the work here. I worked on your PR locally to add tests (which all pass yay!) and also added `connect_reserved` which also wasn't supported. You can view my changes in this commit: https://github.com/stripe/stripe-go/commit/e5fab56069771afbadb8e0b83a51d76b65bc8f46

Could you integrate this locally in your branch and update the PR? That way we can review and release support for this! Also you now have write access on this repo so you can push your branch directly here instead of on your fork!

sherbondy-stripe:
Closing in favor of https://github.com/stripe/stripe-go/pull/1026 per Remi's guidance!

PR #1026, opened 2020-02-05
Support `source_types` and `connect_reserved` from `/v1/balances` and add `fpx` source_type string constants

sherbondy-stripe:
Hi folks,

Closing https://github.com/stripe/stripe-go/pull/1025 in favor of this one per Remi's latest comment.

Have incorporated Remi's new test file (💯) as well as the `connect_reserved` addition, consolidated here into a single PR off this branch rather than a fork.

---

Summary of changes once more for posterity:

- Support the `source_types` hash on the pending/available top-level balance object:
https://stripe.com/docs/api/balance/balance_object#balance_object-available-source_types
- Support the `connect_reserved` top-level object on the `/v1/balances` endpoint:
https://stripe.com/docs/api/balance/balance_object#balance_object-connect_reserved
- Support the new `fpx` source type constant as a developer affordance in preparation for Stripe's broader FPX rollout (currently in private beta).

These changes are necessary to better support Stripe merchants in any markets that involve split balances for compliance/regulatory purposes. This also applies to payment methods like ACH debits in the US (where `source_type=bank_account`). 

The `connect_reserved` change also helps to broadly support connect platforms that require balance reserves for their businesses.

---

Thanks for all of the support,
Ethan

### DIFF
--- a/balance.go
+++ b/balance.go
@@ -1,5 +1,17 @@
 package stripe
 
+// BalanceSourceType is the list of allowed values for the balance amount's source_type field keys.
+type BalanceSourceType string
+
+// List of values that BalanceSourceType can take.
+const (
+	BalanceSourceTypeAlipayAccount   BalanceSourceType = "alipay_account"
+	BalanceSourceTypeBankAccount     BalanceSourceType = "bank_account"
+	BalanceSourceTypeBitcoinReceiver BalanceSourceType = "bitcoin_receiver"
+	BalanceSourceTypeCard            BalanceSourceType = "card"
+	BalanceSourceTypeFPX             BalanceSourceType = "fpx"
+)
+
 // BalanceTransactionStatus is the list of allowed values for the balance transaction's status.
 type BalanceTransactionStatus string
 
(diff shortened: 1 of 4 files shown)


balance.go
```diff
@@ -12,13 +24,15 @@ type BalanceParams struct {
...
+	Pending         []*Amount `json:"pending"`
 }
 
 // Amount is a structure wrapping an amount value and its currency.
 type Amount struct {
-	Value    int64    `json:"amount"`
-	Currency Currency `json:"currency"`
+	Value       int64                       `json:"amount"`
+	Currency    Currency                    `json:"currency"`
+	SourceTypes map[BalanceSourceType]int64 `json:"source_types"`
```
brandur-stripe: I know it was like this already, but we should consider alphabetizing this list just to make it like all the other ones. (That could also easily be pushed as a separate commit to `master` though.)

remi-stripe: fixed it, I totally did not notice at first

remi-stripe:
Thanks for all the patient work @sherbondy-stripe. I updated the PR and squashed the commits and will merge once all tests pass

PR #1055, opened 2020-04-01
stripe-go v71

brandur:
Integration branch for the next major version of stripe-go.

⚠️ = breaking change

* [x] ⚠️Remove all beta features from Issuing APIs (#1052)
* [x] ⚠️Make API response accessible on returned API structs (#1054)
* [x] Start using Go Modules (#1061)
* [x] ⚠️Multiple breaking API changes (#1068)
  * `PaymentIntent` is now expandable on `Charge`
  * `Percentage` was removed as a filter when listing `TaxRate`
  * Removed `RenewalInterval` on `SubscriptionSchedule`
  * Removed `Country` and `RoutingNumber` from `ChargePaymentMethodDetailsAcssDebit`
* [x] ⚠️Default number of network retries to 2 (#1069)
* [x] ⚠️Clean up logging for next major (#1070)
* [x] Add v71 migration guide (#1072)

PR #1061, opened 2020-04-10
Start using Go Modules

brandur:
Similar to the original implementation for Go Modules in #712, here we
add a `go.mod` and `go.sum`, then proceed to use Go Modules style
import paths everywhere that include the current major revision.

Unfortunately, this may still cause trouble for Dep users who are trying
to upgrade stripe-go, but the project's now had two years to help its
users with basic Go Module awareness, and has chosen not to do so. It's
received no commits of any kind since August 2019, and therefore would
be considered unmaintained by most definitions. Elsewhere, Go Modules
now seem to be the only and obvious way forward, so we're likely to see
more and more users on them.

`scripts/check_api_clients/main.go` is also updated to be smarter about
breaking down package paths which may now include the major.

Caveats:

* Go Coverage in tip is broken, and it wasn't immediately obvious to me
  why. We can probably get this fixed at our leisure though because it's
  marked as an allowed failure in the build.

* We still need a solution to put the major version import into the API
  reference docs. I can probably figure something out there next week.

r? @remi-stripe
cc @stripe/api-libraries

[1] https://github.com/golang/dep/pull/1963

---


brandur-stripe:
As suggested by Remi, just pushed one more commit that brings this back to v70 so that we can test that our release script works properly when the time comes.

remi-stripe:
This was rebased on the latest integration-v71 and we can now merge

PR #1065, opened 2020-04-14
Adding support for the `Price` resource and APIs

remi-stripe:
cc @stripe/api-libraries 
cc @areagle-stripe

### DIFF
--- a/price_test.go
+++ b/price_test.go
@@ -1,0 +1,68 @@
+package stripe
+
+import (
+	"encoding/json"
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+	"github.com/stripe/stripe-go/v71/form"
+)
+
+func TestPrice_Unmarshal(t *testing.T) {
+	priceData := map[string]interface{}{
+		"id":     "price_123",
+		"object": "price",
+		"recurring": map[string]interface{}{
+			"aggregate_usage": "last_during_period",
+			"interval":        "month",
+			"interval_count":  6,
+			"usage_type":      "metered",
+		},
+		"tiers": []map[string]interface{}{
+			{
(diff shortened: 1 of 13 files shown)


price_test.go
```diff
@@ -0,0 +1,68 @@
...
+	params := PriceTierParams{
+		UnitAmount: Int64(500),
+		UpToInf:    Bool(true),
+	}
+
+	// TODO: figure out why this doesn't work.
+	body := &form.Values{}
+	form.AppendTo(body, params)
+	t.Logf("body = %+v", body)
+	assert.Equal(t, []string{"inf"}, body.Get("up_to"))
```
remi-stripe, line 67: This test doesn't pass and I don't get why.

brandur-stripe: @remi-stripe This is a really subtle problem that is one of the things that are not so good in Go, but you need to have a pointer to implement an interface. Go doesn't think that your `params` above is an `Appender` implementation because it isn't one. This will fix it:

``` diff
diff --git price_test.go price_test.go
index d76be08..988317b 100644
--- price_test.go
+++ price_test.go
@@ -55,7 +55,7 @@ func TestPrice_Unmarshal(t *testing.T) {
 }
 
 func TestPriceTierParams_AppendTo(t *testing.T) {
-	params := PriceTierParams{
+	params := &PriceTierParams{
 		UnitAmount: Int64(500),
 		UpToInf:    Bool(true),
 	}
```

remi-stripe: oh wow. That explains why the code itself worked, which I tested. I was trying to simplify the tests that exist in `plan_test.go` because I didn't understand them. Thanks a lot, will confirm locally.

PR #1070, opened 2020-04-16
Clean up logging for next major

brandur:
## Background

For a very long time, stripe-go logged by way of the `Printfer`
interface:

``` go
type Printfer interface {
       Printf(format string, v ...interface{})
}
```

`Printfer` had a few problems: it didn't allow any kind of granularity
around logging levels, didn't allow us to behave well by sending errors
to `stderr` and other output to `stdout`, and wasn't interoperable with
any popular Go loggers (it was interoperable with the one in the
stdlib's `log`, but serious programs tend to move away from that).

A while back, I introduced the new `LeveledLogger` interface, which
corrected the deficiencies with `Printfer`.

We made quite a bit of effort to stay compatible with the existing
configuration options around logging like the `Logger` and `LogLevel`
globals, while also adding new ones for the leveled logger like
`DefaultLeveledLogger`.

The downside of that is it made reasoning about the different
combinations complicated. For example, you had to know that if both
`Logger` and `DefaultLeveledLogger` were set, then the former would take
precedence, which isn't necessarily obvious.

 ## Changes in this patch

Since we're on the verge of a major, I've taken the liberty of cleaning
up all the old logging infrastructure. This involves removing its types,
interfaces, helpers, global configuration options, and local
configuration options. We're left with a much simpler setup with just
`stripe.DefaultLeveledLogger` and `BackendConfig.LeveledLogger`.

Users who were already using the leveled logger (as recommended in the
README for sometime) won't have to change anything, nor will users who
had no logging configured. Users on the old infrastructure will have to
make some config tweaks, but ones that are quite easy.

And since we're changing logging things anyway: I've made a small tweak
such that if left unset by the user, the default logger now logs errors
only (instead of errors + informational). This is a more reasonable
default on Unix systems where programs are generally not expected to be
noisy over stdout. If a user wants to get informational messages back,
it's a very easy configuration tweak to `DefaultLeveledLogger`. I think
this change is okay for the reason above, but also because I don't think
a lot of thought was ever put into the original default -- we've just
been cargo culting through releases for a long time now.

r? @ob-stripe @remi-stripe
cc @stripe/api-libraries

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,8 @@ retries are safe.
 
 ### Configuring Logging
 
-Configure logging using the global `DefaultLeveledLogger` variable:
+By default, the library logs error messages only (which are sent to `stderr`).
+Configure default logging using the global `DefaultLeveledLogger` variable:
 
 ```go
 stripe.DefaultLeveledLogger = &stripe.LeveledLogger{
--- a/log.go
+++ b/log.go
@@ -3,7 +3,6 @@ package stripe
 import (
 	"fmt"
 	"io"
-	"log"
 	"os"
 )
 
@@ -12,6 +11,9 @@ import (
 //
 
 const (
+	// LevelNull sets a logger to show no messages at all.
+	LevelNull Level = 0
+
 	// LevelError sets a logger to show error messages only.
 	LevelError Level = 1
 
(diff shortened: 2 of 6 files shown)


brandur-stripe:
Thank you Richard!

PR #1071, opened 2020-04-17
Add Go 1.13 and Go 1.14 to build matrix

brandur-stripe:
Go 1.13 and 1.14 have been released since the last time we updated CI.
We don't expect any problems with them, but add them to the build matrix
for completeness.


brandur-stripe:
Thanks as usual OB!

PR #1072, opened 2020-04-17
Add V71 migration guide

brandur:
V71 has a few non-trivial breaking changes bundled in, so add a guide
with more lengthy explanations on how to handle each one (as we've done
previously on occasion like for the big V32 major).

r? @ob-stripe
cc @stripe/api-libraries

---


brandur-stripe:
Oops, `hub` improperly cached a previous PR message so the title/description of this one were originally completely wrong. Updated them now, but it's actually about a V71 migration guide rather than adding Go versions to Travis.

brandur-stripe:
Thanks as usual Olivier.

Issue #1077, opened 2020-04-22
Missing ErrorCode constant for "card_decline_rate_limit_exceeded" in error.go


jinjlu:
go version: go1.13.4 darwin/amd64
stripe library: Stripe/v1 GoBindings/67.4.1

The error code is listed in the [doc](https://stripe.com/docs/error-codes#card-decline-rate-limit-exceeded) but the constant is not presented in error.go in latest change or the version I used. 
a sample response:
```
POST /v1/charges
{
  "error": {
    "code": "card_decline_rate_limit_exceeded",
    "doc_url": "https://stripe.com/docs/error-codes/card-decline-rate-limit-exceeded",
    "message": "You have exceeded the maximum number of declines on this card in the last 24 hour period. Please contact us via https://support.stripe.com/contact if you need further assistance.",
    "type": "invalid_request_error"
  }
}
```

remi-stripe:
Thanks for the report. I opened a PR in https://github.com/stripe/stripe-go/pull/1078 and it will be fixed in the next version

jinjlu:
Awesome! Thanks for fixing it!

remi-stripe:
This is now fixed in 71.3.0, thanks again for the report!

PR #1078, opened 2020-04-22
Add missing error codes such as `ErrorCodeCardDeclinedRateLimitExceeded`

remi-stripe:
Added all the codes from [the docs](https://stripe.com/docs/error-codes. I ignored the 2 PaymentMethod ones about micro-deposits as those shouldn't be public for now.

Fixes https://github.com/stripe/stripe-go/issues/1077

r? @cjavilla-stripe 
cc @stripe/api-libraries


error.go
```diff
@@ -34,18 +34,23 @@ const (
...
+	ErrorCodeBankAccountVerificationFailed          ErrorCode = "bank_account_verification_failed"
 	ErrorCodeBitcoinUpgradeRequired                 ErrorCode = "bitcoin_upgrade_required"
+	ErrorCodeCardDeclinedRateLimitExceeded          ErrorCode = "card_decline_rate_limit_exceeded"
 	ErrorCodeCardDeclined                           ErrorCode = "card_declined"
 	ErrorCodeChargeAlreadyCaptured                  ErrorCode = "charge_already_captured"
 	ErrorCodeChargeAlreadyRefunded                  ErrorCode = "charge_already_refunded"
 	ErrorCodeChargeDisputed                         ErrorCode = "charge_disputed"
 	ErrorCodeChargeExceedsSourceLimit               ErrorCode = "charge_exceeds_source_limit"
 	ErrorCodeChargeExpiredForCapture                ErrorCode = "charge_expired_for_capture"
+	ErrorCodeChargeInvalidParameter                 ErrorCode = "charge_invalid_params"
```
cjavilla-stripe, line 50: i think this one is `charge_invalid_parameter`

error.go
```diff
@@ -64,6 +70,7 @@ const (
 	ErrorCodeInvalidSourceUsage                     ErrorCode = "invalid_source_usage"
 	ErrorCodeInvoiceNoCustomerLineItems             ErrorCode = "invoice_no_customer_line_items"
 	ErrorCodeInvoiceNoSubscriptionLineItems         ErrorCode = "invoice_no_subscription_line_items"
+	ErrorCodeInvoicePamentIntentRequiresAction      ErrorCode = "invoice_payment_intent_requires_action"
```
cjavilla-stripe, line 73: goes after `invoice_not_editable`?

error.go
```diff
@@ -82,6 +89,7 @@ const (
 	ErrorCodeParameterMissing                       ErrorCode = "parameter_missing"
 	ErrorCodeParameterUnknown                       ErrorCode = "parameter_unknown"
 	ErrorCodeParametersExclusive                    ErrorCode = "parameters_exclusive"
+	ErrorCodePaymentIntentActionRequired            ErrorCode = "payment_intent_action_required"
 	ErrorCodePaymentIntentAuthenticationFailure     ErrorCode = "payment_intent_authentication_failure"
 	ErrorCodePaymentIntentIncompatiblePaymentMethod ErrorCode = "payment_intent_incompatible_payment_method"
 	ErrorCodePaymentIntentInvalidParameter          ErrorCode = "payment_intent_invalid_parameter"
```
cjavilla-stripe, line 95: Should we add `payment_method_microdeposit_verification_amounts_mismatch` and `payment_method_microdeposit_verification_attempts_exceeded` here?

remi-stripe: No, I mentioned both in the PR descriptions, those should not be public

cjavilla-stripe: ha! you mentioned these two in the comment so disregard this.

error.go
remi-stripe: I wish I could rewind my life and understand what happened here since I copy-pasted each value by hand. Oh well.

PR #1086, opened 2020-04-29
Clean up logging in some noisier tests

brandur:
There are a few tests in the top-level package (especially in
`stripe_test.go`) that produce either (1) a lot of logging, or (2) error
logging even on a successful run. This normally isn't a problem because
Go captures output by default when `go test .` is run, but it can be a
little alarming when a test case fails and a swath of unrelated logging
is produced.

Here we still keep most test logging on by default (so you will still
see output when a test case fails), but turn off logging specifically
for tests that produce either lots of output or error output that we
expect to happen. My selection methodology was basically to go through
with a `-test.v` and eyeball for anything in either of these categories.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -23,6 +23,10 @@ func TestErrorResponse(t *testing.T) {
 	defer ts.Close()
 
 	backend := GetBackendWithConfig(APIBackend, &BackendConfig{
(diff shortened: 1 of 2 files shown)


brandur-stripe:
No worries!

PR #1091, opened 2020-05-04
Add support for the `LineItem` resource and APIs

remi-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,7 @@ cache:
 env:
   global:
     # If changing this number, please also change it in `testing/testing.go`.
-    - STRIPE_MOCK_VERSION=0.89.0
+    - STRIPE_MOCK_VERSION=0.90.0
 
 go:
   - "1.10.x"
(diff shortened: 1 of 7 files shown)


remi-stripe:
```golint -set_exit_status ./...
lineitem/client.go:9:6: type name will be used as lineitem.LineItemIter by other packages, and that stutters; consider calling this Iter
Found 1 lint suggestions; failing.
make: *** [lint] Error 1```

@brandur-stripe Any chance you could have a look at this PR tomorrow morning? I'm struggling to understand how to represent the `LineItem` resource's ability to be paginated. We don't really have a pattern line this today in stripe-go and I'm not sure what the right way to approach this is.

brandur-stripe:
@remi-stripe Hey! Just taking a look at some of our existing conventions, it looks like we usually call the "main" iterator in any particular client package just `Iter` usually, with any "secondary" iterators getting an object prefix. For example:

* In `creditnote`, the credit note iterator is `Iter` while the line item iterator is `LineItemIter`.
* In `invoice`, the invoice iterator is `Iter` while the line iterator is `LineIter`.

I imagine for this one you just ended up copying the `LineItemIter` from `creditnote` — would suggest renaming it to just `Iter`.

Issue #1092, opened 2020-05-06
Missing "Metadata" for "CheckoutSessionSubscriptionDataParams"


RensTillmann:
https://github.com/stripe/stripe-go/blob/b363a1c3d64708d837a96d32a498b59d6287ab7b/checkout_session.go#L98

remi-stripe:
@RensTillmann This was fixed in 71.9.0. Thanks again for the report!

RensTillmann:
@remi-stripe Thanks, you are welcome :)

PR #1093, opened 2020-05-07
Add support for `Metadata` for `PaymentIntentData` and `SubscriptionData` on Checkout `Session`

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/1092

r? @ob-stripe 
cc @stripe/api-libraries

### DIFF
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -62,6 +62,7 @@ type CheckoutSessionPaymentIntentDataParams struct {
 	ApplicationFeeAmount      *int64                                              `form:"application_fee_amount"`
 	CaptureMethod             *string                                             `form:"capture_method"`
 	Description               *string                                             `form:"description"`
+	Metadata                  map[string]string                                   `form:"metadata"`
 	OnBehalfOf                *string                                             `form:"on_behalf_of"`
 	ReceiptEmail              *string                                             `form:"receipt_email"`
 	SetupFutureUsage          *string                                             `form:"setup_future_usage"`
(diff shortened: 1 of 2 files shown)


checkout_session.go
```diff
@@ -62,6 +62,7 @@ type CheckoutSessionPaymentIntentDataParams struct {
 	ApplicationFeeAmount      *int64                                              `form:"application_fee_amount"`
 	CaptureMethod             *string                                             `form:"capture_method"`
 	Description               *string                                             `form:"description"`
+	Metadata                  map[string]string                                   `form:"metadata"`
```
remi-stripe, line 65: I opted for `map[string]string` for consistency with plan and price. In theory it should be a pointer I think but in this case it's not updatable so it shouldn't matter right?

brandur-stripe: Maps are kind of a special type in that they're always reference types, and so you shouldn't use a pointer with them. See [this blog post on them](https://blog.golang.org/maps):

> ``` go
> var m map[string]int
> ```
>
> Map types are reference types, like pointers or slices, and so the value of m above is nil; it doesn't point to an initialized map. A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic; don't do that. To initialize a map, use the built in make function:
>
> ```
> m = make(map[string]int)
> ```

remi-stripe: Ah nice, so we didn't overlook this originally when we used map. Thanks for clarifying @brandur-stripe !

PR #1097, opened 2020-05-14
Fixing `LineItems` to be ` LineItemList` on Checkout `Session`

remi-stripe:
r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -203,7 +203,7 @@ type CheckoutSession struct {
 	Deleted                   bool                                      `json:"deleted"`
 	DisplayItems              []*CheckoutSessionDisplayItem             `json:"display_items"`
 	ID                        string                                    `json:"id"`
-	LineItems                 []*LineItem                               `json:"line_items"`
+	LineItems                 *LineItemList                             `json:"line_items"`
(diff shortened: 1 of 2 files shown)


checkout_session.go
```diff
@@ -203,7 +203,7 @@ type CheckoutSession struct {
 	Deleted                   bool                                      `json:"deleted"`
 	DisplayItems              []*CheckoutSessionDisplayItem             `json:"display_items"`
 	ID                        string                                    `json:"id"`
-	LineItems                 []*LineItem                               `json:"line_items"`
+	LineItems                 *LineItemList                             `json:"line_items"`
```
cjavilla-stripe, line 206: Is that a thing? Not seeing `LineItemList` defined...

cjavilla-stripe: nevermind i didn't have latest master...

remi-stripe: It's defined in `lineitem.go` https://github.com/stripe/stripe-go/blob/master/lineitem.go#L27-L32

All resources have a `<Resource>List` version. And so if you look at `Customer` it uses `SubscriptionList` for example: https://github.com/stripe/stripe-go/blob/master/customer.go#L117

Issue #1098, opened 2020-05-15
unable to update beyond v70.15.0


adlmr:
Hello,
I'm using `github.com/stripe/stripe-go v70.15.0+incompatible` as written on go.mod and i have issues with `CheckoutSession LineItem` so i want to update to latest version.

go says:
`go get github.com/stripe/stripe-go@6cf67b3fa584: github.com/stripe/stripe-go@v1.0.4-0.20200514011123-6cf67b3fa584: invalid version: go.mod has post-v1 module path "github.com/stripe/stripe-go/v71" at revision 6cf67b3fa584`

if i put `github.com/stripe/stripe-go latest` in go.mod i get `github.com/stripe/stripe-go v70.15.0+incompatible`

remi-stripe:
@adlmr With 71.* we started using go modules. This means that when you include stripe-go you now need to reference the major version you are depending on. This is documented [here](https://github.com/stripe/stripe-go#installation)

```
import (
    "github.com/stripe/stripe-go/v71"
    "github.com/stripe/stripe-go/v71/customer"
)
```

ls-dac-chartrand:
I'm having the same issue.

When [I look at your docs](https://github.com/stripe/stripe-go#installation), it says I should I type:

```
go get -u github.com/stripe/stripe-go
```

When I do that I get: `go: github.com/stripe/stripe-go upgrade => v70.15.0+incompatible`

### Workaround

```
go get -u github.com/stripe/stripe-go/v71
```

brandur-stripe:
@ls-dac-chartrand I'll get that README updated, but in modern Go, you really don't need to `go get -u` anything anymore.

If you just put the stripe-go module in one of your `require`s then do a `go build`, Go will fetch all the necessary dependencies for you automatically.

brandur-stripe:
Pitched an amendment to the README in #1100.

PR #1100, opened 2020-05-20
Fix misleading install instructions for stripe-go

brandur:
Previous to Go Modules, it was often necessary to manually fetch
dependencies with `go get` before using them.

With Go Modules, the much more standard way now is to make sure a
project is initialized with Go Modules (`go mod init`), then just start
including modules in your `import` paths. Go will resolve them
automatically.

This came up in #1098 because our instructions still recommend `go get`.
Here, we change them to recommend using `go mod init` + `import`
instead.

r? @remi-stripe

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,13 +8,14 @@ The official [Stripe][stripe] Go client library.
 
 ## Installation
 
-Install stripe-go with:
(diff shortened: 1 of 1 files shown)


brandur-stripe:
TY.

PR #1101, opened 2020-05-21
Multiple API Changes

cjavilla-stripe:
r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -40,6 +40,7 @@ const (
 	ChargePaymentMethodDetailsTypeFPX               ChargePaymentMethodDetailsType = "fpx"
 	ChargePaymentMethodDetailsTypeGiropay           ChargePaymentMethodDetailsType = "giropay"
 	ChargePaymentMethodDetailsTypeIdeal             ChargePaymentMethodDetailsType = "ideal"
+	ChargePaymentMethodDetailsTypeInteracPresent    ChargePaymentMethodDetailsType = "interac_present"
 	ChargePaymentMethodDetailsTypeKlarna            ChargePaymentMethodDetailsType = "klarna"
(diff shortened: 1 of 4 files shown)


charge.go
```diff
@@ -379,6 +380,36 @@ type ChargePaymentMethodDetailsIdeal struct {
...
+	Funding        string                                           `json:"funding"`
+	GeneratedCard  string                                           `json:"generated_card"`
+	Last4          string                                           `json:"last4"`
+	Network        string                                           `json:"network"`
+	ReadMethod     string                                           `json:"read_method"`
+	Receipt        *ChargePaymentMethodDetailsInteracPresentReceipt `json:"receipt"`
+}
+
+// ChargePaymentMethodDetailsInteracPresentReceipt represents details about the InteracPresent Receipt.
+type ChargePaymentMethodDetailsInteracPresentReceipt struct {
```
remi-stripe, line 401: I always put "nested" resources above in go so this should be before. Not blocking but FYI

sub.go
```diff
@@ -95,7 +95,8 @@ type SubscriptionPendingInvoiceItemIntervalParams struct {
 
 // SubscriptionTransferDataParams is the set of parameters allowed for the transfer_data hash.
 type SubscriptionTransferDataParams struct {
-	Destination *string `form:"destination"`
+	AmountPercent *float64 `form:"amount_percent"`
```
remi-stripe, line 98: I don't understand this change at all. It's already in master https://github.com/stripe/stripe-go/blob/master/sub.go#L230-L234
I don't understand what's going on with the diff here.

remi-stripe:
Okay worked with @jlomas-stripe on this and the issue was that the branch was not on a recent master and for some reason it showed the commit's changes as if they were new even if master had them.

Issue #1107, opened 2020-05-29
TopUp Object does not include "Metadata"


nemosupremo:
https://github.com/stripe/stripe-go/blob/d5e092ab5fe6391f076cf2f7cde5756d346006ac/topup.go#L40

The TopUp object does not have a metadata key - although the TopUp object in the API reference supports storing and retrieving metadata.

remi-stripe:
@nemosupremo Thanks for the report! I opened a PR to fix this and we'll release the fix tomorrow!

remi-stripe:
This has been fixed in 71.16.0

PR #1108, opened 2020-05-29
Add `Metadata` and `Object` on `Topup`

remi-stripe:
Fix https://github.com/stripe/stripe-go/issues/1107

r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/topup.go
+++ b/topup.go
@@ -40,7 +40,6 @@ type TopupList struct {
 type Topup struct {
 	APIResource
 	Amount                   int64               `json:"amount"`
-	ArrivalDate              int64               `json:"arrival_date"`
 	BalanceTransaction       *BalanceTransaction `json:"balance_transaction"`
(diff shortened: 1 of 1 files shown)


topup.go
```diff
@@ -50,8 +49,13 @@ type Topup struct {
 	FailureMessage           string              `json:"failure_message"`
 	ID                       string              `json:"id"`
 	Livemode                 bool                `json:"livemode"`
+	Metadata                 map[string]string   `json:"metadata"`
+	Object                   string              `json:"object"`
```
cjavilla-stripe, line 53: I think this one is an enum with just one value `topup` for now, right? Do we want to add that one value as TopupObject or something?

remi-stripe: We've never done this so no. It's really just a string and I've added it manually to resources when I catch it.

PR #1115, opened 2020-06-04
Add support for `DeclineCode` on `CardError` top-level

### LINKED ISSUE #1114: Missing DeclineCode when CardError on `Error` top-level
The Error object has most properties top-level, also the DeclineCode. (add by #949)

like #1064 We can get `card_declined` from `CardError`.
When we print the err.(Error) object, card_declined is empty.

If Error.DeclineCode is added, it should be initialized when CardError occurs.
Otherwise, remove it.

luw2007:
hi there,

If Error.DeclineCode is added, it should be initialized when CardError occurs.
Otherwise, remove it.

fix #1114

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -550,6 +550,7 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
 	case ErrorTypeCard:
 		cardErr := &CardError{stripeErr: raw.E.Error}
 		if raw.E.DeclineCode != nil {
+			cardErr.stripeErr.DeclineCode = *raw.E.DeclineCode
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Hey @luw2007, thanks for the patch! I dug into this one a little bit, and found that the reason `DeclineCode` was not being populated was actually do to a more subtle internal bug. Because of that, I've opted to fix this in a slightly different way than you did in #1116.

BTW, I think our our CIBot closed your pull request a little too aggressively. We should probably fix that, but for future reference: it might be easier to use a named feature branch on your forked repo (e.g. `fix-decline-code`) instead of `master` just so that this doesn't happen again.

PR #1116, opened 2020-06-04
Fix top-level `DeclineCode` deserialization in errors

### LINKED ISSUE #1114: Missing DeclineCode when CardError on `Error` top-level

brandur:
Issue #1114 reveals a bug with `DeclineCode` on the top-level error
object.

In #949, we added `DeclineCode` on the top-level error object in
addition to where it was before on the specific `CardError` type as a
minor quality of life improvement.

It seemed like a very simple change, so we didn't put in a test, but
what we forgot about at the time was that we were already doing some
weird stuff in order to extract `DeclineCode` for `CardError`.
Specifically, we'd added this special wrapper type that extended
`Error`:

``` go
// rawErrorInternal embeds Error to deserialize all the standard error fields,
// but also adds other fields that may or may not be present depending on error
// type to help with deserialization. (e.g. DeclineCode).
type rawErrorInternal struct {
	*Error
	DeclineCode *DeclineCode `json:"decline_code,omitempty"`
}
```

It added `DeclineCode` in addition to the `DeclineCode` on the embedded
`Error`. The trouble was that Go will only deserialize any given field
_once_, so `rawErrorInternal.DeclineCode` was being populated, and
`rawErrorInternal.Error.DeclineCode` was not.

I fix the problem by essentially eliminating `rawErrorInternal` and
flattening our error type structure somewhat. This results in slightly
nicer code, and also fixes deserialization of `Error.DeclineCode`.
We continue to also support `CardError.DeclineCode` for backwards
compatibility.

Fixes #1114.

r? @remi-stripe
cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -520,11 +520,11 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
 	if s.Type == ConnectBackend {
 		// If this is an OAuth request, deserialize as Error because OAuth errors
 		// are a different shape from the standard API errors.
-		var topLevelError rawErrorInternal
+		var topLevelError Error
 		if err := s.UnmarshalJSONVerbose(res.StatusCode, resBody, &topLevelError); err != nil {
 			return err
 		}
-		raw.E = &topLevelError
+		raw.Error = &topLevelError
 	} else {
 		if err := s.UnmarshalJSONVerbose(res.StatusCode, resBody, &raw); err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


stripe.go
```diff
@@ -520,51 +520,56 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
...
-	raw.E.RequestID = res.Header.Get("Request-Id")
+	raw.Error.HTTPStatusCode = res.StatusCode
+	raw.Error.RequestID = res.Header.Get("Request-Id")
 
 	var typedError error
-	switch raw.E.Type {
+	switch raw.Error.Type {
 	case ErrorTypeAPI:
-		typedError = &APIError{stripeErr: raw.E.Error}
+		typedError = &APIError{stripeErr: raw.Error}
```
brandur-stripe, line 545: All the changes in here are basically because I "flattened" the error types by killing `rawErrorInternal`. I also renamed `rawError.E` to `rawError.Error` to make it a little more legible.

stripe_test.go
```diff
@@ -900,6 +900,7 @@ func TestResponseToError(t *testing.T) {
 	assert.Equal(t, res.Header.Get("Request-Id"), stripeErr.RequestID)
 	assert.Equal(t, res.StatusCode, stripeErr.HTTPStatusCode)
 	assert.Equal(t, expectedErr.Type, stripeErr.Type)
+	assert.Equal(t, expectedDeclineCode, stripeErr.DeclineCode)
```
brandur-stripe, line 903: Same new test statement from #1115.

PR #1117, opened 2020-06-05
Add `PaymentMethodIdealParams` to `PaymentMethodParams`

si-m:
This PR adds the missing `ideal` parameters to the `PaymentMethodParams`.

This parameters are referenced on the stripe [documentation](https://stripe.com/docs/api/payment_methods/object#payment_method_object-ideal) but it was not available to use it when creating a payment method with the go client.


remi-stripe:
@si-m Just released as 71.21.0

PR #1120, opened 2020-06-10
Add support for Cartes Bancaires payments on `PaymentIntent` and `PaymentMethod`

remi-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -190,6 +190,7 @@ type PaymentIntentPaymentMethodOptionsCardInstallmentsParams struct {
 type PaymentIntentPaymentMethodOptionsCardParams struct {
 	Installments        *PaymentIntentPaymentMethodOptionsCardInstallmentsParams `form:"installments"`
 	MOTO                *bool                                                    `form:"moto"`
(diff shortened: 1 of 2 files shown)


paymentintent.go
```diff
@@ -190,6 +190,7 @@ type PaymentIntentPaymentMethodOptionsCardInstallmentsParams struct {
 type PaymentIntentPaymentMethodOptionsCardParams struct {
 	Installments        *PaymentIntentPaymentMethodOptionsCardInstallmentsParams `form:"installments"`
 	MOTO                *bool                                                    `form:"moto"`
+	Network             *string                                                  `form:"network"`
```
richardm-stripe, line 193: Would this normally be one of those type alias "enum"s eg
```
const (
 NetworkAmex
 NetworkCartesBancaires
 NetworkDiners
...
```
or do we not do that always?

remi-stripe: We don't use enums for params because those are pointers and you'd need a custom converter for every single enums.

PR #1125, opened 2020-06-17
Add support for Sofort on `PaymentMethod` and `PaymentIntent`

richardg-stripe:
Adding on behalf a user that needs this. It's my first time writing Go.

I grepped for 'ideal' in the files paymentintent.go and paymentmethod.go and followed the same pattern for Sofort. 

I don't need PaymentMethods but I figured it makes sense to add Sofort for Payment Intents / Payment Methods only once!


remi-stripe:
The changes look good though Sofort is not yet public and is in private beta. We don't add features in private beta to client libraries, to avoid breaking change to the library whenever we make changes.
This means we wouldn't merge this until Sofort gets released publicly (or closer to the release date when the API is guaranteed not to change).

remi-stripe:
closing in favor of https://github.com/stripe/stripe-go/pull/1180 since it was missing `payment_method_options` but I mostly copy-pasted what you had, thanks for making it easy!

PR #1131, opened 2020-07-13
Adds billing_cycle_anchor to default_settings and phases for SubscriptionSchedules

cjavilla-stripe:
* Adds `billing_cycle_anchor` to `default_settings` and `phases` for `SubscriptionSchedules`

r? @remi-stripe 
cc @stripe/api-libraries

### DIFF
--- a/subschedule.go
+++ b/subschedule.go
@@ -27,6 +27,16 @@ const (
 	SubscriptionScheduleStatusTrialing  SubscriptionScheduleStatus = "released"
 )
 
+// SubscriptionSchedulePhaseBillingCycleAnchor is the list of allowed values for the
+// schedule's billing_cycle_anchor.
+type SubscriptionSchedulePhaseBillingCycleAnchor string
+
+// List of values for SubscriptionSchedulePhaseBillingCycleAnchor
+const (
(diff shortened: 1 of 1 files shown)


subschedule.go
```diff
@@ -27,6 +27,16 @@ const (
 	SubscriptionScheduleStatusTrialing  SubscriptionScheduleStatus = "released"
 )
 
+// SubscriptionScheduleBillingCycleAnchor is the list of allowed values for the
+// schedule's billing_cycle_anchor.
+type SubscriptionScheduleBillingCycleAnchor string
```
remi-stripe, line 32: I'm curious if this is the right name? That naming implies "top level" and it's not. It's nested in DefaultSettings and Phases. I think we want `SubscriptionSchedulePhaseBillingCycleAnchor` instead

cjavilla-stripe: Okay, yeah I can add that name wasn't sure if we add Phase if that implies it wouldn't be used for default settings, but I can see that a default settings is kinda like an abstract version of a phase?

remi-stripe: yep that's how my brain thought of this and how I usually name things. No idea how we will encode that in codegen lol

PR #1137, opened 2020-07-18
The title property is missing from the SigmaScheduledQueryRun object

jredl-va:
This morning while parsing a received webhook I noticed that the title field is not included within the the golang struct: https://stripe.com/docs/api/sigma/scheduled_queries/object#scheduled_query_run_object-title


remi-stripe:
Released in 71.33.0

jredl-va:
Thanks for getting this in!

PR #1139, opened 2020-07-21
Test only on Go 1.13+

brandur:
Unfortunately, Testify introduced use of `errors.As` and `errors.Is`
[1], which are only available in Go 1.13+, and therefore broke use on
all older versions of Go. Speaking in a purely technical sense only,
they're in the right because only the last two Go minors are supported
at any given time (currently, 1.13 and 1.14).

We still want to support older versions of Go for now though, so here we
modify the build to only test on 1.13+. We still build the older
versions of Go which gives us decent assurance they still work. It's not
perfect, but is a decent compromise given the circumstances in that at
least they're still in our build matrix.

r? @remi-stripe
cc @stripe/api-libraries

[1] https://blog.golang.org/go1.13-errors#TOC_3.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,20 @@ before_install:
   # Install various build dependencies. We use `travis_retry` because `go get`
   # will occasionally fail intermittently.
 
-  # The testify require framework is used for assertions in the test suite
-  - travis_retry go get -u github.com/stretchr/testify/require
+  # Testify is used for assertions in the test suite.
+  #
+  # Unfortunately, they broke versions of Go < 1.13. so we conditionally fetch
+  # and test only on newer versions of Go. We can back this conditional out
+  # once we've dropped < 1.13.
+  #
+  # Comparing versions lexically like this is technically not safe, but I'm
+  # cheating a bit by taking advantage of the fact that we only support > 1.9,
+  # so comparing Go versions lexically will be safe for quite some time -- well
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,8 +2,20 @@ before_install:
...
+  # Unfortunately, they broke versions of Go < 1.13. so we conditionally fetch
+  # and test only on newer versions of Go. We can back this conditional out
+  # once we've dropped < 1.13.
+  #
+  # Comparing versions lexically like this is technically not safe, but I'm
+  # cheating a bit by taking advantage of the fact that we only support > 1.9,
+  # so comparing Go versions lexically will be safe for quite some time -- well
+  # beyond the point where we back this out. `tip` also sorts after any version.
+  - |
+    if [[ "$TRAVIS_GO_VERSION" > "1.13" ]]; then
```
brandur-stripe, line 16: See comment above this line — I know comparisons like this are generally problematic, but I think that we're okay in this particular case (because it works, and also because I couldn't easily find any built-ins to help). I manually checked every build in the matrix to make sure the expected thing is happening. If I'm wrong down the line then whoever notices a problem is free to yell at me.

Note that `1.13.x > 1.13`, `1.13.0 > 1.13`, `1.13.10 > 1.13`, etc.

brandur-stripe:
Thanks Remi!

PR #1142, opened 2020-07-24
Copy the JSON data of EphemeralKey

shogo82148:
The godoc of json.Unmarshaler says:

> https://golang.org/pkg/encoding/json/#Unmarshaler
> UnmarshalJSON must copy the JSON data if it wishes to retain the data after returning.

This code comes from the implementation of RawMessage.UnmarshalJSON.

https://github.com/golang/go/blob/074f2d800f2c7b741a080081cfcc5295b375b23d/src/encoding/json/stream.go#L275

### DIFF
--- a/ephemeralkey.go
+++ b/ephemeralkey.go
@@ -46,7 +46,7 @@ func (e *EphemeralKey) UnmarshalJSON(data []byte) error {
 		*e = EphemeralKey(ee)
 	}
 
-	e.RawJSON = data
+	e.RawJSON = append(e.RawJSON[:0], data...)
 
 	return nil
 }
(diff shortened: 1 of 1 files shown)


remi-stripe:
@shogo82148 Thanks for the PR, we'll have a look hopefully next week!

ephemeralkey.go
```diff
@@ -46,7 +46,7 @@ func (e *EphemeralKey) UnmarshalJSON(data []byte) error {
 		*e = EphemeralKey(ee)
 	}
 
-	e.RawJSON = data
+	e.RawJSON = append(e.RawJSON[:0], data...)
```
brandur-stripe, line 49: Would you mind adding a comment above this line to the effect of:

```
// Go does guarantee the longevity of `data`, so copy when assigning `RawJSON`
```

shogo82148: Thanks for your review. I added a comment.

brandur-stripe:
Thank you!

brandur-stripe:
Released as 71.39.0.

PR #1148, opened 2020-07-30
Make original list object accessible on iterators

### LINKED ISSUE #1141: List based calls don't set RequestID
Go version: 1.14.2
Stripe Go version: v71.28.0

I was having an issue when trying to list charges on a customer with many charges where sometimes the return set was null. I added logic to log the stripe request ID to further investigate but saw that the stripe request ID was never being logged. My logging logic was doing a bounds check like so
```golang
if len(charges) > 0 && charges[0].LastResponse != nil {
    // add transactions[0].LastResponse.RequestID to logging context
}
```
and it was never being met.

I entered the debugger and saw that `LastResponse` was nil on every charge in the slice.
I read the client code in the charge package and I think I identified where the `LastResponse` information is being lost.
https://github.com/stripe/stripe-go/blob/v71.28.0/charge/client.go#L79-L86
```golang
// List returns an iterator that iterates all charges.
func (c Client) List(listParams *stripe.ChargeListParams) *Iter {
	return &Iter{stripe.GetIter(listParams, func(p *stripe.Params, b *form.Values) ([]interface{}, stripe.ListMeta, error) {
		list := &stripe.ChargeList{}
		err := c.B.CallRaw(http.MethodGet, "/v1/charges", c.Key, b, p, list)

		ret := make([]interface{}, len(list.Data))
		for i, v := range list.Data {
			ret[i] = v
		}

		return ret, list.ListMeta, err
	})}
}
```

The value `list` does contain a non nil`LastResponse`, which contains helpful information for debugging, but each item in `list.Data` does not contain such field. As we can see here the only information parsed from `list` are these data items as they are used to construct the `Iter` instance.

I believe this behavior extends to the other list based calls and I'm wondering if this was intentional.

If a change is required I am happy to draft a PR if I could learn more about your contribution process

brandur:
This patch makes some tweaks to `Iter` so that it doesn't just retain
list metadata, but a reference to the entirety of the list object.

This is mainly useful so that it'd be possible to access the last API
response, which could be used for things like this:

``` go
requestID := it.CustomerList().LastResponse.RequestID
```

Currently, it's not possible to access a list operation's last response
because we only set `LastResponse` on the top level API resource (which
is the list in this case), and `Iter` provides no access to that.

I also introduce a new `ListContainer` interface that all lists inherit
through their embedded `ListMeta` just to convey more typing information
so that we don't have to use an untyped `interface{}`. I think this is
fine though because it's very nicely symmetric with what we already have
for `ListParamsContainer` which has a nearly identical implementation.

The downside is that this changes the definition of `Query`:

``` go
// Old definition
type Query func(*Params, *form.Values) ([]interface{}, ListMeta, error)

// New definition
type Query func(*Params, *form.Values) ([]interface{}, ListContainer, error)
```

This is technically a breaking change because `Query` is exported from
the `stripe` package, but it's one of those changes that's very unlikely
to affect that vast majority of users because it's not a type that you'd
ever be using most of the time.

We also have to make a slight change to every existing `List` function
to adopt the new `Query` signature. I've shown what this would look like
in the changes in `customer/client.go`. Unfortunately, I can't find many
solutions to getting access to `LastResponse` that _don't_ involve
changing every `List` implementation to some degree. Similarly, we also
add a new function to every resource's `Iter` like `CustomerList` to get
back a typed version of the list object, similar to how each already has
a function to get a typed version of its API resource like `Customer()`.

Fixes #1141.

r? @remi-stripe Can you take a perfunctory initial review for this one?
Also see my last comment in #1141.

### DIFF
--- a/charge/client.go
+++ b/charge/client.go
@@ -74,7 +74,7 @@ func List(params *stripe.ChargeListParams) *Iter {
 
 // List returns an iterator that iterates all charges.
 func (c Client) List(listParams *stripe.ChargeListParams) *Iter {
-	return &Iter{stripe.GetIter(listParams, func(p *stripe.Params, b *form.Values) ([]interface{}, stripe.ListMeta, error) {
+	return &Iter{stripe.GetIter(listParams, func(p *stripe.Params, b *form.Values) ([]interface{}, stripe.ListContainer, error) {
 		list := &stripe.ChargeList{}
 		err := c.B.CallRaw(http.MethodGet, "/v1/charges", c.Key, b, p, list)
 
@@ -83,7 +83,7 @@ func (c Client) List(listParams *stripe.ChargeListParams) *Iter {
 			ret[i] = v
 		}
 
-		return ret, list.ListMeta, err
+		return ret, list, err
 	})}
 }
 
@@ -95,6 +95,13 @@ type Iter struct {
 // Charge returns the charge which the iterator is currently pointing to.
 func (i *Iter) Charge() *stripe.Charge {
 	return i.Current().(*stripe.Charge)
+}
+
+// ChargeList returns the current list object which the iterator is currently
+// using. List objects will change as new API calls are made to continue
+// pagination.
+func (i *Iter) ChargeList() *stripe.ChargeList {
+	return i.List().(*stripe.ChargeList)
 }
 
 func getC() Client {
--- a/iter.go
+++ b/iter.go
@@ -22,8 +22,9 @@ type Iter struct {
 	cur        interface{}
 	err        error
 	formValues *form.Values
+	list       ListContainer
 	listParams ListParams
-	meta       ListMeta
+	meta       *ListMeta
 	query      Query
 	values     []interface{}
 }
@@ -42,9 +43,15 @@ func (it *Iter) Err() error {
 	return it.err
 }
 
+// List returns the current list object which the iterator is currently using.
+// List objects will change as new API calls are made to continue pagination.
+func (it *Iter) List() ListContainer {
+	return it.list
+}
+
 // Meta returns the list metadata.
 func (it *Iter) Meta() *ListMeta {
-	return &it.meta
+	return it.meta
 }
 
 // Next advances the Iter to the next item in the list,
@@ -73,7 +80,8 @@ func (it *Iter) Next() bool {
 }
 
 func (it *Iter) getPage() {
-	it.values, it.meta, it.err = it.query(it.listParams.GetParams(), it.formValues)
+	it.values, it.list, it.err = it.query(it.listParams.GetParams(), it.formValues)
+	it.meta = it.list.GetListMeta()
 
 	if it.listParams.EndingBefore != nil {
 		// We are moving backward,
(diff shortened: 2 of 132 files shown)


iter.go
```diff
@@ -22,8 +22,9 @@ type Iter struct {
 	cur        interface{}
 	err        error
 	formValues *form.Values
+	list       ListContainer
 	listParams ListParams
-	meta       ListMeta
+	meta       *ListMeta
```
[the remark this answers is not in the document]
brandur-stripe: Nah, we are actually good on that one luckily! `meta` is lowercase and therefore not exported from the package.

As noted elsewhere though, the change to `Query` is technically breaking.

brandur-stripe:
Thanks Remi.

> I flagged what I think might be a breaking change.

Yeah, it's actually the change to `Query` that makes this breaking, but yes, it is technically breaking. It's only "mildly" breaking though in that while `Query` is exported, it's a type that we'd never expect a user to use. It's only exported so that each of the resource packages (`customer/`, `plan/`, etc.) can get at it.

> Could you also add some tests to cover this behavior?

I added a basic test around the accessors on `Iter`. The code that sets the last response on an `APIResource` is extensively tested elsewhere already though, so I didn't bother testing any of that.

@ob-stripe / @richardm-stripe Definitely feel free to take a look as well. I'll push up the rest of the changes with all the other `List` functions changed tomorrow or Monday.

iter_test.go
```diff
@@ -9,23 +9,23 @@ import (
 )
 
 func TestIterEmpty(t *testing.T) {
-	tq := testQuery{{nil, ListMeta{}, nil}}
+	tq := testQuery{{nil, &ListMeta{}, nil}}
```
brandur-stripe, line 12: These all changes so that I can have `testQuery` take a `ListContainer` instead of `ListMeta`. `ListContainer` is an interface, so it requires a pointer.

charge/client.go
```diff
@@ -97,6 +97,13 @@ func (i *Iter) Charge() *stripe.Charge {
 	return i.Current().(*stripe.Charge)
 }
 
+// ChargeList returns the current list object which the iterator is currently
+// using. List objects will change as new API calls were made to continue
```
richardm-stripe, line 101:
```suggestion
// using. List objects will change as new API calls are made to continue
```

brandur-stripe: Oops, good catch. Thanks Richard!

brandur-stripe:
Alright, open for a final round of feedback on this one everyone! Core code hasn't changed, but I've now modified `List` and `Iter` in every package. Tests are now passing.

brandur-stripe:
And did one final round of adding tests for all new accessors to get our coverage back up (which also caught a couple errors, phew).

brandur-stripe:
Just rebased and added the updates to `promotioncode` as well. Going to bring this in.

I talked this over with Remi, and although there is technically a breaking change here to an exported type (`Query`), because it's only supposed to be used internally for other stripe-go packages and therefore won't be breaking to the vast majority of users, I'm going to release this is a new minor version instead of going for a major. I'll add a more detailed note in the CHANGELOG for how to fix a breakage if you were one of the rare few to run into one.

brandur-stripe:
Released as 71.44.0.

PR #1149, opened 2020-07-30
Add a note on using `go get` to get the package

brandur-stripe:
Adds a note to the `README.md` that mentions how it's also possible to fetch the project using `go get`.


brandur-stripe:
This is a tiny README amendment, so just going to self-approve and merge.

PR #1152, opened 2020-08-04
Add support for `AccountType` in `ChargePaymentMethodDetailsCardPresentReceipt`

remi-stripe:
That one is an enum while the previous one for `interac_present` was a string so I went and did the constants!

r? @cjavilla-stripe 
cc @stripe/api-libraries

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -21,6 +21,17 @@ const (
 	ChargeFraudStripeReportFraudulent ChargeFraudStripeReport = "fraudulent"
 )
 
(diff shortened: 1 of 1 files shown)


cjavilla-stripe:
Do you want to add the constants for the interact_present account types? they are not the same

remi-stripe:
> Do you want to add the constants for the interact_present account types? they are not the same

I can't, changing from `string` to the constant type would be a breaking change so no. We would do this in a future major, but more likely when we codegen instead!

PR #1156, opened 2020-08-12
Re-enable H/2 on the default HTTP client for Go 1.15+

dcarney-stripe:
[Go 1.15](https://blog.golang.org/go1.15) fixes a [long-standing bug](https://github.com/golang/go/issues/32441) in the HTTP/2 client that led to HTTP/2 being disabled by default in stripe-go as part of https://github.com/stripe/stripe-go/pull/903.

This PR modifies the default HTTP/2 client for those users running Go 1.15 or above to remove the restriction against using HTTP/2.  It accomplishes this via the use of a `+build` tag (aka conditional compliation) of the new source file `stripego115.go`.

The bug is scheduled to be backported to [1.13.16](https://github.com/golang/go/issues/39278) and [1.14.8](https://github.com/golang/go/issues/39279), but even once those backports are complete, we can only conditionally compile against a _major_ release of Go, per https://golang.org/cmd/go/#hdr-Build_constraints:

> - a term for each Go major release, through the current version:
> “go1.1” from Go version 1.1 onward, “go1.12" from Go 1.12, and so on.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -1091,6 +1091,11 @@ var httpClient = &http.Client{
 	// own HTTP client with it enabled. See `testing/testing.go`.
 	//
 	// (Written 2019/07/24.)
+	//
+	// UPDATE: With the release of Go 1.15, this bug has been fixed.
+	// As such, stripego115.go contains conditionally-compiled code that sets
+	// a different HTTP client as the default, since Go 1.15+ does not contain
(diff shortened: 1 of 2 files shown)


brandur-stripe:
Thanks @dcarney-stripe! This LGTM.

We can re-evaluate a potentially broader application of an HTTP/2 client in a few months depending on how those backports are coming.

PR #1157, opened 2020-08-12
Add a note on using `go get` to get the package

brandur:
Adds a note to the `README.md` that mentions how it's also possible to
fetch the project using `go get`.


brandur-stripe:
Weird, somehow my `hub` totally failed to grab the last commit message. Will open a fresh PR with the right one.

PR #1162, opened 2020-08-17
Add secret to ephemeral key.

hectorfh:
Ephemeral key won't work without "secret" attribute.

### DIFF
--- a/ephemeralkey.go
+++ b/ephemeralkey.go
@@ -25,6 +25,7 @@ type EphemeralKey struct {
 	Expires  int64  `json:"expires"`
 	ID       string `json:"id"`
 	Livemode bool   `json:"livemode"`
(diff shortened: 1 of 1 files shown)


remi-stripe:
@hectorfh You should never have to access `Secret` on `EphemeralKey`. This resource is supposed to be "opaque" in the sense that you're simply getting the `RawJSON` value and sending that back to the Mobile SDK for them to deserialize it directly. You never have to read the properties yourself or get the secret directly, which is why it's not in stripe-go today.

Does that make sense?

hectorfh:
Hey @remi-stripe ,

thanks for your quick answer. I see. It makes sense now. I'll forward the raw json to the mobile app.

Thank you very much!!!

Regards,
Héctor Hernández

PR #1178, opened 2020-09-02
Fix the constant names for `BankAccountAvailablePayoutMethod`

remi-stripe:
Fixes the issue introduced in https://github.com/stripe/stripe-go/pull/1177

r? @cjavilla-stripe 
cc @stripe/api-libraries


cjavilla-stripe (approved):
ah nice catch LGTM

PR #1187, opened 2020-09-22
Don't retry requests on context cancellation + a few other errors

### LINKED ISSUE #1184: Prevent automatic retry on context cancellation
When making requests with a `context.Context` set in the `Params`, there's no need to retry requests if the error is due to the context being done (e.g. `context.Canceled`).

I noticed this organically while testing edge cases in a context-aware codebase, and then I found this TODO in the codebase:

https://github.com/stripe/stripe-go/blob/81f75d56592cae0f0a478bbc5f052d3f2adaaffa/stripe.go#L624-L630

I think a context check would be a relatively simple place to start.

***

For reference, here's another open-source library's approach to "retry network errors": 

https://github.com/hashicorp/go-retryablehttp/blob/6861976bd32109e69c73cb1546e7671102fb703b/client.go#L399-L445

brandur:
Currently, the retry logic is written to mostly just be smart about
status codes, like if we see a 409 or 429. We fairly naively retry any
non-Stripe error that comes back from `net/http`.

As pointed out in #1184, we should at the very least handle a context
that's been cancelled or hit its deadline (retries should stop if this
is the case), so I've done that here.

I've also tweaked some of the error logic so that we don't retry on some
types of non-Stripe errors which we know will never succeed. I don't
expect this set to be completely exhaustive, but it gives us a
framework to build on in case we want to add more errors to it in the
future. The current set is inspired by go-retryablehttp.

Fixes #1184.

r? @remi-stripe
cc @stripe/api-libraries

[1] https://github.com/hashicorp/go-retryablehttp

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -4,6 +4,7 @@ package stripe
 import (
 	"bytes"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -14,6 +15,7 @@ import (
 	"net/url"
 	"os/exec"
 	"reflect"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync"
@@ -611,6 +613,14 @@ func (s *BackendImplementation) UnmarshalJSONVerbose(statusCode int, body []byte
 	return nil
 }
 
+// Regular expressions used to match a few error types that we know we don't
+// want to retry. Unfortunately these errors aren't typed so we match on the
+// error's message.
+var (
+	redirectsErrorRE = regexp.MustCompile(`stopped after \d+ redirects\z`)
+	schemeErrorRE    = regexp.MustCompile(`unsupported protocol scheme`)
+)
+
 // Checks if an error is a problem that we should retry on. This includes both
 // socket errors that may represent an intermittent problem and some special
 // HTTP statuses.
@@ -621,11 +631,37 @@ func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *
 
 	stripeErr, _ := err.(*Error)
 
-	// TODO: This retries any non-Stripe errors produced as part of the
-	// communication process. It generally works, but there are many errors
-	// that should *not* be retried. Try to make this more granular by
-	// including only connection errors, timeout errors, etc.
+	// Don't retry if the context was was cancelled or its deadline was
+	// exceeded.
+	if req.Context() != nil && req.Context().Err() != nil {
+		return false
+	}
+
+	// We retry most errors that come out of HTTP requests except for a curated
+	// list that we know not to be retryable. This list is probably not
+	// exhaustive, so it'd be okay to add new errors to it. It'd also be okay to
+	// flip this to an inverted strategy of retrying only errors that we know
+	// to be retryable in a future refactor, if a good methodology is found for
+	// identifying that full set of errors.
 	if stripeErr == nil && err != nil {
+		if urlErr, ok := err.(*url.Error); ok {
+			// Don't retry too many redirects.
+			if redirectsErrorRE.MatchString(urlErr.Error()) {
+				return false
+			}
+
+			// Don't retry invalid protocol scheme.
+			if schemeErrorRE.MatchString(urlErr.Error()) {
+				return false
+			}
+
+			// Don't retry TLS certificate validation problems.
+			if _, ok := urlErr.Err.(x509.UnknownAuthorityError); ok {
+				return false
+			}
+		}
+
+		// Do retry every other type of non-Stripe error.
 		return true
 	}
 
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -3,10 +3,12 @@ package stripe
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"regexp"
 	"runtime"
 	"sync"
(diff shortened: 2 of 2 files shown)


stripe.go
```diff
@@ -611,6 +613,14 @@ func (s *BackendImplementation) UnmarshalJSONVerbose(statusCode int, body []byte
 	return nil
 }
 
+// Regular expressions used to match a few error types that we know we don't
+// want to retry. Unfortunately these errors aren't typed so we match on the
+// error's message.
+var (
+	redirectsErrorRE = regexp.MustCompile(`stopped after \d+ redirects\z`)
+	schemeErrorRE    = regexp.MustCompile(`unsupported protocol scheme`)
```
remi-stripe, line 621: What fires those errors exactly? I'm not sure we should ever attempt something based on an error message

stripe.go
```diff
@@ -621,11 +631,37 @@ func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *
...
 	stripeErr, _ := err.(*Error)
 
-	// TODO: This retries any non-Stripe errors produced as part of the
-	// communication process. It generally works, but there are many errors
-	// that should *not* be retried. Try to make this more granular by
-	// including only connection errors, timeout errors, etc.
+	// Don't retry if the context was was cancelled or its deadline was
+	// exceeded.
+	if req.Context() != nil && req.Context().Err() != nil {
+		return false
```
remi-stripe, line 637: Should we log more clearly here?

stripe.go
```diff
@@ -621,11 +631,37 @@ func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *
...
+	}
+
+	// We retry most errors that come out of HTTP requests except for a curated
+	// list that we know not to be retryable. This list is probably not
+	// exhautive, so it'd be okay to add new errors to it. It'd also be okay to
```
remi-stripe, line 642: exhaustive

stripe.go
brandur-stripe: They're fired from inside the `net/http` internals. String matching is definitely not great, but unless Go Core's blessed us with actual types, it's unfortunately the best we can do.

remi-stripe: But would it be the same message as Stripe's own error message or are those entirely separate?

stripe.go
brandur-stripe: Hmm, good point, but can we do this in a follow up PR? I think it might be a good idea to produce a good logging message for all nixed retries.

stripe.go
brandur-stripe: Thanks! Fixed.

stripe.go
brandur-stripe: Entirely separate — they're hard-coded from pretty deep inside the Go standard libraries. For example:

https://github.com/golang/go/blob/d140c357442471ca0d56615811fb4226e99fadd5/src/net/http/transport.go#L538

stripe.go
```diff
@@ -621,11 +631,37 @@ func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *
 
 	stripeErr, _ := err.(*Error)
 
-	// TODO: This retries any non-Stripe errors produced as part of the
-	// communication process. It generally works, but there are many errors
-	// that should *not* be retried. Try to make this more granular by
-	// including only connection errors, timeout errors, etc.
+	// Don't retry if the context was was cancelled or its deadline was
+	// exceeded.
+	if req.Context() != nil && req.Context().Err() != nil {
```
ctrudeau-stripe, line 636: req.Context() being nil or it's error not being nil means it's deadline was exceeded or cancelled. How do we know this? Can we add an end to end test (https://github.com/stripe/stripe-go/pull/1187/files#diff-1c7223ac1329b65f54265bc927d22a78R68) where the Context is explicitly cancelled?

brandur-stripe: @ctrudeau-stripe Basically going by the documentation on this one. See [context.Context](https://godoc.org/context#Context):

``` go
    // If Done is not yet closed, Err returns nil.
    // If Done is closed, Err returns a non-nil error explaining why:
    // Canceled if the context was canceled
    // or DeadlineExceeded if the context's deadline passed.
    // After Err returns a non-nil error, successive calls to Err return the same error.
    Err() error
```

Don't you think the existing test is good enough? It does basically what you requested: generates a context, explicitly cancels it, and then check that we bailed out of this function with `false`.

ctrudeau-stripe: Ah that sounds great to me.

brandur-stripe: @ctrudeau-stripe Awesome. Okay if I call this PR approved then? I'll write a follow up with more detailed logging after.

ctrudeau-stripe: :shipit: LGTM

stripe.go
```diff
@@ -621,11 +631,37 @@ func (s *BackendImplementation) shouldRetry(err error, req *http.Request, resp *
...
+
+			// Don't retry invalid protocol scheme.
+			if schemeErrorRE.MatchString(urlErr.Error()) {
+				return false
+			}
+
+			// Don't retry TLS certificate validation problems.
+			if _, ok := urlErr.Err.(x509.UnknownAuthorityError); ok {
+				return false
+			}
```
mickeyreiss, line 661: I don't know much about Stripe's HTTP frontend infrastructure, but I wonder if this case should be omitted for your use-case.

In the exceedingly rare case that Stripe rolled out a bad configuration that included bad TLS certificates, wouldn't there be a chance that the retry is routed to a healthy server, thus recovering the transaction?

mickeyreiss (commented):
Thanks for the quick fix!

stripe.go
brandur-stripe: I think it's okay to leave — it'd be a massive catastrophe if we deployed an invalid TLS certificate, so what this is more likely to protect against is users who have misconfigured their DNS or proxy, and are hitting a non-Stripe server instead of Stripe. We've seen this kind of issue once in a while in the past.

PR #1189, opened 2020-09-23
When not retrying a request, log reason at info level

brandur:
This patch follows up #1187 by having `shouldRetry` return a second
string value when no retry should be attempted which contains a short
informational message as to why there should be no retry. A caller can
use this for logging/informational purposes to make the retry system
less opaque.

r? @ctrudeau-stripe For review.

Also cc @remi-stripe who requested this feature.

cc @stripe/api-libraries

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -3,6 +3,7 @@ package stripe
 
 import (
 	"bytes"
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
@@ -471,7 +472,9 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
 
 		// If the response was okay, or an error that shouldn't be retried,
 		// we're done, and it's safe to leave the retry loop.
-		if !s.shouldRetry(err, req, res, retry) {
+		shouldRetry, noRetryReason := s.shouldRetry(err, req, res, retry)
+		if !shouldRetry {
+			s.LeveledLogger.Infof("Not retrying request: %v", noRetryReason)
 			break
 		}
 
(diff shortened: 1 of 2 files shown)


stripe_test.go
```diff
@@ -149,142 +149,188 @@ func TestShouldRetry(t *testing.T) {
 	).(*BackendImplementation)
 
 	// Exceeded maximum number of retries
-	assert.False(t, c.shouldRetry(
-		nil,
-		&http.Request{},
-		&http.Response{},
-		int(MaxNetworkRetries),
-	))
+	t.Run("DontRetryOnExceededRetries", func(t *testing.T) {
```
brandur-stripe, line 152: I also refactored this test case to use subtests, which is a pretty typical pattern in Go. This has a few advantages:

* Subtests run inside their own closure `func ... { ... }` so variables are localized to them, which means variables can't leak into other subtests and keeps things generally cleaner.
* Go tooling allows us to run any specific subtest with `-name` so we can target exactly one of them without having to run every single one.

stripe_test.go
```diff
@@ -149,142 +149,188 @@ func TestShouldRetry(t *testing.T) {
...
+		shouldRetry, _ := c.shouldRetry(
```
brandur-stripe, line 153: I'm not checking against the returned message because:

* The boolean is by far the more important value.
* Messages aren't super likely to regress in an important way.
* Not checking the messages means that it's easier for us to change one because we don't have to update the test case as well.

stripe.go
```diff
@@ -624,17 +627,28 @@ var (
...
 	if req.Context() != nil && req.Context().Err() != nil {
-		return false
+		switch req.Context().Err() {
+		case context.Canceled:
+			return false, "context canceled"
+		case context.DeadlineExceeded:
+			return false, "context deadline exceeded"
+		default:
+			return false, fmt.Sprintf("unknown context error: %v", req.Context().Err())
+		}
```
ctrudeau-stripe, line 651: This is really nice.

ctrudeau-stripe (approved):
LGTM 🚢 :shipit:

brandur-stripe:
Thank you Chris!

brandur-stripe:
Released as 72.6.0.

PR #1197, opened 2020-10-04
Update version to v72

msamoylov:
Fix the outated documentation


remi-stripe (approved):

remi-stripe:
Nice catch thank you for the fix @msamoylov

brandur-stripe:
Oops, thanks. I just updated our internal script so that this won't happen again.

PR #1200, opened 2020-10-06
handle idempotence key randomness error

KlotzAndrew:
reading from rand can return an error and fail to read bytes, this
leads to the second component not being random.

This solution panics in that case, similar to getting a new uuid
https://github.com/google/uuid/blob/master/version4.go#L13-L15

the side effect is, in the case the second component is not random (an
error happened) we are now panicing instead of continuing

alternatives: 
- continue as before (maybe leave a comment)
- change the function signature to return an error, more invasive 
change and makes the function API less nice to work with
 

Signed-off-by: Andrew Klotz <agc.klotz@gmail.com>

### DIFF
--- a/params.go
+++ b/params.go
@@ -268,7 +268,9 @@ type RangeQueryParams struct {
 func NewIdempotencyKey() string {
 	now := time.Now().UnixNano()
 	buf := make([]byte, 4)
-	rand.Read(buf)
+	if _, err := rand.Read(buf); err != nil {
+		panic(err)
+	}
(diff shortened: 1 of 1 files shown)


brandur-stripe:
This seems fine to me @KlotzAndrew. Thanks! In terms of alternatives: I think this is an unusual enough situation that just panicking is okay — technically `fmt.Printf` returns an error too, but no one bothers to rescue it.

Would you mind signing the CLA? After that, we should be good to merge.

KlotzAndrew:
signed!

brandur-stripe:
Thanks!

brandur-stripe:
Released as 72.11.0.

PR #1213, opened 2020-10-21
Add the deprecation note to the total_count field

ethervoid:
total_count property is not available anymore becuse it was removed in May 26th 2020


brandur-stripe:
Thanks @ethervoid!

I think it's a good idea to clean this up, but you've changed the sentence so that its entirety doesn't quite make sense:

> The Count property is only populated if the total_count is a deprecated option that doesn't work anymore.

Going to close this out and push a tweak on a different PR.

brandur-stripe:
Opened #1214.

PR #1217, opened 2020-10-23
Add support for passing `Bank` for P24 on `PaymentIntent` or `PaymentMethod`

remi-stripe:
r? @richardm-stripe 
cc @stripe/api-libraries

### DIFF
--- a/paymentmethod.go
+++ b/paymentmethod.go
@@ -158,6 +158,7 @@ type PaymentMethodOXXOParams struct {
 // PaymentMethodP24Params is the set of parameters allowed for the `p24` hash when creating a
 // PaymentMethod of type P24.
 type PaymentMethodP24Params struct {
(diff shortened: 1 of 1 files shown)


paymentmethod.go
```diff
@@ -329,6 +330,7 @@ type PaymentMethodOXXO struct {
 
 // PaymentMethodP24 represents the P24 properties.
 type PaymentMethodP24 struct {
+	Bank string `json:"bank"`
```
remi-stripe, line 333: Technically `bank` is an enum so it should have a list of constants but we haven't done that for other `bank` properties on `PaymentMethod` yet so I want to stay consistent (easier for code examples and such) and we can fix with codegen later (or manually if needed).

richardm-stripe (approved):

PR #1227, opened 2020-11-19
Add support for Account and Person Tokens creation

### LINKED ISSUE #1226: Provide account token creation function
I'm trying to create a new account using the Stripe API but I need to create a new Account Token but this method is not implemented in the Go client (the only client without this implementation 😶).

It would be simpler to have it inside the client and create account tokens like Bank Account Token (for example) instead to send a POST request to get it.

Is this already planned to be implemented or should I do a PR?

Jerome1337:
Fixes #1226

Tested using Golang Testing tool and with my own project

Log of my project, I correctly get my account token
```sh
2020/11/19 12:21:43 token: &{{0xc000108ba0} <nil> <nil> <nil> 1.2.3.4 1605788503  ct_1HpC8xCMhQMU3AqAy65DO2ku false account false}
2020/11/19 12:21:43 token.ID: ct_1HpC8xCMhQMU3AqAy65DO2ku
2020/11/19 12:21:43 token.TYPE: account
```

### DIFF
--- a/token.go
+++ b/token.go
@@ -17,14 +17,28 @@ type TokenCVCUpdateParams struct {
 	CVC *string `form:"cvc"`
 }
 
+// TokenAccountParams is the set of parameters that can be used when creating a Account token.
+type TokenAccountParams struct {
+	BusinessType        *string               `form:"business_type"`
+	Company             *AccountCompanyParams `form:"company"`
+	Individual          *PersonParams         `form:"individual"`
+	TOSShownAndAccepted *bool                 `form:"tos_shown_and_accepted"`
+}
+
 // TokenParams is the set of parameters that can be used when creating a token.
-// For more details see https://stripe.com/docs/api#create_card_token and https://stripe.com/docs/api#create_bank_account_token.
+// For more details see:
+// - https://stripe.com/docs/api#create_card_token
+// - https://stripe.com/docs/api#create_bank_account_token
+// - https://stripe.com/docs/api/tokens/create_account
+// - https://stripe.com/docs/api/tokens/create_person
 type TokenParams struct {
 	Params      `form:"*"`
+	Account     *TokenAccountParams   `form:"account"`
 	BankAccount *BankAccountParams    `form:"bank_account"`
 	Card        *CardParams           `form:"card"`
 	Customer    *string               `form:"customer"`
 	CVCUpdate   *TokenCVCUpdateParams `form:"cvc_update"`
+	Person      *PersonParams         `form:"person"`
 
 	// Email is an undocumented parameter used by Stripe Checkout
 	// It may be removed from the API without notice.
--- a/token/client_test.go
+++ b/token/client_test.go
@@ -60,3 +60,31 @@ func TestTokenNew_SharedCustomerCard(t *testing.T) {
 	assert.Nil(t, err)
 	assert.NotNil(t, token)
 }
+
+func TestTokenNew_WithAccount(t *testing.T) {
+	token, err := New(&stripe.TokenParams{
+		Account: &stripe.TokenAccountParams{
+			Individual: &stripe.PersonParams{
+				FirstName: stripe.String("Jane"),
+				LastName:  stripe.String("Doe"),
+			},
+			TOSShownAndAccepted: stripe.Bool(true),
+		},
+	})
+	assert.Nil(t, err)
+	assert.NotNil(t, token)
+}
+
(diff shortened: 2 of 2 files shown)


token.go
```diff
@@ -40,6 +52,7 @@ type Token struct {
 
 	BankAccount *BankAccount `json:"bank_account"`
 	Card        *Card        `json:"card"`
+	Account     *Account     `json:"account"`
```
remi-stripe, line 55: This is not something that exists in the API. While it's a Token you create about an account, it doesn't have an `account` property right now. Even if it did, it'd be a hash and not an `Account` because this is to collect PII client-side and then you add this to an account later/server-side.

remi-stripe:
@Jerome1337 Thanks for opening up the PR and doing the work here, we really appreciate it I left one blocking comment!

Would you also be willing to also do the Person Token creation since they are similar? Docs: https://stripe.com/docs/api/tokens/create_person

Also, the reason we didn't add this is that Account and Person Tokens should never be created server-side, they should always be created client-side in the browser. They were introduced specifically to collect PII securely client-side (and are required for platforms in France) but if you have the PII server-side you're better off using the Account/Person APIs directly.

Jerome1337:
@remi-stripe I added the Person Token creation and remove the wrong `Account` field inside `Token` struct.

> Also, the reason we didn't add this is that Account and Person Tokens should never be created server-side, they should always be created client-side in the browser. They were introduced specifically to collect PII securely client-side (and are required for platforms in France) but if you have the PII server-side you're better off using the Account/Person APIs directly.

Oh thanks for the advice, I will rework this part of my project then!

token.go
```diff
@@ -17,14 +17,40 @@ type TokenCVCUpdateParams struct {
...
+// TokenRelationshipParams is the set of Relationship parameters that can be used inside TokenPersonParams.
+type TokenRelationshipParams struct {
+	Owner *bool `form:"owner"`
+}
+
+// TokenPersonParams is the set of parameters that can be used when creating a Person token.
+type TokenPersonParams struct {
+	FirstName    *string                  `form:"first_name"`
+	LastName     *string                  `form:"last_name"`
+	Relationship *TokenRelationshipParams `form:"relationship"`
```
remi-stripe, line 37: This has only 3 parameters but there are many more in the documentation: https://stripe.com/docs/api/tokens/create_person

Can you clarify why you choose only those 3? We should support all of them (or none) but I know it's a lot of work so I'm not requiring you to do it yourself.

The easiest (though not cleanest) would be to use `PersonParams` instead so removing those classes and replacing
```
Person      *TokenPersonParams    `form:"person"`
```
with 
```
Person      *PersonParams    `form:"person"`
```

remi-stripe (changes requested):

token.go
Jerome1337: > Can you clarify why you choose only those 3? We should support all of them (or none)

I used this custom struct because I thought they were only 3 usable fields (those mentioned inside the documentation example).

It's a bad documentation reading 😆

> The easiest (though not cleanest) would be to use PersonParams instead so removing those classes and replacing

It's funny because I used this struct before creating the custom one.

> but I know it's a lot of work so I'm not requiring you to do it yourself.

Yeah, it's not a problem, I like contributing!
I will take a look tomorrow 😄

Jerome1337: @remi-stripe It should be ok!

token.go
```diff
@@ -17,14 +17,28 @@ type TokenCVCUpdateParams struct {
...
+// - https://stripe.com/docs/api/tokens/create_account
+// - https://stripe.com/docs/api/tokens/create_person
 type TokenParams struct {
 	Params      `form:"*"`
 	BankAccount *BankAccountParams    `form:"bank_account"`
 	Card        *CardParams           `form:"card"`
 	Customer    *string               `form:"customer"`
 	CVCUpdate   *TokenCVCUpdateParams `form:"cvc_update"`
+	Account     *TokenAccountParams   `form:"account"`
+	Person      *PersonParams         `form:"person"`
```
remi-stripe, line 41: Can you alphabetize that list please?

Jerome1337: @remi-stripe done

remi-stripe (approved):
This looks perfect. Thank you again for the doing the work here and being willing to add a bit more work into it :)

remi-stripe:
This was released as 72.26.0, thanks again @Jerome1337!

PR #1228, opened 2020-11-19
Redact client_secret from logs

vcheung-stripe:
cc @stripe/api-libraries @richardm-stripe 

Redacts API keys in logs. The redaction is done on the client as opposed to the server.

Added a unit test. All tests pass.

### DIFF
--- a/error.go
+++ b/error.go
@@ -295,8 +295,12 @@ func (e *RateLimitError) Error() string {
 }
 
 // redact returns a copy of the error object with sensitive fields replaced with
-// a placeholder value
+// a placeholder value.
 func (e *Error) redact() *Error {
+	// Fast path, since this applies to most cases
+	if e.PaymentIntent == nil && e.SetupIntent == nil {
+		return e
+	}
 	errCopy := *e
 	if e.PaymentIntent != nil {
 		pi := *e.PaymentIntent
--- a/error.go
+++ b/error.go
@@ -294,6 +294,20 @@ func (e *RateLimitError) Error() string {
 	return e.stripeErr.Error()
 }
 
+// redact returns a copy of the error object with sensitive fields replaced with
+// a placeholder value
+func (e *Error) redact() *Error {
+	errCopy := *e
+	if e.PaymentIntent != nil {
+		pi := *e.PaymentIntent
+		errCopy.PaymentIntent = &pi
+		errCopy.PaymentIntent.ClientSecret = "REDACTED"
+		return &errCopy
+	} else {
+		return e
+	}
+}
+
 // rawError deserializes the outer JSON object returned in an error response
 // from the API.
 type rawError struct {
--- a/error.go
+++ b/error.go
@@ -302,10 +302,13 @@ func (e *Error) redact() *Error {
 		pi := *e.PaymentIntent
 		errCopy.PaymentIntent = &pi
 		errCopy.PaymentIntent.ClientSecret = "REDACTED"
-		return &errCopy
-	} else {
-		return e
 	}
+	if e.SetupIntent != nil {
+		si := *e.SetupIntent
+		errCopy.SetupIntent = &si
+		errCopy.SetupIntent.ClientSecret = "REDACTED"
+	}
+	return &errCopy
 }
 
 // rawError deserializes the outer JSON object returned in an error response
--- a/error_test.go
+++ b/error_test.go
@@ -65,7 +65,7 @@ func TestErrorRedact(t *testing.T) {
 		assert.Nil(t, redacted.PaymentIntent)
 	})
 
-	t.Run("PaymentIntentAlonee", func(t *testing.T) {
+	t.Run("PaymentIntentAlone", func(t *testing.T) {
 		err := &Error{PaymentIntent: pi}
 		redacted := err.redact()
 		assert.Equal(t, int64(400), err.PaymentIntent.Amount)
--- a/error_test.go
+++ b/error_test.go
@@ -43,8 +43,7 @@ func TestErrorRedact(t *testing.T) {
 	pi := &PaymentIntent{Amount: int64(400), ClientSecret: "foo"}
 	si := &SetupIntent{Description: "keepme", ClientSecret: "foo"}
 
-	// Both
-	{
+	t.Run("BothIntentObjects", func(t *testing.T) {
 		err := &Error{PaymentIntent: pi, SetupIntent: si}
 		redacted := err.redact()
 		assert.Equal(t, int64(400), err.PaymentIntent.Amount)
@@ -57,18 +56,16 @@ func TestErrorRedact(t *testing.T) {
 		assert.Equal(t, "foo", si.ClientSecret)
 		assert.Equal(t, "REDACTED", redacted.PaymentIntent.ClientSecret)
 		assert.Equal(t, "REDACTED", redacted.SetupIntent.ClientSecret)
-	}
+	})
 
-	// Neither
-	{
+	t.Run("NeitherIntentObject", func(t *testing.T) {
 		err := Error{PaymentIntent: nil, SetupIntent: nil}
 		redacted := err.redact()
 		assert.Nil(t, err.PaymentIntent)
 		assert.Nil(t, redacted.PaymentIntent)
-	}
+	})
 
-	// Just PaymentIntent
-	{
+	t.Run("PaymentIntentAlonee", func(t *testing.T) {
 		err := &Error{PaymentIntent: pi}
 		redacted := err.redact()
 		assert.Equal(t, int64(400), err.PaymentIntent.Amount)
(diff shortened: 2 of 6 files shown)


log.go
```diff
@@ -76,33 +77,60 @@ type LeveledLogger struct {
 	stdoutOverride io.Writer
 }
 
+var apiKeyRegex *regexp.Regexp
```
brandur-stripe, line 80: Try just:

``` go
var apiKeyRegex *regexp.Regexp = regexp.MustCompile(...)
```

Also, mind using either `Regexp` or `RE` as a suffix? I don't think we should mix the abbreviations "regex" and "regexp" which are very similar but different.

log.go
```diff
@@ -76,33 +77,60 @@ type LeveledLogger struct {
...
+		return redactMiddle(str, 10, 4)
+	})
+}
+
+func (l *LeveledLogger) printLog(format string, tag string, level Level, output io.Writer, v ...interface{}) {
+	if l.Level >= level {
+		s := fmt.Sprintf(format, v...)
+		fmt.Fprintf(output, "[%s] %s\n", tag, redact(s))
+	}
+}
```
brandur-stripe, line 114: I've tried to keep files like this organized so that you have the public-facing types/functions nearer to the top, and the internal ones below (see comments like "Public types". Can we move all of this down to a new section at the bottom (except `printLog` which belongs on `LeveledLogger`, but try to order it alphabetically so it appears below `Warnf` and above `stderr`.

brandur-stripe (commented):
Left a couple comments, but in general, not super crazy about this one! We make logging slower and there's other secrets beyond keys that could be logged too.

And on that note: are we even sure that the original ticket that this is what's needed for the fix? It looked like the field in question was `client_secret`, which by my read of the docs, refers to a secret from something like a payment intent, setup intent, or source. It should still not be logged, but it looks more like `pi_1DijHt2eZvKYlo2C0VCV90Tf_secret_24N2Yr44VXRpt5Np2mTIWgRDX`.

ptal @vcheung-stripe

vcheung-stripe:
> And on that note: are we even sure that the original ticket that this is what's needed for the fix? It looked like the field in question was `client_secret`, which by my read of the docs, refers to a secret from something like a payment intent, setup intent, or source. It should still not be logged, but it looks more like `pi_1DijHt2eZvKYlo2C0VCV90Tf_secret_24N2Yr44VXRpt5Np2mTIWgRDX`.

This is a mistake, I completely misread `client_secret` as an API key 🤦. Thank you for catching this.

> We make logging slower and there's other secrets beyond keys that could be logged too.

This is a concern I have as well. One approach I can think of is we could redact `client_secret` when we receive an error response, right after we deserialize a response body. This way we don't do the redaction unless there is an error, and we do this for all errors. Or do you have a better recommendation for how to go about this?

@brandur-stripe

log.go
```diff
@@ -76,33 +77,60 @@ type LeveledLogger struct {
 	stdoutOverride io.Writer
 }
 
+var apiKeyRegex *regexp.Regexp
+
+func init() {
+	var err error
+	apiKeyRegex, err = regexp.Compile("(sk_live|sk_test|rk_live|rk_test)_[a-zA-Z0-9]{2}[a-zA-Z0-9]{1,}")
```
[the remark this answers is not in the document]
remi-stripe: Lurking in: Historical API keys did not have `sk_` prefix or `live_` in them so that would not catch all of them.

I'm also curious why we're adding this. The internal ask I'm aware of is specifically about `client_secret` which is a property on PaymentIntent and SetupIntent, is this related or orthogonal?

brandur-stripe:
> This is a concern I have as well. One approach I can think of is we could redact `client_secret` when we receive an error response, right after we deserialize a response body. This way we don't do the redaction unless there is an error, and we do this for all errors. Or do you have a better recommendation for how to go about this?

I'd like something like that better. Basically have a function that we only use when dumping JSON blobs out to logs which will recursively spalunk the structure and redact knowingly problematic fields in it. That way, we wouldn't have to run it on every single logging statement either.

We could also consider something much simpler like just not dumping entire JSON structs out in logs anymore. I don't think we do that for most languages, and users now have access to the raw response, so if they wanted, they could dump it out for themselves and take the various risks inherent in doing so.

richardm-stripe:
Closing this particular PR, since it sounds like we're going to take a different approach altogether.

richardm-stripe:
r? @brandur-stripe take a look at this version? This just redacts client-secret. The tricky bit is that I'm afraid of mutating the error object and not sure how/whether to make a deep copy.

I think if there are more sensitive fields it might make sense to use struct annotations and `reflect`. That would certainly feel cleaner than this.

I'm interested in your suggestion of dropping logging JSON structs, but I do think it is pretty nice to have? It's true the user has access to the raw response, but even so `Level: LevelInfo` provides an easy way to toggle the logs in a single place, and if we removed it the user might have to either add logging statements in multiple places or wrap the stripe client.

log.go
```diff
@@ -140,3 +131,14 @@ type LeveledLoggerInterface interface {
...
 }
+
+//
+// Private functions
+//
+
+func (l *LeveledLogger) printLog(format string, tag string, level Level, output io.Writer, v ...interface{}) {
+	if l.Level >= level {
+		s := fmt.Sprintf(format, v...)
+		fmt.Fprintf(output, "[%s] %s\n", tag, s)
```
brandur-stripe, line 142: What do you think about collapsing these two print calls into just one. i.e.

``` go
fmt.Printf(output, "[%s] " + format + "\n", tag v...)
```

(Not sure that exact line works, but something like it should.)

I gotta say too that I'm generally a fan of extracting things into small functions, but I'm not sure that we're getting our money's worth here. The previous standard of 3 lines per error/warn/info was really not that bad, and we've made this incrementally more difficult to understand in that `if l.Level >= level` is harder to read than `if l.Level >= LevelWarn {`.

error.go
```diff
@@ -294,6 +294,23 @@ func (e *RateLimitError) Error() string {
 	return e.stripeErr.Error()
 }
 
+// redact returns a copy of the error object with sensitive fields replaced with
+// a placeholder value
```
brandur-stripe, line 298: Super nit, but mind punctuating here (it's pretty common throughout the library now).

error_test.go
```diff
@@ -38,3 +38,55 @@ func TestErrorResponse(t *testing.T) {
...
+	{
+		err := &Error{SetupIntent: si}
+		redacted := err.redact()
+		assert.Equal(t, "keepme", err.SetupIntent.Description)
+		assert.Equal(t, "keepme", redacted.SetupIntent.Description)
+		assert.Equal(t, "foo", err.SetupIntent.ClientSecret)
+		assert.Equal(t, "foo", si.ClientSecret)
+		assert.Equal(t, "REDACTED", redacted.SetupIntent.ClientSecret)
+	}
+
```
brandur-stripe, line 91: Another nit, but could we kill this extra newline?

error_test.go
```diff
@@ -38,3 +38,55 @@ func TestErrorResponse(t *testing.T) {
 	assert.Equal(t, "req_123", stripeErr.RequestID)
 	assert.Equal(t, 401, stripeErr.HTTPStatusCode)
 }
+
+func TestErrorRedact(t *testing.T) {
+	pi := &PaymentIntent{Amount: int64(400), ClientSecret: "foo"}
+	si := &SetupIntent{Description: "keepme", ClientSecret: "foo"}
+
+	// Both
+	{
```
brandur-stripe, line 47: This isn't quite a standard yet because I thought about it too late, but consider changing these blocks to subtests like:

``` go
t.Run("BothIntentObjects", func(t *testing.T) {
...
```

This is a little nicer because it means that you can now target one specific subtest to run with Go test:

``` sh
go test . -run BothIntentObjects
```

error.go
```diff
@@ -294,6 +294,23 @@ func (e *RateLimitError) Error() string {
...
+func (e *Error) redact() *Error {
```
brandur-stripe, line 299: Not a big deal, but could consider a fast path here like:

``` go
if e.PaymentIntent == nil && e.SetupIntent == nil {
    return e
}
```

(Since that's probably like 95%+ of calls to this function.)

brandur-stripe (commented):
Nice! Yeah, this works for me and shouldn't add too much overhead. Left a couple minor comments.

ptal @richardm-stripe

error_test.go
```diff
@@ -38,3 +38,49 @@ func TestErrorResponse(t *testing.T) {
...
+	})
+
+	t.Run("NeitherIntentObject", func(t *testing.T) {
+		err := Error{PaymentIntent: nil, SetupIntent: nil}
+		redacted := err.redact()
+		assert.Nil(t, err.PaymentIntent)
+		assert.Nil(t, redacted.PaymentIntent)
+	})
+
+	t.Run("PaymentIntentAlonee", func(t *testing.T) {
```
brandur-stripe, line 68: Thanks Richard!

Tiny typpo "Alonee" here. Otherwise LGTM!

ptal @richardm-stripe

richardm-stripe, line 68:
```suggestion
	t.Run("PaymentIntentAlone", func(t *testing.T) {
```

error_test.go
```diff
@@ -38,3 +38,49 @@ func TestErrorResponse(t *testing.T) {
...
+
+	t.Run("SetupIntentAlone", func(t *testing.T) {
+		err := &Error{SetupIntent: si}
+		redacted := err.redact()
+		assert.Equal(t, "keepme", err.SetupIntent.Description)
+		assert.Equal(t, "keepme", redacted.SetupIntent.Description)
+		assert.Equal(t, "foo", err.SetupIntent.ClientSecret)
+		assert.Equal(t, "foo", si.ClientSecret)
+		assert.Equal(t, "REDACTED", redacted.SetupIntent.ClientSecret)
+	})
```
brandur-stripe, line 86: Looks like also need one more ending curly brace down here.

richardm-stripe, line 86:
```suggestion
	})
}
```

PR #1236, opened 2020-12-11
Add support for bank on eps/p24

richardm-stripe:
r? @brandur-stripe 

Add support for `Bank` for eps and p24 on Charge and PaymentMethod.
Add support for `TOSShownAndAccepted` on `PaymentMethodP24Params`

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -172,10 +172,12 @@ type PaymentIntentPaymentMethodDataParams struct {
 	AUBECSDebit    *PaymentMethodAUBECSDebitParams `form:"au_becs_debit"`
 	BillingDetails *BillingDetailsParams           `form:"billing_details"`
 	Card           *PaymentMethodCardParams        `form:"card"`
+	EPS            *PaymentMethodEPSParams         `form:"eps"`
 	FPX            *PaymentMethodFPXParams         `form:"fpx"`
 	Grabpay        *PaymentMethodGrabpayParams     `form:"grabpay"`
 	Ideal          *PaymentMethodIdealParams       `form:"ideal"`
 	OXXO           *PaymentMethodOXXOParams        `form:"oxxo"`
+	P24            *PaymentMethodP24Params         `form:"p24"`
 	SepaDebit      *PaymentMethodSepaDebitParams   `form:"sepa_debit"`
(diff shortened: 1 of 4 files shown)


paymentmethod.go
```diff
@@ -164,7 +165,8 @@ type PaymentMethodOXXOParams struct {
 // PaymentMethodP24Params is the set of parameters allowed for the `p24` hash when creating a
 // PaymentMethod of type P24.
 type PaymentMethodP24Params struct {
-	Bank *string `form:"bank"`
+	Bank                *string `form:"bank"`
+	TOSShownAndAccepted *bool   `from:"tos_shown_and_accepted"`
```
richardm-stripe, line 169:
```suggestion
	TOSShownAndAccepted *bool   `form:"tos_shown_and_accepted"`
```

paymentintent.go
```diff
@@ -172,10 +172,12 @@ type PaymentIntentPaymentMethodDataParams struct {
 	AUBECSDebit    *PaymentMethodAUBECSDebitParams `form:"au_becs_debit"`
 	BillingDetails *BillingDetailsParams           `form:"billing_details"`
 	Card           *PaymentMethodCardParams        `form:"card"`
+	EPS            *PaymentMethodEPSParams         `form:"eps"`
```
brandur-stripe, line 175: Arg, it looks like this should be "EPS", but it's called `Eps` on `Charge`. We should probably keep it the same between the two just for the sake of consistency.

brandur-stripe (commented):
Thanks Richard. Left one comment, but LGTM.

brandur-stripe:
ptal @richardm-stripe

paymentintent.go
richardm-stripe: It's Eps on Charge, but unfortunately it is also EPS in paymentmethod.go and account.go and I think it is probably better to be consistent with that?
ptal @brandur

brandur-stripe: > It's Eps on Charge, but unfortunately it is also EPS in paymentmethod.go and account.go and I think it is probably better to be consistent with that?

Ah, fair. Yep, let's go with `EPS` then.

brandur-stripe (approved):

PR #1241, opened 2021-01-14
Multiple API Changes

richardm-stripe:
r? @brandur-stripe 
Contains changes corresponding to [openapi releases](https://github.com/stripe/openapi/releases) v11 and v12.

## Changelog
* Added support for `dynamic_tax_rates` on `CheckoutSessionParams.line_items`
* Added support for `customer_details` on `CheckoutSession`
* Added support for `type` on `IssuingTransactionListParams`
* Added support for `country` and `state` on `TaxRateParams` and `TaxRate`

### DIFF
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -252,7 +252,7 @@ type CheckoutSessionCustomerDetailsTaxIDs struct {
 type CheckoutSessionCustomerDetails struct {
 	Email     string                                  `json:"email"`
 	TaxExempt CheckoutSessionCustomerDetailsTaxExempt `json:"tax_exempt"`
-	TaxIDs    []CheckoutSessionCustomerDetailsTaxIDs  `json:"tax_ids"`
+	TaxIDs    []*CheckoutSessionCustomerDetailsTaxIDs `json:"tax_ids"`
 }
 
 // CheckoutSessionShippingAddressCollection is the set of parameters allowed for the
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -4,45 +4,45 @@ import (
 	"encoding/json"
 )
 
-// CheckoutSessionCustomerDetailsTaxIdsType is the list of allowed values for type
+// CheckoutSessionCustomerDetailsTaxIDsType is the list of allowed values for type
 // on the tax_ids inside customer_details of a checkout session.
-type CheckoutSessionCustomerDetailsTaxIdsType string
+type CheckoutSessionCustomerDetailsTaxIDsType string
 
-// List of values that CheckoutSessionCustomerDetailsTaxIdsType can take.
+// List of values that CheckoutSessionCustomerDetailsTaxIDsType can take.
 const (
-	CheckoutSessionCustomerDetailsTaxIdsTypeAETRN   CheckoutSessionCustomerDetailsTaxIdsType = "ae_trn"
-	CheckoutSessionCustomerDetailsTaxIdsTypeAUABN   CheckoutSessionCustomerDetailsTaxIdsType = "au_abn"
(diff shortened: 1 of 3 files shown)


checkout_session.go
```diff
@@ -187,6 +240,21 @@ type CheckoutSessionListParams struct {
...
+	Type  CheckoutSessionCustomerDetailsTaxIdsType `json:"type"`
+	Value string                                   `json:"value"`
+}
+
+// CheckoutSessionCustomerDetails represent the customer details including
+// the tax exempt status and the customer's tax IDs.
+type CheckoutSessionCustomerDetails struct {
+	Email     string                                  `json:"email"`
+	TaxExempt CheckoutSessionCustomerDetailsTaxExempt `json:"tax_exempt"`
+	TaxIds    *[]CheckoutSessionCustomerDetailsTaxIds `json:"tax_ids"`
```
richardm-stripe, line 255: Not sure about this. Need to double check that this should be a pointer

checkout_session.go
```diff
@@ -4,6 +4,58 @@ import (
 	"encoding/json"
 )
 
+// CheckoutSessionCustomerDetailsTaxIdsType is the list of allowed values for type
+// on the tax_ids inside customer_details of a checkout session.
+type CheckoutSessionCustomerDetailsTaxIdsType string
```
brandur-stripe, line 9: Any chance this could be `CheckoutSessionCustomerDetailsTaxIDsType`? (And similar for other other names in here.)

checkout_session.go
brandur-stripe: Nah, I'd make it a non-pointer. All slices are pointers already in the way that they can hold a `nil` value (which behaves like an empty list).

checkout_session.go
```diff
@@ -187,6 +240,21 @@ type CheckoutSessionListParams struct {
...
+	Type  CheckoutSessionCustomerDetailsTaxIDsType `json:"type"`
+	Value string                                   `json:"value"`
+}
+
+// CheckoutSessionCustomerDetails represent the customer details including
+// the tax exempt status and the customer's tax IDs.
+type CheckoutSessionCustomerDetails struct {
+	Email     string                                  `json:"email"`
+	TaxExempt CheckoutSessionCustomerDetailsTaxExempt `json:"tax_exempt"`
+	TaxIDs    []CheckoutSessionCustomerDetailsTaxIDs  `json:"tax_ids"`
```
remi-stripe, line 255: This should be `[]*CheckoutSessionCustomerDetailsTaxIDs` right?

issuing_transaction.go
```diff
@@ -40,10 +50,11 @@ type IssuingTransactionParams struct {
...
 	ListParams   `form:"*"`
-	Card         *string           `form:"card"`
-	Cardholder   *string           `form:"cardholder"`
-	Created      *int64            `form:"created"`
-	CreatedRange *RangeQueryParams `form:"created"`
+	Card         *string                     `form:"card"`
+	Cardholder   *string                     `form:"cardholder"`
+	Created      *int64                      `form:"created"`
+	CreatedRange *RangeQueryParams           `form:"created"`
+	Type         *IssuingTransactionListType `form:"type"`
```
remi-stripe, line 57: We don't use enums in params, this should be `*string` instead and we don't need the type. Also if we used the type we would use `IssuingTransactionType` instead

checkout_session.go
richardm-stripe: Good catch! Fixed.

issuing_transaction.go
richardm-stripe: Woot! Great to know.

remi-stripe (approved):
Change lgtm

PR #1246, opened 2021-02-02
Add SubscriptionStartDate to InvoiceParams (to use with GetNext)

ees37:
The Stripe API supports passing a `subscription_start_date` to the `invoices/upcoming` endpoint. This is missing in the `InvoiceParams` struct.


remi-stripe (approved):
Thanks for doing the work @ees37 !
I'm re-assigning to @richardm-stripe to take the release in the next version

PR #1248, opened 2021-02-13
CI: github actions

richardm-stripe:
r? @ob-stripe 
Introduces github actions for CI.

- skips goveralls
- skips testing on "tip" which isn't supported (yet) by the setup-go action: https://github.com/actions/setup-go/issues/21
  - Unfortunately GHA seems to lack an exact analog to "allowed failures" in Travis, so testing on tip wouldn't really work as desired anyway.
  
Also fixes some linter errors. Travis isn't failing when golint fails, for some reason, and so there were some lingering lint problems that need to be fixed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
      strategy:
        matrix:
          go:
+           - "1.16"
            - "1.15"
            - "1.14"
            - "1.13"
@@ -80,4 +81,3 @@ jobs:
               cd $DIR/$MAJOR_VERSION &&
               go get -u "golang.org/x/net/http2" &&
               make build
-
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,83 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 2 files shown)


ob-stripe (changes requested):
Go 1.16 was released a couple weeks ago so we should start testing it.

Also it looks like you've got an unrelated file that snuck into your PR.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,83 @@
...
+          go get -u golang.org/x/lint/golint &&
+          $HOME/go/bin/golint -set_exit_status ./... &&
+          make vet
+
+  test:
+     runs-on: ubuntu-latest
+     strategy:
+       matrix:
+         go:
+           - "1.15"
```
richardm-stripe, line 29:
```suggestion
           - "1.16"
           - "1.15"
```

richardm-stripe:
r? @ob-stripe 
Added 1.16! The extra file is a real fix -- Travis isn't failing properly on the "lint" step, although I could commit that separately if you'd like.

ob-stripe (approved):
That's fine, lgtm!

PR #1249, opened 2021-02-14
Fix edge case panic in ParseID

bayandin:
This PR fixes panic in `stripe.ParseID`, which could happen for a particular input (`"`):
```
panic: runtime error: slice bounds out of range [1:0]
```

I highly doubt that this is a real-world case, but since panic is a pretty bad thing I believe it's worth to fix it.
Also added a test case for it.

Found this with the help of [go-fuzz](https://github.com/dvyukov/go-fuzz).

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -977,6 +977,11 @@ func ParseID(data []byte) (string, bool) {
 		return "", false
 	}
 
+	// Edge case that should never happen; found via fuzzing
+	if s == "\"" {
+		return "", false
+	}
+
 	return s[1 : len(s)-1], true
 }
 
--- a/stripe_test.go
+++ b/stripe_test.go
@@ -777,6 +777,13 @@ func TestParseID(t *testing.T) {
 		assert.Equal(t, "", id)
 		assert.False(t, ok)
 	}
+
+	// Edge case that should never happen; found via fuzzing
+	{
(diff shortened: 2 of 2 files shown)


remi-stripe:
@bayandin Thanks a lot for the report and taking the time to submit a fix for the issue. I agree it's worth fixing even if it's not supposed to happen. We'll have a look and get back to you if we need changes and otherwise release in the next version next week!

stripe.go
```diff
@@ -977,6 +977,10 @@ func ParseID(data []byte) (string, bool) {
 		return "", false
 	}
 
+	if s == "\"" {
```
brandur-stripe, line 980: Mind also porting the comment from below up here? Maybe something like this:

> Edge case that should never happen; found via fuzzing

brandur-stripe (changes requested):
Thanks @bayandin!

I don't think this is super necessary as we've never seen it in the wild before, and I'm pretty sure we'd fail to unmarshal JSON before we got to this point, but it can't hurt. Left one minor ask for an additional comment, but will merge after that.

stripe.go
bayandin: This totally makes sense. I've added a comment here and updated a comment in a test.

brandur-stripe:
Looks good. Thanks!

PR #1257, opened 2021-02-28
Adds ErrorType idempotency_error

merrickfox:
Hi Stripe folk 👋

**This PR adds the `ErrorType` `ErrorTypeIdempotency`**

As a consumer of your go SDK I ran into some idempotency errors and while trying to gracefully handle them on my end I noticed the error type `idempotency_error` wasn't mapped to your other `ErrorType`s

All the other main errors seem to be mapped:

Adding this as an error type will allow consumers to cleanly do stripe error type assertions in their code, eg.

```go
pi, err := stripe.somePaymentIntentThing(paymentIntentID)
if err != nil {
    if se, ok := err.(*stripe.Error); ok {
    switch se.Type {
    case stripe.ErrorTypeAPIConnection:
        // handle this kind of error
    case stripe.ErrorTypeIdempotency:
        // handle this kind of error
    case ....    
    }
} 	

```

Currently you instead have to do something like: 

```go
pi, err := stripe.somePaymentIntentThing(paymentIntentID)
if err != nil {
    if se, ok := err.(*stripe.Error); ok {
    switch se.Type {
    case stripe.ErrorTypeAPIConnection:
        // handle this kind of error
    case stripe.ErrorType("idempotency_error"):
        // handle this kind of error
    case ....
    }
} 	
```

Thanks!

### DIFF
--- a/error.go
+++ b/error.go
@@ -14,6 +14,7 @@ const (
 	ErrorTypeInvalidRequest ErrorType = "invalid_request_error"
 	ErrorTypePermission     ErrorType = "more_permissions_required"
 	ErrorTypeRateLimit      ErrorType = "rate_limit_error"
+	ErrorTypeIdempotency    ErrorType = "idempotency_error"
 )
 
 // ErrorCode is the list of allowed values for the error's code.
@@ -292,6 +293,17 @@ type RateLimitError struct {
 
 // Error serializes the error object to JSON and returns it as a string.
 func (e *RateLimitError) Error() string {
+	return e.stripeErr.Error()
+}
+
+// IdempotencyError occurs when an Idempotency-Key is re-used on a request
+// that does not match the first request's API endpoint and parameters.
+type IdempotencyError struct {
+	stripeErr *Error
+}
+
+// Error serializes the error object to JSON and returns it as a string.
+func (e *IdempotencyError) Error() string {
 	return e.stripeErr.Error()
 }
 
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -14,6 +14,7 @@ const (
 	ErrorTypeInvalidRequest ErrorType = "invalid_request_error"
 	ErrorTypePermission     ErrorType = "more_permissions_required"
 	ErrorTypeRateLimit      ErrorType = "rate_limit_error"
+	ErrorTypeIdempotency    ErrorType = "idempotency_error"
```
brandur-stripe, line 17: @merrickfox Thanks for the contribution! Very minor, would you mind just moving this up a few lines so that it's sorted alphabetically?

stripe.go
```diff
@@ -566,6 +566,8 @@ func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []by
 		typedError = &PermissionError{stripeErr: raw.Error}
 	case ErrorTypeRateLimit:
 		typedError = &RateLimitError{stripeErr: raw.Error}
+	case ErrorTypeIdempotency:
+		typedError = &IdempotencyError{stripeErr: raw.Error}
```
brandur-stripe, line 570: And similarly, could you move this case so that this list is sorted alphabetically as well? Thanks!

error.go
merrickfox: Cool, sorted both statements.

brandur-stripe:
Perfect. Thank you!

brandur-stripe:
Released as 72.37.0.

PR #1261, opened 2021-03-09
Begin codegenning 5 files

richardm-stripe:
r? @brandur-stripe 
cc @stripe/api-libraries
We are beginning to automatically generate parts of `stripe-go`. This PR adds five of the files that were simplest to support.

- Adds a comment to the top of these files.
- Adds `Object` and `Deleted` properties. These properties are present on many stripe resources and should be present, but in the past it appears we occasionally omit them as the user need for them is quite rare.
- Rearranging contents.
- variable renaming in UnmarshalJSON method. We don't really follow consistent conventions with these methods today and kind of abbreviate things arbitrarily. Codegen will have to change this.

### DIFF
--- a/coupon.go
+++ b/coupon.go
@@ -1,4 +1,8 @@
+//
+//
 // File generated from our OpenAPI spec
+//
+//
 
 package stripe
 
--- a/coupon.go
+++ b/coupon.go
@@ -1,3 +1,4 @@
+// File generated from our OpenAPI spec
 package stripe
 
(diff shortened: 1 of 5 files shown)


brandur (commented):
Very cool man! I love that codegen ends up catching a couple unintentional alphabetization errors and the like too.

Left one comment about ... comments, but LGTM otherwise.

ptal @richardm-stripe

PR #1264, opened 2021-03-12
Codegen 10 more files

richardm-stripe:
r? @remi-stripe - removed the potential breaking change, this is ready for a real review

- `applepaydomain.go`
  - Add a missing `DomainName` field to the ListParams.
  https://github.com/stripe/stripe-go/pull/1264/files#diff-cbfd212f7ddc855ea5f8d193fa576a280cc362cd76c23e619917129eaffbbdf3R18
  - Add missing `Object` field
 - `application.go`
   - Add missing `Object` field
 - `customerbalancetransaction.go`
   - Add two missing members to `CustomerBalanceTransactionType` enum
 - `feerefund.go`
   - Add missing `Object` field
 - `loginlink.go`
   - Add missing `Object` field
 - `product.go`
   - Add missing `Object` and `Deleted` field
 - `terminal_location.go`
   - ~Corrected `*AccountAddressParams` reference in resource, should have been `*AccountAddress`. https://github.com/stripe/stripe-go/pull/1264/files#diff-1cec9d2aed3b04955c7fb6b3e5c43d9146271bbbf65e010fc5cf63e331210902R24 (is this a breaking bugfix? I don't think so because this didn't have `json:` form annotations, so anybody trying to read this would already be broken)~ (postponing this until the major)
 - `usagerecord.go`
   - Added missing `Object` field

### DIFF
--- a/applepaydomain.go
+++ b/applepaydomain.go
@@ -1,8 +1,20 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
 package stripe
 
 // ApplePayDomainParams is the set of parameters that can be used when creating an ApplePayDomain object.
 type ApplePayDomainParams struct {
 	Params     `form:"*"`
+	DomainName *string `form:"domain_name"`
+}
+
+// ApplePayDomainListParams are the parameters allowed during ApplePayDomain listing.
+type ApplePayDomainListParams struct {
+	ListParams `form:"*"`
 	DomainName *string `form:"domain_name"`
 }
 
@@ -14,11 +26,7 @@ type ApplePayDomain struct {
 	DomainName string `json:"domain_name"`
 	ID         string `json:"id"`
 	Livemode   bool   `json:"livemode"`
-}
-
-// ApplePayDomainListParams are the parameters allowed during ApplePayDomain listing.
-type ApplePayDomainListParams struct {
-	ListParams `form:"*"`
+	Object     string `json:"object"`
 }
 
 // ApplePayDomainList is a list of ApplePayDomains as returned from a list endpoint.
--- a/customerbalancetransaction.go
+++ b/customerbalancetransaction.go
@@ -1,3 +1,9 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
 package stripe
 
 import "encoding/json"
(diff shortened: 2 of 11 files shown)


applepaydomain.go
```diff
@@ -6,6 +12,12 @@ type ApplePayDomainParams struct {
 	DomainName *string `form:"domain_name"`
 }
 
+// ApplePayDomainListParams are the parameters allowed during ApplePayDomain listing.
```
remi-stripe, line 15: OOC will we add real comments from the API ref like we do in other libraries in the future?

customerbalancetransaction.go
```diff
@@ -22,8 +30,8 @@ const (
 // For more details see https://stripe.com/docs/api/customers/create_customer_balance_transaction
 type CustomerBalanceTransactionParams struct {
 	Params      `form:"*"`
+	Customer    *string `form:"-"` // Included in URL
```
remi-stripe, line 33: That isn't in alphabetical anymore. Not the end of the world but is it something we can easily tweak?

terminal_location.go
```diff
@@ -15,13 +21,13 @@ type TerminalLocationListParams struct {
...
 type TerminalLocation struct {
 	APIResource
-	Address     *AccountAddressParams `json:"address"`
-	Deleted     bool                  `json:"deleted"`
-	DisplayName string                `json:"display_name"`
-	ID          string                `json:"id"`
-	Livemode    bool                  `json:"livemode"`
-	Metadata    map[string]string     `json:"metadata"`
-	Object      string                `json:"object"`
+	Address     *AccountAddress   `json:"address"`
```
remi-stripe, line 24: Is that one a breaking change?

applepaydomain.go
richardm-stripe: Yes, I want to do that as a single commit

customerbalancetransaction.go
richardm-stripe: The convention now is embeds, path params, then normal params, and alphabetical within those categories. A saw the hoisting of path params in a couple files and thought it made sense even though we don't follow it consistently, but yes I can easily alphabetize the whole thing if we think that is better.

Maybe a newline between the path params and the remainder?

terminal_location.go
richardm-stripe: It's either a breaking change or a bugfix.

AccountAddressParams right now only has `form` annotatations and not `json` annotations, so I think right now it's pretty broken and these will always be zero values.

customerbalancetransaction.go
remi-stripe: I personally prefer everyone alphabetical. What's URL param and whats POST param shouldn't really matter to developers and it's easier for people's brain (and ours) to parse alphabetically. Not a blocker though.

remi-stripe (approved):

PR #1269, opened 2021-03-22
codegen: 13 more files

richardm-stripe:
r? @remi-stripe 

Contains changes for 13 files to be compatible with codegenned output, bringing the total to 28/74 resource files (`client.go` files and index files are still outstanding).

None of these changes should be breaking. See conversation about `string -> enum` being the non-breaking addition of a type alias.

* accountlink.go - noop
* billingportal_session.go - noop
* countryspec.go
  * Add support for `Object` to `CountrySpec`
* filelink.go
  * Add support `ExpiresAtNow` to `FileLinkParams`
* invoiceitem.go
  * Add support for `Object` and `SubscriptionItem` to `InvoiceItem`
* issuing_transaction.go - noop
* radar_earlyfraudwarning.go - noop
* rader_valuelistitem.go - noop
* reporting_reporttype.go - noop
* subitem.go
  * Add support for `Object` to `SubscriptionItem`
* taxid.go - noop
* terminal_reader.go
  * Add enum definitions for `TerminalReader.DeviceType`
* topup.go
  * Add enum definitions for `Topup.Status`
  * Add support for `Amount`, `AmountRange`, and `Status` to `TopupListParams`.
  * Added custom `UnmarshalJSON` method for `Topup`

Re: UnmarshalJSON for `Topup` -- Codegen is calculating that this is needed because according to the API definition it looks like `Topup` can be returned in unexpanded form on BalanceTransaction.Source, but it looks like [support for that is incomplete](https://github.com/stripe/stripe-go/blob/master/balancetransaction.go#L82) in balancetransaction.go.

### DIFF
--- a/filelink.go
+++ b/filelink.go
@@ -20,7 +20,7 @@ type FileLinkParams struct {
 // AppendTo implements custom encoding logic for FileLinkParams.
 func (f *FileLinkParams) AppendTo(body *form.Values, keyParts []string) {
 	if BoolValue(f.ExpiresAtNow) {
-		body.Add(form.FormatKey(append(keyParts, "ExpiresAt")), "now")
+		body.Add(form.FormatKey(append(keyParts, "expires_at")), "now")
 	}
 }
 
--- a/filelink.go
+++ b/filelink.go
@@ -1,14 +1,27 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
 package stripe
 
-import (
-	"encoding/json"
-)
+import "encoding/json"
+import "github.com/stripe/stripe-go/v72/form"
 
 // FileLinkParams is the set of parameters that can be used when creating or updating a file link.
 type FileLinkParams struct {
-	Params    `form:"*"`
-	ExpiresAt *int64  `form:"expires_at"`
-	File      *string `form:"file"`
+	Params       `form:"*"`
+	ExpiresAt    *int64  `form:"expires_at"`
+	ExpiresAtNow *bool   `form:"-"` // See custom AppendTo
+	File         *string `form:"file"`
+}
+
+// AppendTo implements custom encoding logic for FileLinkParams.
+func (f *FileLinkParams) AppendTo(body *form.Values, keyParts []string) {
+	if BoolValue(f.ExpiresAtNow) {
+		body.Add(form.FormatKey(append(keyParts, "ExpiresAt")), "now")
+	}
 }
 
 // FileLinkListParams is the set of parameters that can be used when listing file links.
@@ -35,12 +48,12 @@ type FileLink struct {
 	URL       string            `json:"url"`
 }
 
-// UnmarshalJSON handles deserialization of a file link.
+// UnmarshalJSON handles deserialization of a FileLink.
 // This custom unmarshaling is needed because the resulting
-// property may be an ID or the full struct if it was expanded.
-func (c *FileLink) UnmarshalJSON(data []byte) error {
+// property may be an id or the full struct if it was expanded.
+func (f *FileLink) UnmarshalJSON(data []byte) error {
 	if id, ok := ParseID(data); ok {
-		c.ID = id
+		f.ID = id
 		return nil
 	}
 
@@ -50,7 +63,7 @@ func (c *FileLink) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
-	*c = FileLink(v)
+	*f = FileLink(v)
 	return nil
 }
 
--- a/subitem.go
+++ b/subitem.go
@@ -52,11 +52,8 @@ type SubscriptionItemBillingThresholdsParams struct {
 // SubscriptionItemListParams is the set of parameters that can be used when listing invoice items.
 // For more details see https://stripe.com/docs/api#list_invoiceitems.
 type SubscriptionItemListParams struct {
-	EndingBefore  *string   `form:"ending_before"`
-	Expand        []*string `form:"expand"`
-	Limit         *int64    `form:"limit"`
-	StartingAfter *string   `form:"starting_after"`
-	Subscription  *string   `form:"subscription"`
+	ListParams   `form:"*"`
+	Subscription *string `form:"subscription"`
 }
 
 // SubscriptionItem is the resource representing a Stripe subscription item.
--- a/subitem.go
+++ b/subitem.go
@@ -29,7 +29,7 @@ type SubscriptionItemParams struct {
 	Params            `form:"*"`
 	BillingThresholds *SubscriptionItemBillingThresholdsParams `form:"billing_thresholds"`
 	ClearUsage        *bool                                    `form:"clear_usage"`
-	OffSession        *bool                                    `form:"off_session"`
+	UsageGTE          *bool                                    `form:"off_session"` // Only supported on update
 	PaymentBehavior   *string                                  `form:"payment_behavior"`
 	Plan              *string                                  `form:"plan"`
 	Price             *string                                  `form:"price"`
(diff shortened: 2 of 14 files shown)


filelink.go
```diff
@@ -1,14 +1,27 @@
...
+	Params       `form:"*"`
+	ExpiresAt    *int64  `form:"expires_at"`
+	ExpiresAtNow *bool   `form:"-"` // See custom AppendTo
+	File         *string `form:"file"`
+}
+
+// AppendTo implements custom encoding logic for FileLinkParams.
+func (f *FileLinkParams) AppendTo(body *form.Values, keyParts []string) {
+	if BoolValue(f.ExpiresAtNow) {
+		body.Add(form.FormatKey(append(keyParts, "ExpiresAt")), "now")
```
remi-stripe, line 23: This should be `body.Add(form.FormatKey(append(keyParts, "expires_at")), "now")` right? It jumped out at me but now I'm worried I missed some in previous reviews.
We likely need to add custom tests for all custom `AppendTo` to avoid this

subitem.go
```diff
@@ -21,21 +27,20 @@ type SubscriptionItemPriceDataParams struct {
 // For more details see https://stripe.com/docs/api#create_subscription_item and https://stripe.com/docs/api#update_subscription_item.
 type SubscriptionItemParams struct {
 	Params            `form:"*"`
-	ID                *string                                  `form:"-"` // Handled in URL
 	BillingThresholds *SubscriptionItemBillingThresholdsParams `form:"billing_thresholds"`
 	ClearUsage        *bool                                    `form:"clear_usage"`
+	OffSession        *bool                                    `form:"off_session"`
```
remi-stripe, line 32: Do we want to keep the comment for what is only supported on updates? I think it was useful

subitem.go
```diff
@@ -21,21 +27,20 @@ type SubscriptionItemPriceDataParams struct {
...
-	ProrationDate     *int64                                   `form:"proration_date"`
 	ProrationBehavior *string                                  `form:"proration_behavior"`
+	ProrationDate     *int64                                   `form:"proration_date"`
 	Quantity          *int64                                   `form:"quantity"`
 	Subscription      *string                                  `form:"subscription"`
 	TaxRates          []*string                                `form:"tax_rates"`
 
-	// The following parameters are only supported on updates
-	OffSession *bool `form:"off_session"`
+	ID *string `form:"-"` // Deprecated
```
richardm-stripe, line 43: As far as I can tell, this is completely redundant. It's not used by `subitem/client.go`, and since `form:"-"` it is not transmitted. Planning to remove in the next major.

subitem.go
```diff
@@ -47,30 +52,51 @@ type SubscriptionItemBillingThresholdsParams struct {
 // SubscriptionItemListParams is the set of parameters that can be used when listing invoice items.
 // For more details see https://stripe.com/docs/api#list_invoiceitems.
 type SubscriptionItemListParams struct {
-	ListParams   `form:"*"`
-	Subscription *string `form:"subscription"`
+	EndingBefore  *string   `form:"ending_before"`
```
remi-stripe, line 55: Why are we losing `ListParams` suddenly?

subitem.go
```diff
@@ -47,30 +52,51 @@ type SubscriptionItemBillingThresholdsParams struct {
...
+	Expand        []*string `form:"expand"`
+	Limit         *int64    `form:"limit"`
+	StartingAfter *string   `form:"starting_after"`
+	Subscription  *string   `form:"subscription"`
+}
+type SubscriptionItemUsageRecordSummariesParams struct {
```
remi-stripe, line 61: What is this class about?

subitem.go
```diff
@@ -47,30 +52,51 @@ type SubscriptionItemBillingThresholdsParams struct {
...
-	Created           int64                             `json:"created"`
-	Deleted           bool                              `json:"deleted"`
-	ID                string                            `json:"id"`
-	Metadata          map[string]string                 `json:"metadata"`
-	Plan              *Plan                             `json:"plan"`
-	Price             *Price                            `json:"price"`
-	Quantity          int64                             `json:"quantity"`
-	Subscription      string                            `json:"subscription"`
-	TaxRates          []*TaxRate                        `json:"tax_rates"`
+	BillingThresholds *SubscriptionItemBillingThresholds `json:"billing_thresholds"`
```
remi-stripe, line 83: Damn that one is a breaking change so it will need a todo/override

terminal_reader.go
```diff
@@ -24,18 +37,18 @@ type TerminalReaderListParams struct {
...
-	Label           string            `json:"label"`
-	Livemode        bool              `json:"livemode"`
-	Location        string            `json:"location"`
-	Metadata        map[string]string `json:"metadata"`
-	Object          string            `json:"object"`
-	SerialNumber    string            `json:"serial_number"`
-	Status          string            `json:"status"`
+	Deleted         bool                     `json:"deleted"`
+	DeviceSwVersion string                   `json:"device_sw_version"`
+	DeviceType      TerminalReaderDeviceType `json:"device_type"`
```
remi-stripe, line 42: that's a breaking change right?

topup.go
```diff
@@ -53,9 +69,35 @@ type Topup struct {
 	Object                   string              `json:"object"`
 	Source                   *PaymentSource      `json:"source"`
 	StatementDescriptor      string              `json:"statement_descriptor"`
-	Status                   string              `json:"status"`
+	Status                   TopupStatus         `json:"status"`
```
remi-stripe, line 72: breaking change right?

topup.go
```diff
@@ -53,9 +69,35 @@ type Topup struct {
...
 	TransferGroup            string              `json:"transfer_group"`
-
 	// The following property is deprecated
+
```
remi-stripe, line 75: why an empty string here?

remi-stripe (changes requested):

subitem.go
richardm-stripe: I don't think these comments exist consistently, and rather than preserve them piecemeal I'd rather re-introduce them wholesale in a single commit once codegen is ~complete.

remi-stripe: I would recommend having those stay now since it will limit the diff in future codegen instead of having to do it twice and it's quite specific to codegen here. The earliest you decide how to handle the diff between create and update the more likely you are to find potential issues.
I won't block since it doesn't change functionality and no one can really see those unless they read the code, it's just a nudge to think about doing it early

filelink.go
richardm-stripe: **Excellent** catch, slipped right by me. This is the first `AppendTo` I've tried to codegen so nothing slipped past I don't think.

terminal_reader.go
richardm-stripe: No, `string -> enum` in stripe-go is not breaking because `TerminalReaderDeviceType` is just a type alias for string (see line 9).

Here's a Go playground as demonstration: https://play.golang.org/p/5S-W7uC7pbD

topup.go
richardm-stripe: No, see above in `terminal_reader.go`: `string -> enum` is non-breaking.

topup.go
richardm-stripe: I misplaced the newline. Fixed!

subitem.go
richardm-stripe: Added the override!

subitem.go
richardm-stripe: I've removed it -- it didn't belong here.

Long story short, this should be `UsageRecordSummaryListParams` in a different file `usagerecordsummary.go`.

subitem.go
richardm-stripe: Fixed.

Codegen bug. This was because SubscriptionItem has a  "list" method for SubscriptionItems, but in the API definitions also has a method for "list"ing usage record summaries. Codegen was assuming only a single "list" method and was resolving to the wrong one.

subitem.go
richardm-stripe: I've added the comment back but "inline" as

```go
OffSession *bool `form:"off_session"` // Only supported on update
```

Again quite willing to revisit later and segregate properties by method, but for now my focus is on extending codegen coverage as quickly and with as small diffs as possible.

filelink.go
richardm-stripe: (this is fixed)

richardm-stripe:
Re-pinging for this one @remi-stripe.

remi-stripe (approved):

PR #1271, opened 2021-03-31
codegen: add several client.go files

richardm-stripe:
Converts several `client.go` files to the codegenned format. This should be a complete noop. Most of the diff is changing variable names, arbitrary comment content, and some whitespace/ordering changes.

### DIFF
--- a/accountlink/client.go
+++ b/accountlink/client.go
@@ -1,6 +1,10 @@
-// Package accountlink provides API functions related to account links.
 //
-// For more details, see: https://stripe.com/docs/api/go#account_links.
+//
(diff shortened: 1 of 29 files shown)


remi-stripe (approved):
Went through the whole PR carefully and it looks correct. Really exciting to see so many files codegen.

My one worry is that we have ids for resources that can be chosen by merchant and have special characters so we need to make sure the id are always properly encoded in URLs when we set them. The PR seems to properly take this into account already so you likely got this but wanted to mention this as a "gotcha" that bit us before.

PR #1274, opened 2021-04-07
Fix names of SubscriptionScheduleStatus constants

ees37:
This PR updates the names of the `SubscriptionScheduleStatusPastDue` and `SubscriptionScheduleStatusTrialing` constants to match their underlying values. For some reason these are off, which caused some slight confusion when working with them.

### DIFF
--- a/subschedule.go
+++ b/subschedule.go
@@ -20,11 +20,11 @@ type SubscriptionScheduleStatus string
 
 // List of values that SubscriptionScheduleStatus can take.
 const (
-	SubscriptionScheduleStatusActive    SubscriptionScheduleStatus = "active"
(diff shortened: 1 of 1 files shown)


brandur-stripe:
Doh, we just have missed this during review. Thank you @ees37, and sorry about the trouble.

brandur-stripe:
I released this as a patch on v72.41.1 despite the fact that it's technically a breaking change because (1) the previous constants were totally misnamed, and (2) this feature is new enough that we probably haven't picked up a huge number of users on it yet. Made a note in the change log.

ees37:
thanks @brandur-stripe!

PR #1282, opened 2021-05-14
Miscellaneous fixes to README.md

jim-minter:
* Remove reference to IdempotencyKey from README.md
    
  1. It doesn't compile because PaymentIntentListParams doesn't have a Params
     field.
  2. According to https://stripe.com/docs/api/idempotent_requests, "sending
     idempotency keys in GET and DELETE requests has no effect and should be
     avoided, as these requests are idempotent by definition".

* Rename stripeClient -> client in README.md
   
  Fixes compilation failure.

* Miscellaneous README.md corrections
    
  1. Comment out stripe.SetBackend by default for simplicity.
  2. Rewrite stripe.$Resource$Params -> &stripe.$Resource$Params{} as a pointer
     to struct is expected.
  3. Rewrite $resource$, err -> resource, err where needed to avoid accidentally
     redeclaring over the imported package.
  4. Add missing id argument to Update invocation.

*  Comment out code which is deliberately wrong in README.md

*  Fix broken README.md code under "Expanding Objects" section


yejia-stripe (approved):
@jim-minter Thank you for the fixes!

PR #1283, opened 2021-05-17
Add streaming requests

richardm-stripe:
## Notify
r? @dcr-stripe 
cc @stripe/api-libraries 

## Description
Implements support for streaming requests.

How this will be used

```go
// In foo/client.go, something like:
func (c Client) PDF(id string, params *stripe.FooPDFParams) (*stripe.APIStream, error) {
	path := stripe.FormatURLPath("/v1/foo/%s/pdf", id)
	stream := &stripe.APIStream{}
	err := c.B.CallStreaming(http.MethodGet, path, c.Key, params, stream)
	return stream, err
}
```
and then the user can
```go
stream, err := stripe.foo.PDF("foo_xyz", nil)
if err != nil {
  return err  
}
writer, err := os.Create("/tmp/foo.pdf")
defer writer.Close()
io.Copy(writer, stream.LastResponse.Body)
```

It might help to review this commit-by-commit. There's some refactoring into helper functions.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -439,6 +439,47 @@ func resetBodyReader(body *bytes.Buffer, req *http.Request) {
 	}
 }
 
+func (s *BackendImplementation) handleResponse(res *http.Response, err error) ([]byte, error) {
+	var resBody []byte
+	if err == nil {
+		resBody, err = ioutil.ReadAll(res.Body)
+		res.Body.Close()
+	}
+
+	if err != nil {
+		s.LeveledLogger.Errorf("Request failed with error: %v", err)
+	} else if res.StatusCode >= 400 {
+		err = s.ResponseToError(res, resBody)
+
+		if stripeErr, ok := err.(*Error); ok {
+			// The Stripe API makes a distinction between errors that were
+			// caused by invalid parameters or something else versus those
+			// that occurred *despite* valid parameters, the latter coming
+			// back with status 402.
+			//
+			// On a 402, log to info so as to not make an integration's log
+			// noisy with error messages that they don't have much control
+			// over.
+			//
+			// Note I use the constant 402 instead of an `http.Status*`
+			// constant because technically 402 is "Payment required". The
+			// Stripe API doesn't comply to the letter of the specification
+			// and uses it in a broader sense.
+			if res.StatusCode == 402 {
+				s.LeveledLogger.Infof("User-compelled request error from Stripe (status %v): %v",
+					res.StatusCode, stripeErr.redact())
+			} else {
+				s.LeveledLogger.Errorf("Request error from Stripe (status %v): %v",
+					res.StatusCode, stripeErr.redact())
+			}
+		} else {
+			s.LeveledLogger.Errorf("Error decoding error from Stripe: %v", err)
+		}
+	}
+
+	return resBody, err
+}
+
 // Do is used by Call to execute an API request and parse the response. It uses
 // the backend's HTTP client to execute the request and unmarshals the response
 // into v. It also handles unmarshaling errors returned by the API.
@@ -451,7 +492,6 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
 	var resBody []byte
 	for retry := 0; ; {
 		start := time.Now()
-
 		resetBodyReader(body, req)
 
 		res, err = s.HTTPClient.Do(req)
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -118,17 +132,38 @@ func newAPIResponse(res *http.Response, resBody []byte) *APIResponse {
...
+
 // APIResource is a type assigned to structs that may come from Stripe API
 // endpoints and contains facilities common to all of them.
 type APIResource struct {
 	LastResponse *APIResponse `json:"-"`
 }
 
+// APIStream is a type assigned to streaming responses that may come from Stripe API
+type APIStream struct {
+	LastResponse *StreamingAPIResponse `json:"-"`
```
dcr-stripe, line 154: What does the JSON annotation here do?

stripe.go
```diff
@@ -107,6 +107,20 @@ type APIResponse struct {
 	StatusCode int
 }
 
+// StreamingAPIResponse encapsulates some common features of a response from the
+// Stripe API whose body can be streamed. This is used for "file downloads", and
+// the `Body` property is an io.ReadCloser, so the user can stream it to another
+// location such as a file or network request without buffering the entire body
+// into meemory.
```
dcr-stripe, line 114: Nit: meemory -> memory

stripe.go
```diff
@@ -262,6 +298,46 @@ func (s *BackendImplementation) Call(method, path, key string, params ParamsCont
...
 
+// CallStreaming is the Backend.Call implementation for invoking Stripe APIs
+// without buffering the response into memory.
+func (s *BackendImplementation) CallStreaming(method, path, key string, params ParamsContainer, v StreamingLastResponseSetter) error {
+	var formValues *form.Values
+	var commonParams *Params
+
+	if params != nil {
+		reflectValue := reflect.ValueOf(params)
+
```
dcr-stripe, line 309: Could we either a) copy the comment from Call explaining why we do this or b) potentially share the implementation between both to de-dup?

stripe.go
```diff
@@ -380,99 +451,100 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
...
-			req.GetBody = func() (io.ReadCloser, error) {
-				reader := bytes.NewReader(body.Bytes())
-				return nopReadCloser{reader}, nil
-			}
+			return nopReadCloser{reader}, nil
 		}
+	}
+}
+
+// Do is used by Call to execute an API request and parse the response. It uses
```
dcr-stripe, line 518: Stale comment here?

stripe.go
```diff
@@ -487,33 +559,106 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
...
+			s.LeveledLogger.Infof("User-compelled request error from Stripe (status %v): %v",
+				statusCode, stripeErr.redact())
+		} else {
+			s.LeveledLogger.Errorf("Request error from Stripe (status %v): %v",
+				statusCode, stripeErr.redact())
 		}
+	} else {
+		s.LeveledLogger.Errorf("Error decoding error from Stripe: %v", err)
 	}
+}
```
dcr-stripe, line 597: Let's document this method similar to `Do` to make it clear the difference?

dcr-stripe (changes requested):
Looks good overall! Mainly minor comments - thanks Richard!

stripe.go
richardm-stripe, line 114:
```suggestion
// into memory.
```

stripe.go
richardm-stripe: It will suppress the field from `json.Marshal`. This is appropriate for `APIResource` because APIResource is embedded in other structs and the user (or e.g. stripe-go's test suite) would expect `json.Marshal` to produce the same JSON that was returned from the stripe API and not to contain this weird extra LastResponse field, but I think it is not appropriate here as there should never have been any JSON from the stripe API to begin with.
```suggestion
	LastResponse *StreamingAPIResponse
```

stripe.go
```diff
@@ -487,32 +555,112 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
...
+		} else if res.StatusCode >= 400 {
+			err = s.ResponseToError(res, resBody)
+
+			s.logError(res.StatusCode, err)
+		}
 
+		return resBody, err
+	}
+
+	res, result, err := s.requestWithRetriesAndTelemetry(req, body, handleResponse)
```
dcr-stripe, line 656: Super tiny nit: What do you think of resp instead of res? Having `res` and `result` is a bit confusing to me - `resp` would make it more clear that it's a response.

dcr-stripe (approved):
Looks great - thanks Richard! 🥳 Just one comment about naming. Thanks for all the work on this!!

stripe.go
richardm-stripe: 100%, I blame too much express.js for blinding me to the fact that res can mean anything other than response, lol.

PR #1288, opened 2021-05-28
golint -> staticcheck

richardm-stripe:
Golint [considers itself to be deprecated](https://github.com/golang/go/issues/38968).

This replaces Golint with `staticcheck` which seems to be the most often-recommended replacement. This is more like a typical linter and can be configured globally/locally to ignore certain types of errors. In particular, it will not (by default) complain about the absent docstrings in [VerificationSession](https://github.com/stripe/stripe-go/pull/1285/files#diff-48afee747dad594368398d1e4264e75a01764120dd5200f49263a5b76612a4a0), and we will soon be able to import the `description` from OpenAPI as docstrings without triggering an error.


.github/workflows/ci.yml
```diff
@@ -17,8 +17,8 @@ jobs:
 
       - name: golint & go vet
```
dcr-stripe, line 18: Should we update the name of this job?

dcr-stripe (commented):
Thanks Richard - this looks awesome 🎉 just one question pending for ci.yml and then good to go!

.github/workflows/ci.yml
richardm-stripe, line 18:
```suggestion
      - name: lint
```

richardm-stripe: Indeed!

PR #1290, opened 2021-06-01
codegen: price.go

richardm-stripe:


price.go
```diff
@@ -99,16 +101,14 @@ type PriceTierParams struct {
...
+	UpTo              *int64   `form:"up_to"`
+	UpToInf           *bool    `form:"-"` // See custom AppendTo
 }
 
-// AppendTo implements custom up_to serialisation logic for tiers configuration
+// AppendTo implements custom encoding logic for PriceTierParams.
 func (p *PriceTierParams) AppendTo(body *form.Values, keyParts []string) {
 	if BoolValue(p.UpToInf) {
 		body.Add(form.FormatKey(append(keyParts, "up_to")), "inf")
-	} else {
```
richardm-stripe, line 110: This clause appears to be redundant with the default encoder for int64: https://github.com/stripe/stripe-go/blob/master/form/form.go#L319-L325

remi-stripe: i confirmed we have a test for it in https://github.com/stripe/stripe-go/blob/master/price_test.go#L52

remi-stripe (approved):

PR #1302, opened 2021-06-28
Update Support with wechat Pay

Evsio0n:
referencing https://stripe.com/docs/sources/wechat-pay

### DIFF
--- a/source.go
+++ b/source.go
@@ -289,6 +289,13 @@ type SourceSourceOrder struct {
 	Shipping *ShippingDetails          `json:"shipping"`
 }
 
+// SourceWechatParams returns wechat qrcode when create payment source with wechat.
+// For more details please see https://stripe.com/docs/sources/wechat-pay
+//
+type SourceWechatParams struct {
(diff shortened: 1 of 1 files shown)


remi-stripe:
@Evsio0n Thanks for submitting a PR. The `Source` struct is used for the resource, which is what is returned in the API. It's not used for input parameters, which is what you send in the API to create it.

Today, Sources are "polymorphic" and you don't really need to add a WeChat Pay specific information. Instead, you can use `TypeData` which is [here](https://github.com/stripe/stripe-go/blob/master/source.go#L171). The code would look like this:

```
	params := &SourceObjectParams{
		Type: String("wechat"),
		TypeData: map[string]string{
			"qr_code_url": "https://example.com",
		},
	}
```

WeChat Pay support is launching on PaymentMethod in a few days so I'd recommend reaching out to our support team instead though to integrate the latest API:  https://support.stripe.com/contact/email

I'm going to close as we won't merge this change but I hope this unblocks you!

Evsio0n:
Thank you and that helps me a lot.

Issue #1307, opened 2021-07-01
Change type for Error field in SigmaScheduledQueryRun struct


groovypalak:
I am unable to unmarshal the JSON  response from https://api.stripe.com/v1/sigma/scheduled_query_runs to the struct SigmaScheduledQueryRun. In the struct of SigmaScheduledQueryRun, the Error field is of type string, whereas the JSON response returns a field of type Error instead.

remi-stripe:
@groovypalak sorry for the trouble and thank you for the report. I've opened a PR https://github.com/stripe/stripe-go/pull/1309 though it's technically a breaking change and might need to wait a bit until we plan the next major release

PR #1309, opened 2021-07-06
Fix deserialization of Error on Sigma ScheduledQueryRun

remi-stripe:
Fixes https://github.com/stripe/stripe-go/issues/1307

This is a breaking change, but the library would crash on any case of a `ScheduledQueryRun` that has `Error` set so maybe it's worth releasing a patch? We could otherwise shelve until the next major.

r? @richardm-stripe 
cc @stripe/api-libraries


richardm-stripe (approved):
LGTM, I think we should release as a patch.

PR #1310, opened 2021-07-09
API Updates

richardm-stripe:
Codegen for openapi 15c00c6.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `FinalizeQuote` method on resource `Quote`
* Remove support for `Finalize` method on resource `Quote`

### DIFF
--- a/quote.go
+++ b/quote.go
@@ -1,0 +1,356 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
+package stripe
+
+import "encoding/json"
+import "github.com/stripe/stripe-go/v72/form"
+
+// The status of the most recent automated tax calculation for this quote.
+type QuoteAutomaticTaxStatus string
+
+// List of values that QuoteAutomaticTaxStatus can take
+const (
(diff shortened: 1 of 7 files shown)


quote.go
```diff
@@ -0,0 +1,356 @@
...
+	Status     *string `form:"status"`
+}
+
+// Cancels the quote.
+type QuoteCancelParams struct {
+	Params `form:"*"`
+}
+
+// Finalizes the quote.
+type QuoteFinalizeQuoteParams struct {
```
dcr-stripe, line 156: For Invoices, we shortened these to `InvoiceFinalizeParams`. Do we want to do the same here?

dcr-stripe (changes requested):
LG, just one comment about consistent naming and then good to go!

quote.go
dcr-stripe: Spoke offline, this is OK to leave as is. While not consistent in this case, we don't want to have to special case this every time.

dcr-stripe (approved):
LGTM!

PR #1311, opened 2021-07-09
Remove too-opinionated docstring linter rule

richardm-stripe:
## Notify
r? @dcr-stripe 

## Summary
We are beginning to generate docstrings based on the Openapi specification. These docstrings don't begin with the name of the Go struct representing what they describe, so the Golang convention of starting your docstrings this way isn't going to fly. This PR removes that linter rule from the `staticcheck` (our linter) config.

See https://staticcheck.io/docs/checks#ST1021

### DIFF
--- a/staticcheck.conf
+++ b/staticcheck.conf
@@ -1,1 +1,1 @@
-checks = ["all", "-ST1005"]
+checks = ["all", "-ST1005", "-ST1021"]


dcr-stripe (approved):
Interesting linting rule. Makes sense - LG!

PR #1312, opened 2021-07-12
codegen: 14 more files

richardm-stripe:
## Notify
r? @stripe/api-library-reviewers 

## Summary
Makes 14 more files eligible for codegen. This includes adding a number of missing fields (see below).

The most involved thing here is the type alias `LineItemItem` introduced in `checkout/session/client.go`, the type alias ensures that this will not be a breaking change.

There is a go vet error that I will fix in a follow-up, please don't block on that.

## Changelog
  * Add support for `BillingAddressCollection` to `CheckoutSession`
  * Add support for `NetworkReasonCode` to `DisputeReason`
  * Add support for `Object` to `EphemeralKey`, `ApplicationFee`, and `DisputeReason`
  * Add support for `Description` to `Refund`
  * Add const definition for value `blocked` on enum `IssuingCardholderStatus`
  * Bugfix: add support for `Rate` on `CheckoutSessionTotalDetailsBreakdownTax` -- the existing field `TaxRate` has the wrong json annotation and should be deprecated.

### DIFF
--- a/checkout/session/client.go
+++ b/checkout/session/client.go
@@ -1,4 +1,10 @@
-// Package session provides API functions related to checkout sessions.
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
+// Package session provides the /checkout/sessions APIs
 package session
 
 import (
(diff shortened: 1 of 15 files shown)


checkout/session/client.go
```diff
@@ -81,22 +110,7 @@ func (c Client) ListLineItems(id string, listParams *stripe.CheckoutSessionListL
...
-	return i.Current().(*stripe.CheckoutSession)
-}
-
-// CheckoutSessionList returns the current list object which the iterator is
-// currently using. List objects will change as new API calls are made to
-// continue pagination.
-func (i *Iter) CheckoutSessionList() *stripe.CheckoutSessionList {
-	return i.List().(*stripe.CheckoutSessionList)
-}
+type LineItemIter lineitem.Iter
```
remi-stripe, line 113: Does this make it a breaking change on existing integrations?
Asking as there's an `Iter` in `lineitem.go` and some broken tests at the bottom

checkout_session.go
```diff
@@ -429,7 +444,8 @@ type CheckoutSessionTotalDetailsBreakdownDiscount struct {
 // CheckoutSessionTotalDetailsBreakdownTax represent the details of tax rate applied to a session.
 type CheckoutSessionTotalDetailsBreakdownTax struct {
 	Amount  int64    `json:"amount"`
-	TaxRate *TaxRate `json:"tax_rate"`
+	Rate    *TaxRate `json:"rate"`
```
remi-stripe, line 447: Why do we have `Rate` appear here?

richardm-stripe: Ah great question, need to call this out in the changelog and should add a comment. This is a bugfix.

The json field is `tax_rate`, not `rate`, so `TaxRate` will never be populated and will always be nil.

https://github.com/stripe/stripe-node/blob/master/types/2020-08-27/Checkout/Sessions.d.ts#L724

checkout/session/client.go
richardm-stripe: You're right, this is breaking as is -- we need to do `type LineItemIter = lineitem.Iter` to make it a transparent type alias and that will fix the test. I thought I had written the correct override for codegen to do this, need to debug.

dcr-stripe (approved):
LGTM overall - small nits.

PR #1318, opened 2021-07-21
API Updates

richardm-stripe:
Codegen for openapi adfc63f.
r? @richardm-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `Wallet` on `IssuingTransaction`
* Add support for `Ideal` on `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentConfirmPaymentMethodOptionsParams`, and `PaymentIntentPaymentMethodOptions`

### DIFF
--- a/charge/client.go
+++ b/charge/client.go
@@ -1,6 +1,10 @@
-// Package charge provides API functions related to charges.
 //
-// For more details, see: https://stripe.com/docs/api/go#charges.
+//
+// File generated from our OpenAPI spec
+//
+//
+
+// Package charge provides the /charges APIs
 package charge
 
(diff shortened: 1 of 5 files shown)


remi-stripe (commented):
you mentioned removing errors for constants but you didn't. IMO removing is a breaking change we should defer to the next major so it's best anyway

dcr-stripe:
Spoke with Richard - OK to remove the error removal from the changelog as it's not in the code.

dcr-stripe (approved):

PR #1320, opened 2021-07-22
Stop using uploads.stripe.com for the files backend.

dcr-stripe:
r? @richardm-stripe 

Move to using `files.stripe.com`. While both point to the same location, all other SDKs use the newer `files.stripe.com`. This also removes some confusion in this code - both constants were defined but we were using the older constant.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -1050,7 +1050,7 @@ func GetBackendWithConfig(backendType SupportedBackend, config *BackendConfig) B
 	switch backendType {
 	case APIBackend:
 		if config.URL == nil {
-			config.URL = String(apiURL)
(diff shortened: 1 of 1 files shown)


akalinin-stripe:
Looks good to me

richardm-stripe:
Do we consider this breaking? I think the URL to the stripe api is in one sense an "internal to Stripe" implementation detail, but then I do anticipate there could be users whitelisting urls specifically for egress in their infra...

remi-stripe:
@richardm-stripe the egress part is a good point but we didn't treat this as a breaking change in other client libraries at least so I think it's fine to merge as patch

PR #1324, opened 2021-08-09
Add support for search API

akalinin-stripe:
r? @richardm-stripe @dcr-stripe 

## Summary
Add support for search results by adding a new iterator. The new iterator functionality and interface is very similar to existing list iterator, so it should be very familiar to use.

## Testing
- Unit tests for search iterator
- Unit tests against stripe-mock (forthcoming with actual API update)
- Testing via a small test program against the actual search API


search_params.go
```diff
@@ -0,0 +1,100 @@
...
+// structs that embed Params will inherit it, and thus implement the
+// ParamsContainer interface.
+func (p *SearchParams) GetParams() *Params {
+	return p.ToParams()
+}
+
+// SetStripeAccount sets a value for the Stripe-Account header.
+func (p *SearchParams) SetStripeAccount(val string) {
+	p.StripeAccount = &val
+}
```
dcr-stripe, line 82: I don't think we actually need this for search infrastructure right?

search_params.go
```diff
@@ -0,0 +1,100 @@
...
+	// Note that a cancelled or timed out context does not provide any
+	// guarantee whether the operation was or was not completed on Stripe's API
+	// servers. For certainty, you must either retry with the same idempotency
+	// key or query the state of the API.
+	Context context.Context `form:"-"`
+
+	Query        string    `form:"query"`
+	Expand       []*string `form:"expand"`
+	SearchWindow *string   `form:"search_window"`
+	SortOrder    *string   `form:"sort_order"`
```
dcr-stripe, line 48: We don't have any of these parameters in other languages so I think we can drop them here.

search_params.go
```diff
@@ -0,0 +1,100 @@
...
+	Limit        *int64    `form:"limit"`
+	Page     	 *string   `form:"page"`
```
dcr-stripe, line 50: Nit: I think we need to run `go fmt` here.

dcr-stripe (changes requested):
Looks good overall - thanks Pavel! Just some small changes and then I think we're good to go.

akalinin-stripe:
Forgot that I worked on this PR

search_params.go
pakrym-stripe: This is consistent with List params. A way to set stripe account on the search request.

PR #1334, opened 2021-09-01
API Updates

richardm-stripe:
Codegen for openapi 9e4d34e.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `FutureRequirements` on `Account`, `Capability`, and `Person`
* Add support for `Alternatives` on `AccountRequirements`, `CapabilityRequirements`, and `PersonRequirements`

### DIFF
--- a/account.go
+++ b/account.go
@@ -637,6 +637,36 @@ type AccountController struct {
 	Type         AccountControllerType `json:"type"`
 }
 
+// Fields that are due and can be satisfied by providing the corresponding alternative fields instead.
+type AccountFutureRequirementsAlternative struct {
+	AlternativeFieldsDue []string `json:"alternative_fields_due"`
+	OriginalFieldsDue    []string `json:"original_fields_due"`
+}
+
+// Fields that are `currently_due` and need to be collected again because validation or verification failed.
+type AccountFutureRequirementsError struct {
+	Code        string `json:"code"`
+	Reason      string `json:"reason"`
+	Requirement string `json:"requirement"`
+}
+
+type AccountFutureRequirements struct {
+	Alternatives        []*AccountFutureRequirementsAlternative `json:"alternatives"`
+	CurrentDeadline     int64                                   `json:"current_deadline"`
+	CurrentlyDue        []string                                `json:"currently_due"`
+	DisabledReason      string                                  `json:"disabled_reason"`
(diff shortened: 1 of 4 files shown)


account.go
```diff
@@ -745,28 +776,29 @@ type AccountTOSAcceptance struct {
...
+	Company            *AccountCompany            `json:"company"`
+	Controller         *AccountController         `json:"controller"`
+	Country            string                     `json:"country"`
+	Created            int64                      `json:"created"`
+	DefaultCurrency    Currency                   `json:"default_currency"`
+	Deleted            bool                       `json:"deleted"`
+	DetailsSubmitted   bool                       `json:"details_submitted"`
+	Email              string                     `json:"email"`
+	ExternalAccounts   *ExternalAccountList       `json:"external_accounts"`
+	FutureRequirements *AccountFutureRequirements `json:"future_requirements"`
```
remi-stripe, line 792: Shouldn't this be named `AccountRequirementsFutureRequirements` instead?

account.go
```diff
@@ -637,6 +637,36 @@ type AccountController struct {
...
+
+// Fields that are `currently_due` and need to be collected again because validation or verification failed.
+type AccountFutureRequirementsError struct {
+	Code        string `json:"code"`
+	Reason      string `json:"reason"`
+	Requirement string `json:"requirement"`
+}
+
+type AccountFutureRequirements struct {
+	Alternatives        []*AccountFutureRequirementsAlternative `json:"alternatives"`
```
remi-stripe, line 654: This should be named `AccountRequirementsFutureRequirementsAlternative`

account.go
```diff
@@ -637,6 +637,36 @@ type AccountController struct {
...
+	CurrentDeadline     int64                                   `json:"current_deadline"`
+	CurrentlyDue        []string                                `json:"currently_due"`
+	DisabledReason      string                                  `json:"disabled_reason"`
```
remi-stripe, line 657: This is tricky, it's the same as `AccountRequirementsDisabledReason` right? It should be either that enum or its own enum

remi-stripe (changes requested):
Stopping mid review as many names are not correct I think, can you have a look first?

remi-stripe (approved):

PR #1337, opened 2021-09-02
Generated go test suites

etsai-stripe:
## Notify
r? @richardm-stripe 

```
st-etsai1:stripe-go etsai$ go test ./example
ok  	github.com/stripe/stripe-go/v72/example	0.360s
```

## Summary

In addition to the API spec used to generate most of this SDK, Stripe also
maintains a corpus of example requests that are expected to work consistently
across all supported languages.

This introduces a new functional test, generated from the spec and example
corpus, to increase confidence that critical flows and use-cases work as
expected in Go.

See similar PRs in [stripe-java](https://github.com/stripe/stripe-java/pull/1256)
This PR introduces a codegenned test suite based off examples similar to the ones in the [Stripe API Reference](https://stripe.com/docs/api)

### DIFF
--- a/example/example_generated_test.go
+++ b/example/example_generated_test.go
@@ -48,6 +48,7 @@ import (
 	source "github.com/stripe/stripe-go/v72/source"
 	taxid "github.com/stripe/stripe-go/v72/taxid"
 	taxrate "github.com/stripe/stripe-go/v72/taxrate"
+	location "github.com/stripe/stripe-go/v72/terminal/location"
 	reader "github.com/stripe/stripe-go/v72/terminal/reader"
 	_ "github.com/stripe/stripe-go/v72/testing"
 	topup "github.com/stripe/stripe-go/v72/topup"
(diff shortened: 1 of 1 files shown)


example/example_generated_test.go
```diff
@@ -0,0 +1,1447 @@
...
+	result := customer.List(params)
+	assert.NotNil(t, result)
+}
+
+func TestBalanceTransactionRetrieve1(t *testing.T) {
+	result, _ := balancetransaction.Get("txn_xxxxxxxxxxxxx", nil)
+	assert.NotNil(t, result)
+}
+
+func TestBalanceTransactionList1(t *testing.T) {
```
richardm-stripe, line 68: Could we remove the `1` for test names that aren't duplicated?

etsai-stripe: ok let me regenerate the suites

example/example_generated_test.go
```diff
@@ -0,0 +1,1447 @@
...
+}
+
+func TestRefundList1(t *testing.T) {
+	params := &stripe.RefundListParams{}
+	params.Filters.AddFilter("limit", "", "3")
+	result := refund.List(params)
+	assert.NotNil(t, result)
+}
+
+// TODO Token test cases skipped because not all Token methods are supported by GO
```
richardm-stripe, line 416: Could we deduplicate these comments, or maybe remove them altogether? As long as we have todo items inside the codegen source code, that is enough. We don't need to track the TODOs here as well.

etsai-stripe: wil do

richardm-stripe (approved):
Nice! This looks good to me!

I've confirmed the test runs and is passing in CI! 

https://github.com/stripe/stripe-go/pull/1337/checks?check_run_id=3500981651#step:5:37

richardm-stripe (approved):
Looks good! I've confirmed that the example tests run and pass in CI!

https://github.com/stripe/stripe-java/pull/1256/checks?check_run_id=3500826713#step:7:785

PR #1340, opened 2021-09-10
paymentmethod: add Mandate string field to PaymentMethodSepaDebit

anicet:
As Stripe API return `mandate` in JSON payload (at least in PaymentIntent or Charge), field could be present in the struct too.

### DIFF
--- a/paymentmethod.go
+++ b/paymentmethod.go
@@ -409,6 +409,7 @@ type PaymentMethodSepaDebit struct {
 	Fingerprint   string                               `json:"fingerprint"`
 	Last4         string                               `json:"last4"`
 	GeneratedFrom *PaymentMethodSepaDebitGeneratedFrom `json:"generated_from"`
+	Mandate       string                               `json:"mandate"`
 }
 
 // PaymentMethodSofort represents the Sofort-specific properties.
--- a/paymentmethod_test.go
+++ b/paymentmethod_test.go
@@ -1,0 +1,27 @@
+package stripe
+
+import (
+	"encoding/json"
+	"testing"
+
+	assert "github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


remi-stripe:
@anicet thanks a lot for the PR. Do you have a bit more details on where you see `mandate` returned on a SEPA Debit PaymentMethod? This property is not documented in our [API reference](https://stripe.com/docs/api/payment_methods/object) and a SEPA Debit PaymentMethod should not have an associated mandate link since you could have N separate mandates for the same PaymentMethod. 

I looked at our code and I don't see us returning `sepa_debit[mandate]` on `PaymentMethod` though I might be missing something in the code (a lot of it has multiple layers of abstractions).

Is it possible you're mixing this with `payment_method_details[sepa_debit][mandate]` on `Charge` as documented [here](https://stripe.com/docs/api/charges/object#charge_object-payment_method_details-sepa_debit-mandate) and available in the stripe-go library [here](https://github.com/stripe/stripe-go/blob/master/charge.go#L528)?

anicet:
Hi @remi-stripe,
Thanks for your reply. I was trying to access to the `Mandate` from `PaymentIntent.PaymentMethod`, first mistake, it should be done through `PaymentIntent.Charges`.
Then trying to access to the `Mandate` from a bad version of the API, `Mandate` on `ChargePaymentMethodDetailsSepaDebit` seems to have been added recently (https://github.com/stripe/stripe-go/pull/1206). 

`go get github.com/stripe/stripe-go` is fetching v70.15.0
I move the code on `github.com/stripe/stripe-go/v72` and all seems ok.

Sorry for the trouble, I guess we can close the PR !

Thanks

remi-stripe:
@anicet Thanks for circling back and glad you got unblocked!

PR #1384, opened 2021-11-28
godoc is no more

Fale:
Since godoc has been decommissioned in favor of pkg.dev.go, the readme should reflect this change


richardm-stripe:
Hello @Fale, thank you for the contribution!

PR #1393, opened 2021-12-17
API Updates

dcr-stripe:
Codegen for openapi 26c0399.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `Processing` on `PaymentIntent`

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -649,6 +649,13 @@ type PaymentIntentPaymentMethodOptions struct {
 	Sofort           *PaymentIntentPaymentMethodOptionsSofort           `json:"sofort"`
 	WechatPay        *PaymentIntentPaymentMethodOptionsWechatPay        `json:"wechat_pay"`
 }
+type PaymentIntentProcessingCard struct{}
+
+// If present, this property tells you about the processing state of the payment.
+type PaymentIntentProcessing struct {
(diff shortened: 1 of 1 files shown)


paymentintent.go
```diff
@@ -649,6 +649,13 @@ type PaymentIntentPaymentMethodOptions struct {
 	Sofort           *PaymentIntentPaymentMethodOptionsSofort           `json:"sofort"`
 	WechatPay        *PaymentIntentPaymentMethodOptionsWechatPay        `json:"wechat_pay"`
 }
+type PaymentIntentProcessingCard struct{}
+
+// If present, this property tells you about the processing state of the payment.
+type PaymentIntentProcessing struct {
+	Card *PaymentIntentProcessingCard `json:"card"`
+	Type string                       `json:"type"`
```
remi-stripe, line 657: Shouldn't this be `PaymentMethodType` or an enum in general?

remi-stripe (approved):
Approving to unblock but usually we want an enum for those string properties

paymentintent.go
dcr-stripe: Thanks Remi - it's a different set of values than PaymentMethodType (stricter limit), but agree that this should be an enum. I'll merge this and then we're going to update it to be an enum before releasing.

PR #1396, opened 2021-12-21
Make bankaccount and card codegen-able

yejia-stripe:
## Summary
Make bankaccount.go and card.go and related client.go codegen-able.

The first commit is rearranging files and the last is the codegen changes.

## Notify
r? @dcr-stripe
cc @richardm-stripe 

## Changelog
* Add support for `address_city`, `address_country`, `address_line1`, `address_line2`, `address_state`, `address_zip`, `exp_month`, `exp_year`, and `name` on `BankAccountParams`
* Add support for `account_holder_name`, `account_holder_type`, and `owner` on `CardParams`
* Add support for `account` on `Card`

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -44,19 +44,17 @@ const (
 // some unusual logic on creates that necessitates manual handling of all
 // parameters. See AppendToAsSourceOrExternalAccount.
 type BankAccountParams struct {
-	Params `form:"*"`
-
+	Params   `form:"*"`
+	Customer *string `form:"-"`
 	// Account is the identifier of the parent account under which bank
 	// accounts are nested.
-	Account *string `form:"-"`
-
+	Account            *string `form:"-"`
 	AccountHolderName  *string `form:"account_holder_name"`
 	AccountHolderType  *string `form:"account_holder_type"`
-	AccountType        *string `form:"account_type"`
 	AccountNumber      *string `form:"account_number"`
+	AccountType        *string `form:"account_type"`
 	Country            *string `form:"country"`
 	Currency           *string `form:"currency"`
-	Customer           *string `form:"-"`
 	DefaultForCurrency *bool   `form:"default_for_currency"`
 	RoutingNumber      *string `form:"routing_number"`
 
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -1,10 +1,15 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
 package stripe
 
 import (
 	"encoding/json"
-	"strconv"
-
 	"github.com/stripe/stripe-go/v72/form"
+	"strconv"
 )
 
 // BankAccountAvailablePayoutMethod is a set of available payout methods for the card.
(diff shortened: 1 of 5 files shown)


bankaccount.go
```diff
@@ -155,7 +166,7 @@ type BankAccount struct {
 	Account                *Account                           `json:"account"`
 	AccountHolderName      string                             `json:"account_holder_name"`
 	AccountHolderType      BankAccountAccountHolderType       `json:"account_holder_type"`
-	AccountType            *string                            `form:"account_type"`
+	AccountType            string                             `json:"account_type"`
```
yejia-stripe, line 169: I think this is right?

card/client.go
```diff
@@ -23,18 +29,16 @@ func New(params *stripe.CardParams) (*stripe.Card, error) {
...
 	}
 
 	var path string
-	if params.Account != nil {
-		path = stripe.FormatURLPath("/v1/accounts/%s/external_accounts",
-			stripe.StringValue(params.Account))
-	} else if params.Customer != nil {
-		path = stripe.FormatURLPath("/v1/customers/%s/sources",
-			stripe.StringValue(params.Customer))
+	if params.Customer != nil {
```
yejia-stripe, line 36: Re-ordered these conditional cases. I _think_ these should all have either `Customer` or `Account`, so this should be fine.

bankaccount.go
dcr-stripe: Good catch - I probably made a mistake manually porting this in https://github.com/stripe/stripe-go/pull/1322/files#diff-ccd7af701903ec5371bec49de5a76c5574d82e613696da9a0046258b07e476d1R158.

card/client.go
dcr-stripe: This is technically a behavior change since we don't enforce that exclusively one is set... this is generated as part of codegen though and so likely can't be ordered right?

example/generated_examples_test.go
```diff
@@ -463,6 +464,31 @@ func TestPaymentMethodDetach(t *testing.T) {
 	assert.NotNil(t, result)
 }
 
+func TestCardUpdate(t *testing.T) {
```
dcr-stripe, line 467: We should skip these 4 generated tests.

We don't currently set either a `customer` or `account` in these, so the call fails.

You can verify this by changing say:

```
	result, _ := card.Update("ba_xxxxxxxxxxxxx", params)
```

to

```
	result, err := card.Update("ba_xxxxxxxxxxxxx", params)
```

and printing out `err`:

```
Invalid card params: either Customer or Account need to be set
```

Our underlying test manifest should probably be updated to have these cases, but in the mean time we'll rely on the more fleshed out tests in `card/client_test.go` which does set these values.

dcr-stripe (commented):
PTAL @yejia-stripe

card/client.go
yejia-stripe: This part is actually hardcoded. Updated to match existing code!

yejia-stripe:
re-r? @dcr-stripe

yejia-stripe:
Hm. PTAL @dcr-stripe ??

dcr-stripe (approved):
LGTM! 🎉

PR #1397, opened 2021-12-22
API Updates

dcr-stripe:
Codegen for openapi 067bdc5.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `AUBECSDebit` on `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentConfirmPaymentMethodOptionsParams`, and `PaymentIntentPaymentMethodOptions`
* Change type of `PaymentIntentProcessingType` from `string` to `literal('card')`. This is not considered a breaking change as the field was added in the same release.

### DIFF
--- a/paymentintent.go
+++ b/paymentintent.go
@@ -138,6 +138,14 @@ const (
 	PaymentIntentPaymentMethodOptionsWechatPayClientWeb     PaymentIntentPaymentMethodOptionsWechatPayClient = "web"
 )
 
+// Type of the payment method for which payment is in `processing` state, one of `card`.
+type PaymentIntentProcessingType string
+
+// List of values that PaymentIntentProcessingType can take
+const (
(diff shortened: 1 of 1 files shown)


paymentintent.go
```diff
@@ -654,7 +668,7 @@ type PaymentIntentProcessingCard struct{}
 // If present, this property tells you about the processing state of the payment.
 type PaymentIntentProcessing struct {
 	Card *PaymentIntentProcessingCard `json:"card"`
-	Type string                       `json:"type"`
+	Type PaymentIntentProcessingType  `json:"type"`
```
dcr-stripe, line 671: This change is fine since we haven't actually released https://github.com/stripe/stripe-go/pull/1393 yet.

yejia-stripe (approved):

PR #1400, opened 2021-12-27
Make paymentsource.go and client codegen-able

yejia-stripe:
## Summary
Make paymentsource.go and paymentsource/client.go codegen-able. 

Also make some formatting changes for bankaccount/ and card/client.go.

The first commit is rearranging files and the last is the codegen changes.

## Notify
r? @richardm-stripe
cc @dcr-stripe

## Changelog
* Add support for `account_holder_name`, `account_holder_type`, `address_city`, `address_country`, `address_line1`, `address_line2`, `address_state`, `address_zip`, `exp_month`, `exp_year`, `name`, `owner` on `CustomerSourceParams`
* Add support for `PaymentSourceOwnerParams`
* Add support for `Object` on `SourceListParams`

### DIFF
--- a/paymentsource.go
+++ b/paymentsource.go
@@ -1,9 +1,14 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
 package stripe
 
 import (
 	"encoding/json"
 	"fmt"
-
 	"github.com/stripe/stripe-go/v72/form"
 )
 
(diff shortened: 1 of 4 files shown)


paymentsource.go
```diff
@@ -36,20 +41,46 @@ func (p *SourceParams) AppendTo(body *form.Values, keyParts []string) {
...
+	Email   *string        `form:"email"`
+	Name    *string        `form:"name"`
+	Phone   *string        `form:"phone"`
 }
 
 // SourceVerifyParams are used to verify a customer source
 // For more details see https://stripe.com/docs/guides/ach-beta
 type SourceVerifyParams struct {
 	Params   `form:"*"`
+	Customer *string   `form:"-"`       // Included in URL
```
richardm-stripe, line 71: Is this used? The `Verify` method in `paymentsource/client.go` uses a passed-in `id` parameter instead of taking the id off params.

richardm-stripe (changes requested):
I don't think `SourceVerifyParams.Customer` should be included, but otherwise this looks good!

paymentsource.go
yejia-stripe: @richardm-stripe This line was already there, but it got moved (previously on line 49)

richardm-stripe: Ah, missed that!

richardm-stripe (approved):

PR #1405, opened 2022-01-14
Generate struct field docstrings

yejia-stripe:
## Summary
Remove pre-codegen comments and generate struct field docstrings from the OpenAPI spec.

VSCode documentation popup
Screen Shot 2022-01-14 at 1 19 45 PM

Preview using the [godoc](https://pkg.go.dev/golang.org/x/tools/cmd/godoc) tool
Screen Shot 2022-01-14 at 1 44 49 PM

## Notify
r? @richardm-stripe
cc @dcr-stripe

PR #1403, opened 2022-01-12
API Updates

dcr-stripe:
Codegen for openapi 31569c9.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `CustomerCreation` on `CheckoutSessionParams` and `CheckoutSession`
* Add support for `FPX` and `Grabpay` on `PaymentIntentPaymentMethodOptionsParams` and `PaymentIntentPaymentMethodOptions`

PR #1407, opened 2022-01-19
API Updates

dcr-stripe:
Codegen for openapi 7734045.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Change type of `ChargeStatus` from `string` to `enum('failed'|'pending'|'succeeded')`
* Add support for `BACSDebit` and `EPS` on `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentPaymentMethodOptionsParams`, `PaymentIntentConfirmPaymentMethodOptionsParams`, and `PaymentIntentPaymentMethodOptions`
* Add support for `ImageURLPNG` and `ImageURLSVG` on `PaymentIntentNextActionWechatPayDisplayQRCode`

### DIFF
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -47,6 +47,15 @@ const (
 	CheckoutSessionConsentCollectionPromotionsAuto CheckoutSessionConsentCollectionPromotions = "auto"
 )
 
+// Configure whether a Checkout Session creates a Customer when the Checkout Session completes.
+type CheckoutSessionCustomerCreation string
+
+// List of values that CheckoutSessionCustomerCreation can take
+const (
+	CheckoutSessionCustomerCreationAlways     CheckoutSessionCustomerCreation = "always"
+	CheckoutSessionCustomerCreationIfRequired CheckoutSessionCustomerCreation = "if_required"
+)
+
 // The customer's tax exempt status at time of checkout.
 type CheckoutSessionCustomerDetailsTaxExempt string
 
@@ -376,9 +385,12 @@ type CheckoutSessionPaymentIntentDataParams struct {
 	// customer that their payment details will be saved and used for future
 	// payments.
 	//
-	// For both values, Checkout will attach the payment method to either the
-	// provided Customer for the session, or a new Customer created by Checkout
-	// if one has not been provided.
+	// If a Customer has been provided or Checkout creates a new Customer,
+	// Checkout will attach the payment method to the Customer.
+	//
+	// If Checkout does not create a Customer, the payment method is not attached
+	// to a Customer. To reuse the payment method, you can retrieve it from the
+	// Checkout Session's PaymentIntent.
 	//
 	// When processing card payments, Checkout also uses `setup_future_usage`
 	// to dynamically optimize your payment flow and comply with regional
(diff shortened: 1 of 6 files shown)


invoice.go
```diff
@@ -669,6 +669,8 @@ type Invoice struct {
 	OnBehalfOf *Account `json:"on_behalf_of"`
 	// Whether payment was successfully collected for this invoice. An invoice can be paid (most commonly) with a charge or with credit from the customer's account balance.
 	Paid bool `json:"paid"`
+	// Returns true if the invoice was manually marked paid, returns false if the invoice hasn't been paid yet or was paid on Stripe.
+	PaidOutOfBand bool `json:"paid_out_of_band"`
```
yejia-stripe, line 673: Could you add this to the changelog?

paymentintent.go
```diff
@@ -697,6 +717,10 @@ type PaymentIntentNextActionWechatPayDisplayQRCode struct {
 	Data string `json:"data"`
 	// The base64 image data for a pre-generated QR code
 	ImageDataURL string `json:"image_data_url"`
+	// The image_url_png string used to render QR code, can be used as <img src="…" />
+	ImageURLPng string `json:"image_url_png"`
```
yejia-stripe, line 721: PNG and SVG should be capitalized here

checkout_session.go
```diff
@@ -625,6 +637,15 @@ type CheckoutSessionParams struct {
...
 	Customer *string `form:"customer"`
+	// Configure whether a Checkout Session creates a [Customer](https://stripe.com/docs/api/customers) during Session confirmation.
+	//
+	// When a Customer is not created, you can still retrieve email, address, and other customer data entered in Checkout
+	// with [customer_details](https://stripe.com/docs/api/checkout/sessions/object#checkout_session_object-customer_details).
+	//
+	// Sessions that do not create Customers will instead create [Guest Customers](https://support.stripe.com/questions/guest-customer-faq) in the Dashboard.
+	//
+	// Can only be set in `payment` and `setup` mode.
+	CustomerCreation *string `form:"customer_creation"`
```
yejia-stripe, line 648: Could you add this to the changelog?

yejia-stripe (commented):
It looks like this PR contains some changes not mentioned in the changelog, and that I thought were already merged in. Could you confirm that they're expected?

checkout_session.go
dcr-stripe: They got added in https://github.com/stripe/stripe-go/pull/1403. It seems they accidentally got removed in https://github.com/stripe/stripe-go/pull/1405/files#diff-bcecc69ea00cdcf31411069c0bcb9ef71ab4e490488e20c3690db3244a4c79b5R627 though. Maybe that generation wasn't done against HEAD? Thankfully that didn't get released yet though so not a big deal. I'll add it to the changelog.

invoice.go
dcr-stripe: I think this also got removed accidentally in #1405. Since it wasn't released, I don't think we need either one in the changelog since this just brings us back to the original state.

checkout_session.go
yejia-stripe: Oops! Yeah I was doing it against whatever I had in my openapi repo at the time, and I must have not updated that. Sorry 🙈

paymentintent.go
dcr-stripe: Done!

dcr-stripe:
r? @yejia-stripe 

PTAL! I addressed all comments.

yejia-stripe (approved):

PR #1409, opened 2022-01-20
[WIP] Remove dependencies only needed for test

richardm-stripe:


yejia-stripe:
`go.mod` should include all the dependencies needed to build and test the code, so we should keep these.

PR #1413, opened 2022-02-03
API Updates

yejia-stripe:
Codegen for openapi 10fe9f5.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Change type of `RefundReason` from `string` to `enum('duplicate'|'expired_uncaptured_charge'|'fraudulent'|'requested_by_customer')`


dcr-stripe (commented):
This PR is effectively a no-op (comment only), so let's close out. No release needed.

PR #1429, opened 2022-03-09
Remove `,string` struct tag from ApplicationFeePercent

Arsaev:
With this tag in place it causes unmarshalling error.
Backend return a number, tag expects string and field is a float64


remi-stripe:
@Arsaev Thanks a lot for the PR and catching the bug. We're looking into it!

kamil-stripe (approved):

PR #1440, opened 2022-03-29
API Updates

pakrym-stripe:
Codegen for openapi 1700e20.
r? @yejia-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `CancelAction`, `ProcessPaymentIntent`, `ProcessSetupIntent`, and `SetReaderDisplay` methods on resource `Terminal.Reader`
* Add support for `Action` on `TerminalReader`

### DIFF
--- a/testhelpersterminal_reader.go
+++ b/testhelpersterminal_reader.go
@@ -1,0 +1,22 @@
+//
+//
+// File generated from our OpenAPI spec
+//
+//
+
+package stripe
+
+// Simulated card present data
+type TestHelpersTerminalReaderPresentPaymentMethodCardPresentParams struct {
+	// Card Number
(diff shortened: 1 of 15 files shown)


testhelpersterminal_reader.go
```diff
@@ -0,0 +1,22 @@
+//
```
yejia-stripe, line 1: nit: could we add a "_" in the file name between "testhelpers" and "terminal_reader"? or is this the convention?

yejia-stripe (approved):
lgtm!

testhelpersterminal_reader.go
pakrym-stripe: From looking at the code this pattern seems to be very intentional.

PR #1441, opened 2022-03-29
Unmarshals name of plan

sergethompson:
Added the unmarshaling the Plan name attribute from the json event data

### DIFF
--- a/plan.go
+++ b/plan.go
@@ -8,8 +8,9 @@ package stripe
 
 import (
 	"encoding/json"
-	"github.com/stripe/stripe-go/v72/form"
 	"strconv"
+
+	"github.com/stripe/stripe-go/v72/form"
 )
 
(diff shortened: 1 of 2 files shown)


remi-stripe:
@sergethompson The [Plan](https://stripe.com/docs/api/plans/object) API Resource does **not** have a `name` property today, which is why it's not in stripe-go. This changed over 4 years ago in [2018-02-05](https://stripe.com/docs/upgrades#2018-02-05) when `name` moved to `Product` and `Plan` got `nickname` instead.

The stripe-go library is pinned to the most recent API version, currently `2020-08-27`, and all API requests made with the library use that version and expect that version for proper de-serialization. This means it's not possible to get a `Plan` with a `name` property unless you explicitly deserialize an object retrieved/rendered with a different API version.

If you are receiving events with the library, you need to make sure you get events pinned to the exact API version the library expects. Without that you risk real deserialization errors due to a shape mismatch such as a boolean becoming a hash or a string for example. When listening to events, you need to configure your webhook endpoint to specifically pin to the same API version as the library via the [api_version](https://stripe.com/docs/api/webhook_endpoints/create#create_webhook_endpoint-api_version) parameter.

I'm going to close the PR as it's not something we would merge but I hope this helps clarify the confusion!

PR #1456, opened 2022-05-05
Added missing metadata field on checkout

cascoid:
https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-metadata

Metadata field from create checkout session params is missing. 

Present for the subscription and payment intent data fields, but not the checkout session creation.

Switching over from node (ts) to go and this is a small issue I've encountered. Here is it present on the node type for comparison: 
https://github.com/stripe/stripe-node/blob/babff9b2056af9a810f3c3e3da99e25984ac3242/types/2020-08-27/Checkout/Sessions.d.ts#L1040

### DIFF
--- a/checkout/session/client_test.go
+++ b/checkout/session/client_test.go
@@ -30,6 +30,10 @@ func TestCheckoutSessionNew(t *testing.T) {
 				Quantity: stripe.Int64(2),
 			},
 		},
+		Metadata: map[string]string{
+			"attr1": "val1",
+			"attr2": "val2",
+		},
 		PaymentIntentData: &stripe.CheckoutSessionPaymentIntentDataParams{
 			Description: stripe.String("description"),
 			Metadata: map[string]string{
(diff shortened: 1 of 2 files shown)


remi-stripe:
@cascoid The parameter is not missing, it's automatically added because of this part
```
type CheckoutSessionParams struct {
	Params `form:"*"`
```

It's the same on other APIs such as `CustomerParams` [here](https://github.com/stripe/stripe-go/blob/master/customer.go#L136).

It can be confusing, but the library used to be written by hand and it avoided repeating `Metadata` everywhere (though we could change that now that we automatically generate all the files)

Does that make sense? If so, we can close this PR.

cascoid:
> @cascoid The parameter is not missing, it's automatically added because of this part
> 
> ```
> type CheckoutSessionParams struct {
> 	Params `form:"*"`
> ```
> 
> It's the same on other APIs such as `CustomerParams` [here](https://github.com/stripe/stripe-go/blob/master/customer.go#L136).
> 
> It can be confusing, but the library used to be written by hand and it avoided repeating `Metadata` everywhere (though we could change that now that we automatically generate all the files)
> 
> Does that make sense? If so, we can close this PR.

Yep was just about to close the PR as I discovered this myself when playing around a bit more. Apologies, but thanks for the clarification!

PR #1467, opened 2022-05-30
adds missing metadata in create customer params

justJay-dev:


remi-stripe:
Just to clarify, `metadata` is already automatically provided by the inclusion of `Params` [here](https://github.com/stripe/stripe-go/blob/master/customer.go#L137)

PR #1468, opened 2022-05-30
FIX: Missing Metadata field in CustomerParms type

justJay-dev:
Hey there! This is a PR that adds a missing `Metadata` field in the create customer. Found this when working on a project today.

### DIFF
--- a/customer.go
+++ b/customer.go
@@ -157,7 +157,7 @@ type CustomerParams struct {
 	// Default invoice settings for this customer.
 	InvoiceSettings *CustomerInvoiceSettingsParams `form:"invoice_settings"`
 	// Set of [key-value pairs](https://stripe.com/docs/api/metadata) that you can attach to an object. This can be useful for storing additional information about the object in a structured format.
-	Metadata map[string]string `json:"metadata"`
+	Metadata map[string]string `form:"metadata"`
 	// The customer's full name or business name.
 	Name *string `form:"name"`
 	// The sequence to be used on the customer's next invoice. Defaults to 1.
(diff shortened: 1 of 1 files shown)


remi-stripe:
Hello! I wanted to clarify that `metadata` is already automatically provided by the inclusion of `Params` [here](https://github.com/stripe/stripe-go/blob/master/customer.go#L137) which exists on all our APIs. I'm going to close as the current version already works!

justJay-dev:
> Hello! I wanted to clarify that `metadata` is already automatically provided by the inclusion of `Params` [here](https://github.com/stripe/stripe-go/blob/master/customer.go#L137) which exists on all our APIs. I'm going to close as the current version already works!

hey thanks for taking a look! I saw where you pointed to initially, but I was not able to get it to work, thus my PR. 

```go
params := &stripe.CustomerParams{
	Name:  stripe.String(payload.Name),
	Email: stripe.String(payload.Email),
	Metadata: map[string]string{
		"dbId": payload.Metadata.DBID,
	},
}
```

is throwing the following error, which lead me to believe it's not in the type. 

image

Is there somewhere in the docs you can point me to? I was expecting it to match or be similar to the node and python packages behavior, is that not the case?

remi-stripe:
You should either use our `AddMetadata` helper documented [here](https://stripe.com/docs/api/metadata?lang=go) or set `Metadata` inside `Params` like this:
```
params := &stripe.CustomerParams{
  Name: stripe.String("name"),
  Params: stripe.Params{
    Metadata: map[string]string{
      "dbId": "value",
    },
  },
}
```

PR #1470, opened 2022-05-31
Metadata field for CheckoutSessionLineItemPriceDataParams type

RensTillmann:
Added missing Metadata type for `line_items` when creating Checkout Sessions: https://stripe.com/docs/api/checkout/sessions/create#create_checkout_session-line_items-price_data-metadata

### DIFF
--- a/checkout_session.go
+++ b/checkout_session.go
@@ -333,6 +333,8 @@ type CheckoutSessionLineItemPriceDataParams struct {
 	UnitAmount *int64 `form:"unit_amount"`
 	// Same as `unit_amount`, but accepts a decimal value in %s with at most 12 decimal places. Only one of `unit_amount` and `unit_amount_decimal` can be set.
 	UnitAmountDecimal *float64 `form:"unit_amount_decimal,high_precision"`
+	// Set of [key-value pairs](https://stripe.com/docs/api/metadata) that you can attach to an object. This can be useful for storing additional information about the object in a structured format. Individual keys can be unset by posting an empty value to them. All keys can be unset by posting an empty value to `metadata`.
+	Metadata map[string]string `form:"metadata"`
 }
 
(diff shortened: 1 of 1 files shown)


remi-stripe:
@RensTillmann Thanks for the PR! Unfortunately this is not something we can merge for now. The feature you are using is not public today and you only have access to it because your account has a beta enabled. You can confirm this easily in the API reference where it would say `Preview feature`.

Beta features are not in the stripe-go library as they could change over time which would be a breaking change. For now you will need to use the `AddExtra` method to handle gated parameters like this one.

RensTillmann:
Hi, ok I understand. One question, how would one go about using the AddExtra method exactly? I have looked in the code but it only uses a receiver *Params. So how would one implement this on the `CheckoutSessionLineItemPriceDataParams` type? Perhaps it's not possible. Which is fine, because I can add the metadata to the `Product` object instead.

remi-stripe:
Something like this should work:

```
params := &stripe.CheckoutSessionParams{
	CancelURL:  stripe.String("https://example.com/cancel"),
	SuccessURL: stripe.String("https://example.com/success"),
	LineItems: []*stripe.CheckoutSessionLineItemParams{
		{
			PriceData: &stripe.CheckoutSessionLineItemPriceDataParams{
				UnitAmount: stripe.Int64(111),
				Currency:   stripe.String(string(stripe.CurrencyUSD)),
			},
			Quantity: stripe.Int64(2),
		},
	},
	Mode: stripe.String(string(stripe.CheckoutSessionModePayment)),
}
params.AddExtra("line_items[0][price_data][metadata][my_key]", "my value")
params.AddExtra("line_items[0][price_data][metadata][my_other_key]", "my other value")
session, err := session.New(params)
```

PR #1483, opened 2022-06-22
Document use of undocumented parameters/properties

rmanzer-stripe:
r? @yejia-stripe

Adding examples of how to access undocumented parameters/properties to the README.  While I know the code added works it may not match Go standards/style guides.  Please feel free to suggest/make edits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ c.Customer.Name  // Name is now also available (if it had a value)
 
 stripe-go is a typed library and it supports all public properties or parameters.
 
-Stripe sometimes has beta which introduces new properties or parameters that are not immediately public. The library does not support these properties or parameters until they are public but there is still an approach that allows you to use them.
+Stripe sometimes launches private beta features which introduce new properties or parameters that are not immediately public. These will not have typed accessors in the stripe-go library but can still be used. 
 
 #### Parameters
 


rmanzer-stripe:

README.md
```diff
@@ -370,7 +370,47 @@ c, _ := charge.Retrieve("ch_123", p)
...
 ```
+### How to use undocumented parameters and properties
 
+stripe-go is a typed library and it supports all public properties or parameters.
+
+Stripe sometimes has beta which introduces new properties or parameters that are not immediately public. The library does not support these properties or parameters until they are public but there is still an approach that allows you to use them.
+
+#### Parameters
+
+To pass undocumented parameters to Stripe using stripe-go you need to use the `putExtraParam()` method, as shown below:
```
dcr-stripe, line 381: `putExtraParam` here should be `AddExtra` I believe?

README.md
```diff
@@ -370,7 +370,47 @@ c, _ := charge.Retrieve("ch_123", p)
 c.Customer.ID    // ID is still available
 c.Customer.Name  // Name is now also available (if it had a value)
 ```
+### How to use undocumented parameters and properties
 
+stripe-go is a typed library and it supports all public properties or parameters.
+
+Stripe sometimes has beta which introduces new properties or parameters that are not immediately public. The library does not support these properties or parameters until they are public but there is still an approach that allows you to use them.
```
dcr-stripe, line 377:
```suggestion
Stripe sometimes launches private beta features which introduce new properties or parameters that are not immediately public. These will not have typed accessors in the stripe-go library but can still be used. 
```

dcr-stripe: This will be changing as we introduce support for betas.

README.md
```diff
@@ -370,7 +370,47 @@ c, _ := charge.Retrieve("ch_123", p)
...
+	params.AddExtra("secret_feature_enabled", "true")
+	params.AddExtra("secret_parameter[primary]","primary value")
+	params.AddExtra("secret_parameter[secondary]","secondary value")
+
+	customer, err := customer.Create(params)
+```
+
+#### Properties
+
+To retrieve undocumented properties from Stripe using Go you can use an option in the library to return the raw JSON object and return the property. An example of this is shown below:
```
dcr-stripe, line 398:
```suggestion
You can access undocumented properties returned by Stripe by querying the raw response JSON object. An example of this is shown below:
```

dcr-stripe (commented):
LG overall, some minor comments/suggestisons.

PTAL @rmanzer-stripe

README.md
rmanzer-stripe: This makes sense and in light of those coming changes the revised language seems much more informative.

rmanzer-stripe (commented):
Updating README per @dcr-stripe's recommendations

dcr-stripe (approved):

PR #1494, opened 2022-07-12
API Updates

yejia-stripe:
Codegen for openapi v164.
r? @dcr-stripe
cc @stripe/api-libraries

## Changelog
* Add support for `CustomerDetails` on `CheckoutSessionListParams`


dcr-stripe (approved):
LGTM

PR #1495, opened 2022-07-13
feat!: Return an error by default when Webhook API version does not match stripe-go API version.

dcr-stripe:
r? @kamil-stripe @yejia-stripe 
cc @stripe/api-libraries

## Changelog
* ⚠️ Updates the Webhook `ConstructEvent`, `ConstructEventIgnoringTolerance` and `ConstructEventWithTolerance` functions to return an error when the webhook event's API version does not match the stripe-go library API version. 
  * This prevents potential errors where the webhook API version does not match the stripe-go API version, thus deserializing in an incorrect state. This is consistent with stripe-java and stripe-dotnet.
  * You can revert to the original behavior by switching to `ConstructEventWithOptions` and setting `ignoreAPIVersionMismatch: false`.
    * If you wish to opt-out and previously used `ConstructEventIgnoringTolerance`, this would now become:
    ```
    ConstructEventWithOptions(payload, header, secret, ConstructEventOptions{ignoreTolerance: true, ignoreAPIVersionMismatch: false})
     ```
    * If you wish to opt-out and previously used `ConstructEventWithTolerance`, this would now become:
    ```
    ConstructEventWithOptions(payload, header, secret, ConstructEventOptions{tolerance: tolerance, ignoreAPIVersionMismatch: false})
     ```

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -39,6 +39,9 @@ const (
 	// APIURL is the URL of the API service backend.
 	APIURL string = "https://api.stripe.com"
 
+	// ClientVersion is the version of the stripe-go library being used.
+	ClientVersion string = clientversion
+
 	// ConnectURL is the URL for OAuth.
 	ConnectURL string = "https://connect.stripe.com"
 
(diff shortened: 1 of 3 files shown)


kamil-stripe (approved):

yejia-stripe (approved):
Lgtm! Thanks for the detailed changelog + docstrings

PR #1520, opened 2022-08-05
Add GenerateTestSignedPayload to test webhook signing

### LINKED ISSUE #1406: Webhooks testing
How can I mock the requests from Stripe to my webhooks?

The Node library has the `stripe.webhooks.generateTestHeaderString` method that allows to mock such requests:

```javascript
const payload = {
  id: 'evt_test_webhook',
  object: 'event',
};

const payloadString = JSON.stringify(payload, null, 2);
const secret = 'whsec_test_secret';

const header = stripe.webhooks.generateTestHeaderString({
  payload: payloadString,
  secret,
});

const event = stripe.webhooks.constructEvent(payloadString, header, secret);

// Do something with mocked signed event
expect(event.id).to.equal(payload.id);
```

However, I do not see such method on the `webhook` package.

Is there any way we can achieve the same with the Go library?

yejia-stripe:
r? @dcr-stripe 
cc @stripe/api-libraries

## Summary
Add `GenerateTestSignedPayload` function for testing webhook signing, similar to stripe-node's [`webhooks.generateTestHeaderString`](https://github.com/stripe/stripe-node/blob/master/lib/Webhooks.js#L52).

Closes #1406


README.md
```diff
@@ -408,9 +408,35 @@ secret_feature_enabled, _ := string(rawData["secret_feature_enabled"].(bool))
...
+testSecret := "whsec_test_secret"
+
+payloadBytes, err := json.Marshal(payload)
+
+signedPayload := GenerateTestSignedPayload(&UnsignedPayload{payload: payloadBytes, secret: testSecret})
+event, err := ConstructEvent(signedPayload.payload, signedPayload.header, signedPayload.secret)
+
+// Do something with mocked signed event
+if event.ID != payload.id {
+	fmt.Errorf(":(")
```
dcr-stripe, line 436: :)

I I wonder if we should make this more clear... maybe something like:

```
if event.ID = payload.id {
  // Do something with the mocked signed event
} else {
  // Handle invalid event payload. 
}
```

dcr-stripe (approved):
LGTM overall, thanks Yejia! Just one question about the readme sample.

dcr-stripe (approved):
LGTM!

PR #1529, opened 2022-08-19
Allow setting APIVersion

pakrym-stripe:
Allows APIVersion value to be modified. Trims beta headers before using them in WebHook event version check.

r? @dcr-stripe


dcr-stripe (approved):
LGTM! Nice!

PR #1532, opened 2022-08-23
API Updates

dcr-stripe:
Codegen for openapi v184.
r? @pakrym-stripe
cc @stripe/api-libraries

## Changelog
* Change type of `TreasuryOutboundTransferDestinationPaymentMethod` from `string` to `nullable(string)`
* Change return type of `FundCashBalance` method on `Customer` from `Customer` to `CustomerCashBalanceTransaction`
  * This is technically a breaking change, but this return type was actually incorrect and so the result of this method did not deserialize correctly.
* Change return type of `RetrieveFeatures` and `UpdateFeatures` methods on `TreasuryFinancialAccount` from `TreasuryFinancialAccount` to `TreasuryFinancialAccountFeatures`
  * This is technically a breaking change, but this return type was actually incorrect and so the result of this method did not deserialize correctly.


pakrym-stripe:
The changelog entry is wrong. Why are we not seeing the CashBalanceTransaction change here?

pakrym-stripe (changes requested):
The diff seems to be missing the cashBalanceTransaction change.

dcr-stripe:
@pakrym-stripe PTAL - updated changelog and regenerated.

pakrym-stripe (approved):

PR #1533, opened 2022-08-24
Update README.md to clarify that API version can only be change in beta

pakrym-stripe:
r? @dcr-stripe


dcr-stripe (changes requested):
Let's use this formatting across all 3 prs: https://github.com/community/community/discussions/16925

You can do something like:

```
> **Note**
> Test test test
```

to have it render as the following and be more clear:

> **Note**
> Test test test

dcr-stripe (approved):
LGTM - thanks!

PR #1562, opened 2022-10-22
add missing verify with micro-deposits next action

epelc:
The next action type documented a `verify_with_microdeposits` value but did not have a constant for it.

Add `PaymentIntentNextActionTypeVerifyWithMicrodeposits` as a constant for this value.


epelc:
Also proof it is an actual value. From our stripe webhooks

image

pakrym-stripe:
Sorry for the delayed response and thank you for the contribution. This looks great. We found a bug in the generator that caused us to skip this value I pushed another commit that adds more missing values.

PR #1564, opened 2022-11-01
API Updates for beta branch

kamil-stripe:
Codegen for openapi v205.
cc @stripe/api-libraries

## Changelog
* Updated stable APIs to the latest version


account.go
```diff
@@ -265,6 +265,12 @@ type AccountCapabilitiesCartesBancairesPaymentsParams struct {
 	Requested *bool `form:"requested"`
 }
 
+// The cashapp_payments capability.
+type AccountCapabilitiesCashappPaymentsParams struct {
```
pakrym-stripe, line 269: is Cashapp casing correct here?

pakrym-stripe:
The CI is on fire.

account.go
kamil-stripe: Confirmed it is

pakrym-stripe (approved):

PR #1570, opened 2022-11-09
Update sdk-release/next-major with latest changes

pakrym-stripe:


anniel-stripe (approved):

krzysiek1507:
Hey @anniel-stripe. Any chance to split the removal and addition of the fields like `Charges` and `LatestCharge` into 2 releases? One with the addition and the second with removal? It would be great to have an SDK version that works with both so the migration is easier.

PR #1575, opened 2022-11-22
Add version to go reference path

anniel-stripe:
The link to the Go reference docs https://pkg.go.dev/github.com/stripe/stripe-go/ shows v70 by default, instead of the latest version. This fixes the link to be https://pkg.go.dev/github.com/stripe/stripe-go/v74 (verified that running `make update-version` updates this README link with the major version).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Go Stripe
-[![Go Reference](https://pkg.go.dev/badge/github.com/stripe/stripe-go)](https://pkg.go.dev/github.com/stripe/stripe-go)
+[![Go Reference](https://pkg.go.dev/badge/github.com/stripe/stripe-go)](https://pkg.go.dev/github.com/stripe/stripe-go/v74)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 # Go Stripe
-[![Go Reference](https://pkg.go.dev/badge/github.com/stripe/stripe-go)](https://pkg.go.dev/github.com/stripe/stripe-go)
+[![Go Reference](https://pkg.go.dev/badge/github.com/stripe/stripe-go)](https://pkg.go.dev/github.com/stripe/stripe-go/v74)
```
pakrym-stripe, line 2: Should we add this to [the Makefile target](https://github.com/stripe/stripe-go/blob/master/Makefile#L37) ?

pakrym-stripe (changes requested):
Let's add some logic to the Makefile to automatically update this version.

README.md
anniel-stripe: I thought [this line in the Makefile](https://github.com/stripe/stripe-go/blob/master/Makefile#L37) already updates the version. For example, when I run `make update-version VERSION="73.0.0"`, it updates the path to `https://pkg.go.dev/github.com/stripe/stripe-go/v73`.

pakrym-stripe (approved):
🤦

PR #1586, opened 2022-12-22
API Updates

richardm-stripe:
Codegen for openapi v216.
cc @stripe/api-libraries

## Changelog
* Add support for `UsingMerchantDefault` on `CashBalanceSettings`
* Change type of `CheckoutSessionCancelUrl` from `string` to `nullable(string)`


pakrym-stripe:
Which go version do you have? Looks like we still have formatting going back and fourth depending on who generates.

pakrym-stripe (approved):

richardm-stripe:
```
$ go version
go version go1.19.2 darwin/arm64
```

Issue #1193, opened 2020-09-29
Create A Mock Test Client


meggieveggie:
## Proposed Feature

After implementing the Golang SDK there could possibly be a better test suite that can be created. For instance the current issue I am having is mocking a response from:

```go
stripeClient.Invoices.List(params)
```

My thinking here would be to approach this how kubernetes has with their [fake client](https://godoc.org/k8s.io/client-go/kubernetes/fake). I understand there are many ways to mock out responses (This could definitely be documented better). However it seems that having the ability to do something along the lines of:

```go
stripeClient := &stripeFakeClient{}

stripeClient.Invoices.Create(*Invoices{//Some Random Invoice Data})

resp := stripeClient.Invoices.List(params)
```

without actually calling the API would make integration with your platform a lot more seamless and speed up the process.

remi-stripe:
@meggieveggie Thanks for the feedback. We usually recommend using stripe-mock for something like this which is what we use in our test suite. You can read more about this here: https://github.com/stripe/stripe-mock/ 
Would that work for you?

meggieveggie:
@remi-stripe My understanding of stripe-mock is that it is specifically used for integration tests and not unit tests. If this is not the case then I will try and use stripe-mock

ianjabour-stripe:
@meggieveggie thank you for this suggestion. You are correct in that stripe-mock is built mainly for integration tests, but it should still allow you to quickly test integration without actually hitting the Stripe API.

The limitation with stripe-mock is that it is stateless, so fetching objects you create in a test will not work. Also, the data made available is limited to that which we generate based off resource fixtures from within our API.

We have been experimenting with ways around these limitations, but don't have anything readily available right now.

saniales:
to everybody still searching for a solution without stripe mock

I managed to mock without it in the way I explained [here on Stack Overflow](https://stackoverflow.com/a/71073181/2349872)

xis:
> to everybody still searching for a solution for unit tests without stripe mock
> 
> I managed to mock without it in the way I explained [here on Stack Overflow](https://stackoverflow.com/a/71073181/2349872)

I'm using this method in unit tests too but can't mock the list iterators like
```go
invoices := s.client.Invoices.List(params)

for invoices.Next() { // can't dive here
    i := invoices.Invoice()
    log.Println(i.Currency)
}
```

guptaaashutosh:
[...]

PR #1603, opened 2023-02-08
Add an example of client mocking

pakrym-stripe:
Fixes: https://github.com/stripe/stripe-go/issues/1193


anniel-stripe (approved):
What's up with the lint failure?

PR #1607, opened 2023-02-09
fix: deterministic encoding

### LINKED ISSUE #1606: Encode payloads deterministically
### Is your feature request related to a problem? Please describe.

Instead of mocking complex use-cases, we want to speed up our test pipeline by using https://github.com/google/go-replayers through using a custom `*http.Client`. The problem is now that the payloads are not deterministic with the custom form encoder included in this SDK. Specifically map types, as used for metadata, are not encoded deterministic and therefore mess with the replayer:

```
metadata[interval]=monthly&metadata[plan]=Essentials%400&customer=cus_NKH10bl....
metadata[plan]=Essentials%400&metadata[interval]=monthly&customer=cus_NKH10bl....
```

### Describe the solution you'd like

The payload should be encoded deterministically. This could either be achieved by

1. sorting keys
2. using the stdlib JSON

### Describe alternatives you've considered

1. sorting keys
2. using the stdlib JSON

### Additional context

_No response_

zepatrik:
Closes #1606 

As described in the issue, currently payloads are not encoded deterministically. This PR is a proposal to fix that.

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -338,16 +339,25 @@ func isAppender(t reflect.Type) bool {
 }
 
 func mapEncoder(values *Values, v reflect.Value, keyParts []string, _ bool, _ *formOptions) {
+	keys := make([]string, 0, v.Len())
 	for _, keyVal := range v.MapKeys() {
-		if Strict && keyVal.Kind() != reflect.String {
-			panic("Don't support serializing maps with non-string keys")
+		if keyVal.Kind() != reflect.String {
+			if Strict {
+				panic("Don't support serializing maps with non-string keys")
+			} else {
+				// otherwise keyVal.String() will panic later
+				continue
+			}
 		}
-
+		keys = append(keys, keyVal.String())
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
 		// Unlike a property on a struct which will contain a zero value even
 		// if never set, any value found in a map has been explicitly set, so
 		// we always make an effort to encode them, even if a zero value
 		// (that's why we pass through `true` here).
-		reflectValue(values, v.MapIndex(keyVal), true, append(keyParts, keyVal.String()))
+		reflectValue(values, v.MapIndex(reflect.ValueOf(key)), true, append(keyParts, key))
 	}
 }
 
--- a/form/form.go
+++ b/form/form.go
@@ -344,10 +344,9 @@ func mapEncoder(values *Values, v reflect.Value, keyParts []string, _ bool, _ *f
 		if keyVal.Kind() != reflect.String {
 			if Strict {
 				panic("Don't support serializing maps with non-string keys")
-			} else {
-				// otherwise keyVal.String() will panic later
-				continue
 			}
+			// otherwise keyVal.String() will panic later
+			continue
 		}
 		keys = append(keys, keyVal.String())
 	}
--- a/form/form_test.go
+++ b/form/form_test.go
@@ -379,6 +379,43 @@ func TestAppendToPrefixed(t *testing.T) {
 	assert.Equal(t, []string{"foo"}, form.Get("prefix[string]"))
 }
 
+func TestEncode(t *testing.T) {
+	form := &Values{}
+	form.Add("foo", "bar")
+	form.Add("foo[bar]", "baz")
+	assert.Equal(t, "foo=bar&foo[bar]=baz", form.Encode())
+}
+
+func TestEncodeDeterministically(t *testing.T) {
+	data := map[string]string{
+		"first":  "foo",
+		"second": "bar",
+	}
+	initialForm := &Values{}
+	AppendTo(initialForm, data)
+	encoded := initialForm.Encode()
+
+	for i := 0; i < 100; i++ {
+		form := &Values{}
+		AppendTo(form, data)
+		assert.Equal(t, encoded, form.Encode(), "iteration %d", i)
+	}
+}
+
+func TestEncodeMapNonStringKey(t *testing.T) {
+	// Disable strict mode for this test, so the non-string key is ignored.
+	Strict = false
+	defer func() {
+		Strict = true
+	}()
+
+	form := &Values{}
+	assert.NotPanics(t, func() {
+		AppendTo(form, map[int]string{1: "foo"})
+	})
+	assert.Len(t, form.values, 0)
+}
+
 func TestFormatKey(t *testing.T) {
 	assert.Equal(t, "param", FormatKey([]string{"param"}))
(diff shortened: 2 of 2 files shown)


form/form.go
```diff
@@ -541,19 +540,11 @@ func (f *Values) Add(key, val string) {
 // Encode encodes the keys and values into “URL encoded” form
 // ("bar=baz&foo=quux").
 func (f *Values) Encode() string {
-	var buf bytes.Buffer
```
pakrym-stripe, line 544: Do you know where the indeterminism is coming from in the existing implementation? It looks like it should've been deterministic as is.

zepatrik: When values are added using `AddTo`, and those values are maps, they are read in random order (as per go spec).

form/form_test.go
```diff
@@ -379,6 +379,29 @@ func TestAppendToPrefixed(t *testing.T) {
...
+		assert.Equal(t, encoded, form.Encode(), "iteration %d", i)
+	}
+}
+
+func TestEncode(t *testing.T) {
+	form := &Values{}
+	form.Add("foo", "bar")
+	form.Add("foo[bar]", "baz")
+	assert.Equal(t, "foo=bar&foo[bar]=baz", form.Encode())
+}
```
zepatrik, line 403: Test coverage was low on this one. Because of this, we cannot use `url.Values` directly.

form/form.go
pakrym-stripe: I see, thank you for the explanation.

form/form.go
```diff
@@ -542,6 +543,7 @@ func (f *Values) Add(key, val string) {
 // ("bar=baz&foo=quux").
 func (f *Values) Encode() string {
 	var buf bytes.Buffer
+	sort.Sort(f.values)
```
pakrym-stripe, line 546: I feel that having everyone pay the cost of sorting in production is a bad tradeoff to enable a particular testing scenario.

zepatrik: With regard to stripe API latency, I guess sorting a few keys, where most of them are already in the correct order, is of least concern for most users :sweat_smile: 
But your call, I guess the only way to make this a configurable feature will be to implement a custom backend that does deterministic encoding.

zepatrik: To clarify: most keys will be in order already because most of the time structs are encoded. As these are auto generated with sorted keys already, and reflection walks the struct keys in the order they are specified, this will not affect most fields.

pakrym-stripe: > most keys will be in order already

they'll be in order of addition but not sorted. A lot of the time the entire set will get resorted.

pakrym-stripe:
I'm sorry. I don't think we'll take this change. The extra operation is completely unnecessary for production client and in test scenarios the same can be achieved by re-normalizing the body in HttpClient adapter.

zepatrik:
I assumed this is a good idea in general, as even the go standard library, which is usually very focused on the bare minimum of functionality, is doing this for JSON and URL query encoding.
I can revert all changes and only contribute the missing encoding test if you prefer that.

pakrym-stripe:
> is doing this for JSON and URL query encoding

But all they do is preserve the order of insertion, which we do too, to the extend allowed by the standard library. My understanding is that go doesn't sort keys for JSON or query encoding.

>I can revert all changes and only contribute the missing encoding test if you prefer that.

I'll happily merge the test!

zepatrik:
Have a look at https://github.com/golang/go/issues/27050 and https://github.com/golang/go/issues/29985. The `url.Values` sort all keys: https://github.com/golang/go/blob/fb79da299127b3ff85e14f37e4616a26e6c2a720/src/net/url/url.go#L977
We cannot use the standard lib to encode the values here though because of the escaping of `[` and `]`.

zepatrik:
As said, the current approach preserves the order of addition. We could also fix this so that map keys are sorted before being added, but I am not sure whether that will be easy to implement. I can give it a try it you would be fine with that approach.

zepatrik:
> Does it? I thought it resorts all the entries. Isn't that why you had to change [these tests](https://github.com/stripe/stripe-go/pull/1607/files#diff-fbc9e0365993b954234531ddb2c19db9889c949802934c60b71a45de4232e845R490)?

Sorry, I meant the current version on master, not my PR branch.
I'll take a look at inserting map keys ordered later 👍

form/form.go
```diff
@@ -338,16 +339,25 @@ func isAppender(t reflect.Type) bool {
 }
 
 func mapEncoder(values *Values, v reflect.Value, keyParts []string, _ bool, _ *formOptions) {
+	keys := make([]string, 0, v.Len())
```
zepatrik, line 342: Here we now collect and sort keys before adding them to the `values`.

form/form.go
```diff
@@ -338,16 +339,25 @@ func isAppender(t reflect.Type) bool {
...
-
+		keys = append(keys, keyVal.String())
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
 		// Unlike a property on a struct which will contain a zero value even
 		// if never set, any value found in a map has been explicitly set, so
 		// we always make an effort to encode them, even if a zero value
 		// (that's why we pass through `true` here).
-		reflectValue(values, v.MapIndex(keyVal), true, append(keyParts, keyVal.String()))
```
zepatrik, line 350: This was problematic as well because non-string values would always panic, regardless of strict mode. In case of strict mode it would obviously panic with the explicit panic above, but in the non-strict mode it would still panic as `keyVal.String()` would panic.
I added a test for ensuring non-string keys are just ignored.

zepatrik (commented):
Hey @pakrym-stripe, I did now change the approach to sorting map keys only! Here are some explanations for what I did.

pakrym-stripe (approved):

form/form.go
pakrym-stripe: Hm, I tried this test on master and it didn't panic:

```go

func TestEncodeMapNonStringKey(t *testing.T) {
	// Disable strict mode for this test, so the non-string key is ignored.
	Strict = false
	defer func() {
		Strict = true
	}()

	form := &Values{}
	assert.NotPanics(t, func() {
		AppendTo(form, map[int]string{1: "foo"})
	})
	assert.Equal(t, "%3Cint+Value%3E=foo", form.Encode())
}

```

Granted the result is not as useful but we neither had to ignore or panic. What panics in you version is `v.MapIndex(reflect.ValueOf(key))` because the string key is being provided to an int-based map.

zepatrik: Ah sorry, you are correct.

> ```go
> // String is a special case because of Go's String method convention.
> // Unlike the other getters, it does not panic if v's Kind is not String.
> ```

I don't know what behavior you prefer at this point, happy to adjust for either.

pakrym-stripe:
Thank you for the contribution @zepatrik ! And thank you for going through all the iterations to find a solution.

zepatrik:
Thanks for the quick review :+1:

PR #1611, opened 2023-02-24
Run goimports on generated test suite

richardm-stripe:
Our generator used to run `goimports` on the generated test suite file separately from calling `make codegen-format`. This is no longer the case, and the past few releases have required manually running goimports on this file. This PR causes `make codegen-format` to run goimports on this file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,5 +40,6 @@ update-version:
 
 codegen-format:
 	go fmt ./...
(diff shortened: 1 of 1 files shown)


pakrym-stripe (approved):

pakrym-stripe:
My bad. I broke this and didn't notice 🤦

PR #1625, opened 2023-03-17
Use shared stripe-mock action

pakrym-stripe:


.github/workflows/ci.yml
```diff
@@ -54,12 +54,24 @@ jobs:
...
-              make check-api-clients
+         run: make ci-test
        - name: Coveralls
          run: make coverage && make coveralls
          if: matrix.go == '1.16'
          env:
            COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
            COVERALLS_FLAG_NAME: Go-${{ matrix.go }}
+
+  publish:
```
pakrym-stripe, line 65: Added publish target so we get a release notification even though there isn't actually any package push happening.

anniel-stripe (approved):

PR #1646, opened 2023-04-21
Add support for linked to account on card holder

egag:


pakrym-stripe:
Hi @egag, thank you for your contribution! Unfortunately `linked_to_account` is a beta feature so we can't add it to the GA SDK just yet.

PR #1648, opened 2023-04-22
Introduce stripe.RawRequest as a canonical way to request APIs without definitions

richardm-stripe:
Adds a `stripe.RawRequest` method to be able to send untyped requests while still offering core library features like auth, retries, etc.


params.go
```diff
@@ -246,6 +251,29 @@ type ParamsContainer interface {
 	GetParams() *Params
 }
 
+type RawParams struct {
+	Params        `form:"*"`
+	APIMode       APIMode `form:"-"`
+	StripeContext string  `form:"-"`
+}
+
+func (p *RawParams) GetAPIMode() APIMode {
```
pakrym-stripe, line 260: Do we need these functions considering fields are public?

preview/client.go
```diff
@@ -0,0 +1,27 @@
+// Package preview provides preview / private beta APIs
+package preview
+
+import (
+	"net/http"
+
+	stripe "github.com/stripe/stripe-go/v74"
+)
+
+func GetDefaultRequestOptions(params stripe.RawParamsContainer) stripe.RawParamsContainer {
```
pakrym-stripe, line 10:
```suggestion
func getDefaultRequestOptions(params stripe.RawParamsContainer) stripe.RawParamsContainer {
```

preview/client.go
```diff
@@ -0,0 +1,27 @@
...
+		StripeContext: params.GetStripeContext(),
+	}
+	return &rawParams
+}
+
+func Get(path string, params stripe.RawParamsContainer) (*stripe.APIResponse, error) {
+	return stripe.RawRequest(http.MethodGet, path, "", GetDefaultRequestOptions(params))
+}
+func Post(path, content string, params stripe.RawParamsContainer) (*stripe.APIResponse, error) {
+	return stripe.RawRequest(http.MethodPost, path, content, params)
```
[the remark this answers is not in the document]
pakrym-stripe: Maybe a few simple tests to verify that API mode is set

stripe.go
```diff
@@ -275,7 +279,22 @@ type BackendImplementation struct {
 	requestMetricsBuffer chan requestMetrics
 }
 
-func extractParams(params ParamsContainer) (*form.Values, *Params) {
+func extractParamJSON(params ParamsContainer) ([]byte, *Params, error) {
```
pakrym-stripe, line 282: Do we still need this?

README.md
```diff
@@ -529,6 +529,22 @@ To install a beta version of stripe-go use the commit notation of the `go get` c
 go get -u github.com/stripe/stripe-go/v74@v74.3.0-beta.1
 ```
 
+### Custom SDKs
+If you would like to send a request to an undocumented API (for example you are in a private beta), or if you prefer to bypass the method definitions in the library and specify your request details directly, you can use the `RawRequest` method on the `stripe` backend.
+
+```go
+key := "sk_test_xyz" // set to "" to use stripe.Key
+content := `foo=myFoo&bar[]=true` // form-encoded POST body
```
anniel-stripe, line 537: Show example of how to encode

params.go
```diff
@@ -246,6 +251,29 @@ type ParamsContainer interface {
...
+		return StandardAPIMode
+	}
+	return p.APIMode
+}
+
+func (p *RawParams) GetStripeContext() string {
+	return p.StripeContext
+}
+
+type RawParamsContainer interface {
```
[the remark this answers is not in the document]
anniel-stripe: We needed it before to allow users to add request params, but you're right this is no longer necessary since we're switching to the `content` string implementation.

stripe.go
anniel-stripe: Nope, reverting

params.go
anniel-stripe: Good point, we don't

README.md
pakrym-stripe: We have two `content :=` assignments in the example. Can `form.AppendTo` consume a simple map?

stripe.go
```diff
@@ -376,6 +442,71 @@ func (s *BackendImplementation) CallRaw(method, path, key string, form *form.Val
 	return nil
 }
 
+// func newRequestHeader(method, key, contentType string, params *Params) (http.Header, context.Context, error) {
```
pakrym-stripe, line 445: nit: rem?

stripe.go
```diff
@@ -350,14 +354,76 @@ func (s *BackendImplementation) CallMultipart(method, path, key, boundary string
...
+	var contentType string
+	if method != http.MethodPost && method != http.MethodGet && method != http.MethodDelete {
+		return nil, fmt.Errorf("method must be POST, GET, or DELETE. Received %s", method)
+	}
+
+	paramsIsNil := params == nil || reflect.ValueOf(params).IsNil()
+
+	if paramsIsNil {
+		_, commonParams = extractParams(params)
+		contentType = "application/x-www-form-urlencoded"
```
pakrym-stripe, line 371: Should we be setting this value even for preview mode?

README.md
anniel-stripe: Yes, updated README in latest commit

pakrym-stripe:
Do we need developer_error handling as well?

stripe.go
anniel-stripe: oops, thank you

anniel-stripe:
> Do we need developer_error handling as well?
I was thinking of doing that in a separate PR, but happy to put it in this one if you don't reviewing again.

pakrym-stripe (approved):

PR #1663, opened 2023-05-24
Handle developer message in preview error responses

anniel-stripe:
The error message can be returned in a developer_message field, add this to the `Error` type.

### DIFF
--- a/error.go
+++ b/error.go
@@ -232,6 +232,7 @@ type Error struct {
 
 	HTTPStatusCode    int               `json:"status,omitempty"`
 	Msg               string            `json:"message"`
+	DeveloperMsg      string            `json:"developer_message"`
 	Param             string            `json:"param,omitempty"`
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -232,6 +232,7 @@ type Error struct {
 
 	HTTPStatusCode    int               `json:"status,omitempty"`
 	Msg               string            `json:"message"`
```
anniel-stripe, line 234: We might want to add `omitempty` to `Msg` too - right now the error will look like
```
{
  "status":400,
  "message":"",
  "developer_message":"Some fields in the request were invalid.",
  "request_id":"req_...",
  "type":""}
```

pakrym-stripe: If it doesn't affect deserialization let's leave it as is.

pakrym-stripe (approved):

PR #1719, opened 2023-08-24
Move beta version back

anniel-stripe:
Bumped to fix CI, but we need to move it back before releasing. Analogous to https://github.com/stripe/stripe-go/commit/12cb5d6a66adf3810883a62645f9d3b099bcabb4


pakrym-stripe:
Hm, in theory we didn't need bumping after the recent fix.

pakrym-stripe:
Linked PR only reverted the VERSION file, not all the files.

pakrym-stripe (changes requested):
Reverting `VERSION` file only should be enough.

pakrym-stripe (approved):

Issue #1728, opened 2023-09-07
initUserAgent is slow


egonelbre:
### Describe the bug

initUserAgent is relatively slow (100ms) and is run always, even when stripe client is not being used.

https://github.com/stripe/stripe-go/blob/d038ebbf7b3100fb94bfb4ac21874a5dbfaba25e/stripe.go#L1357

### To Reproduce

```
func BenchmarkInitUserAgent(b *testing.B) {
	for i := 0; i < b.N; i++ {
		initUserAgent()
	}
}
```

```
$ go test -bench InitUserAgent -run "$^"
goos: windows
goarch: amd64
pkg: github.com/stripe/stripe-go/v75
cpu: AMD Ryzen Threadripper 2950X 16-Core Processor
BenchmarkInitUserAgent-32             12          92528067 ns/op
```

Note, the performance on Mac/Linux can be different, because the os exec overhead is smaller.

### Expected behavior

initUserAgent is done on-demand / lazily rather than in `init`.

### OS

Windows 11

### Go version

Go 1.21.1

### stripe-go version

master

### API version

not-relevant

### Additional context

This slows down tests unrelated to stripe.

pakrym-stripe:
The fix will ship next week.

PR #1731, opened 2023-09-07
Delay calculation of Stripe-User-Agent

pakrym-stripe:
Fixes: https://github.com/stripe/stripe-go/issues/1728

```

BenchmarkInitUserAgent-10    	95851929	        12.20 ns/op
```

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -417,7 +417,7 @@ func (s *BackendImplementation) NewRequest(method, path, key, contentType string
 	req.Header.Add("Content-Type", contentType)
 	req.Header.Add("Stripe-Version", APIVersion)
 	req.Header.Add("User-Agent", encodedUserAgent)
-	req.Header.Add("X-Stripe-Client-User-Agent", encodedStripeUserAgent)
+	req.Header.Add("X-Stripe-Client-User-Agent", getEncodedStripeUserAgent())
 
 	if params != nil {
 		if params.Context != nil {
@@ -1278,6 +1278,7 @@ type requestTelemetry struct {
 var appInfo *AppInfo
 var backends Backends
 var encodedStripeUserAgent string
+var encodedStripeUserAgentReady *sync.Once
 var encodedUserAgent string
 
 // The default HTTP client used for communication with any of Stripe's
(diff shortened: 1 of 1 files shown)


richardm-stripe (approved):
LGTM, consider adding a test though?

pakrym-stripe:
Exercised in `TestUserAgentWithAppInfo`

stripe.go
```diff
@@ -1278,6 +1278,7 @@ type requestTelemetry struct {
 var appInfo *AppInfo
 var backends Backends
 var encodedStripeUserAgent string
+var encodedStripeUserAgentReady *sync.Once
```
egonelbre, line 1281: Side-by code review :). This doesn't need to be a pointer. It can directly use `encodedStripeUserAgentReady sync.Once` and won't need initialization.

stripe.go
```diff
@@ -1359,22 +1360,28 @@ func initUserAgent() {
...
-	}
-	marshaled, err := json.Marshal(stripeUserAgent)
-	// Encoding this struct should never be a problem, so we're okay to panic
-	// in case it is for some reason.
-	if err != nil {
-		panic(err)
-	}
-	encodedStripeUserAgent = string(marshaled)
+func getEncodedStripeUserAgent() string {
+	encodedStripeUserAgentReady.Do(func() {
```
egonelbre, line 1367: This way of caching would make `SetAppInfo` not update the stripeUserAgent multiple times. Although, I guess it can be fine when the assumption is that `SetAppInfo` is always called before using the client.

pakrym-stripe: We reset the `encodedStripeUserAgentReady` in `initUserAgent` which is called from `SetAppInfo`. So calling `SetAppInfo` will force the StripeUserAgent calculation to rerun.

egonelbre: ah, yes, then it's fine

stripe.go
pakrym-stripe: We would still need initialization to support multiple `SetAppInfo` calls. You are right that the pointer isn't required.


Thank you for the review :)

PR #1748, opened 2023-10-11
add NewBackendsWithConfig helper

### LINKED ISSUE #1747: Better api for creating Backends with config
### Is your feature request related to a problem? Please describe.

It is very difficult to setup a custom client with your own config for logging. You can set the global default logger but a per client one is difficult to configure.

### Describe the solution you'd like

An api to create a client with custom [BackendConfig](https://pkg.go.dev/github.com/stripe/stripe-go/v75#BackendConfig).

ie 

```
stripe.NewBackendsWithConfig(*BackendConfig) (*Backends, error)
```

### Describe alternatives you've considered

[NewBackends](https://pkg.go.dev/github.com/stripe/stripe-go/v75#NewBackends) is only recommended for testing and also  is difficult to use. There doesn't appear to be an api to update backend's config without using globals. You also have to dig into the backend's individual fields to update multiple configs.

It should be easy to set a single `BackendConfig` on an entire backend.

Also if you try to work around this by manually setting up `stripe.Backends` instead of using `NewBackends` then there is no guarantee that you setup all of the backend fields like API, Connect, etc. It would be nice to have a `NewBackendsWithConfig(*BackendConfig)` api.

```go
stripeBackendConf := &stripe.BackendConfig{
	// custom backend config
}
backends := &stripe.Backends{
	// You have to make sure you fill out each of these which is cumbersome and requires keeping up with stripe lib changes
	API:     stripe.GetBackendWithConfig(stripe.APIBackend, stripeBackendConf),
	Connect: stripe.GetBackendWithConfig(stripe.ConnectBackend, stripeBackendConf),
	Uploads: stripe.GetBackendWithConfig(stripe.UploadsBackend, stripeBackendConf),
}

client.New(stripeKey, backends)

```

### Additional context

_No response_

epelc:
Allow creating backends with a single config for all backends easily.

Closes #1747 

Makes it very simple to setup a client with custom logging instead of using globals.

```
conf := &stripe.BackendConfig{
  // setup logger and http client
}
backends := stripe.NewBackendsWithConfig(conf)

cl := client.NewClient(stripeKey, backends)
```

The library should make it easier to use instanced clients like this. It's much easier to test when you are instantiating your own client. Also it ensures you properly configure everything instead of having to hunt for multiple global funcs to set different config options.


pakrym-stripe:
Thank you for your contribution, @epelc !

stripe.go
```diff
@@ -1129,6 +1128,16 @@ func NewBackends(httpClient *http.Client) *Backends {
 	}
 }
 
+// NewBackends creates a new set of backends with the given config for all backends.
```
pakrym-stripe, line 1131:
```suggestion
// NewBackendsWithConfig creates a new set of backends with the given config for all backends.
```

pakrym-stripe (changes requested):
One tiny change to make the linter happy.

epelc:
@pakrym-stripe thanks! Just committed it.

pakrym-stripe (approved):

PR #1779, opened 2023-12-07
Track usage of `raw_request`

richardm-stripe:
## Summary
- Adds infrastructure for reporting `usage` to `X-Stripe-Client-Telemetry`, without breaking changes to the public interface.
- Reports usage for `raw_request` in the beta branch.

### Release
The first commit [Usage infrastructure](https://github.com/stripe/stripe-go/commit/7527cae42f2f4ed8a0f97300ea83267de7b86cf6) I intend to merge into master in a follow-up, after discussion.

The second commit [Usage for raw_reuest](https://github.com/stripe/stripe-go/commit/4c4da88ee14835faeec4d4dec96c0c61c11155ae) is only applicable to the beta branch. I've opened the PR against beta showing both for a more holistic review.

### Implementation

The tricky part about all this is: we have a struct `BackendImplementation` with public `Do` and `Call` methods (`Call` invokes `Do`). Only the `Do` method has access to `Params` which is where we want to put `usage`, but only the `Call` method has access to `duration`, which also needs to go on the telemetry header. The ideal solution would involve changing the signature of `Do` so that it can shuffle the information back to `Call` in the return value, but that's breaking. To avoid a breaking change, what we instead have to do is pass this information through the `LastResponseSetter`.

`BackendImplementation` although "public" is marked as internal and we say that public use is deprecated since 2018. Things would be much simpler if we [broke it](https://github.com/stripe/stripe-go/compare/beta...richardm-usage-but-breaking?expand=1).

It also turns out, there's not much code sharing in the RawRequest implementation right now, it doesn't go through `backend.Call`, for instance.

### DIFF
--- a/params.go
+++ b/params.go
@@ -207,12 +207,20 @@ type Params struct {
 	// account instead of under the account of the owner of the configured
 	// Stripe key.
 	StripeAccount *string `form:"-"` // Passed as header
+
+	usage []string `form:"-"` // Tracked behaviors
 }
 
 // AddExpand on the Params embedded struct is deprecated.
 // Deprecated: please use Expand in the surrounding struct instead.
 func (p *Params) AddExpand(f string) {
 	p.Expand = append(p.Expand, &f)
+}
+
+// InternalSetUsage sets the usage field on the Params struct.
+// Unstable: for internal stripe-go usage only.
+func (p *Params) InternalSetUsage(usage []string) {
+	p.usage = usage
 }
 
 // AddExtra adds a new arbitrary key-value pair to the request data
--- a/stripe.go
+++ b/stripe.go
@@ -108,6 +108,8 @@ type APIResponse struct {
 
 	// StatusCode is a status code as integer. e.g. 200
 	StatusCode int
+
+	duration *time.Duration
 }
 
 // StreamingAPIResponse encapsulates some common features of a response from the
@@ -122,9 +124,10 @@ type StreamingAPIResponse struct {
 	RequestID      string
 	Status         string
 	StatusCode     int
+	duration       *time.Duration
 }
 
-func newAPIResponse(res *http.Response, resBody []byte) *APIResponse {
+func newAPIResponse(res *http.Response, resBody []byte, requestDuration *time.Duration) *APIResponse {
 	return &APIResponse{
 		Header:         res.Header,
 		IdempotencyKey: res.Header.Get("Idempotency-Key"),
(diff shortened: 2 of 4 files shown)


params.go
```diff
@@ -215,6 +217,12 @@ func (p *Params) AddExpand(f string) {
 	p.Expand = append(p.Expand, &f)
 }
 
+// SetUsage sets the usage field on the Params struct.
+// Unstable: for internal stripe-go usage only.
+func (p *Params) SetUsage(usage []string) {
```
pakrym-stripe, line 222: Consider making this function receiverless and renaming it to something more scary (InternalSetUsage?).

pakrym-stripe: There are no usages of this method.

stripe.go
```diff
@@ -279,6 +284,34 @@ type BackendImplementation struct {
 	requestMetricsBuffer chan requestMetrics
 }
 
+type unmarshalTargeter interface {
```
pakrym-stripe, line 287: Why do we need a new interface? Can't we implement the telemetry callback using the LastResponseSetter interface?

pakrym-stripe:
Should we target master first and then merge the infrastructure change into beta?

stripe.go
richardm-stripe: Tackled in [custom UnmarshalJSON instead of unmarshaltargeter](https://github.com/stripe/stripe-go/pull/1779/commits/e3566667d74928a8685dabb3e5e3aa34d68fdbb5)

stripe.go
```diff
@@ -279,6 +284,33 @@ type BackendImplementation struct {
 	requestMetricsBuffer chan requestMetrics
 }
 
+type lastResponseSetterWrapper struct {
```
pakrym-stripe, line 287: consider specializing these to metric enquing 

metricsResponseSetter

richardm-stripe: Why? Do you mean
```go
type metricsResponseSetter struct {
	LastResponseSetter
	backend *BackendImplementation
        params *Params
}

func (s metricsResponseSetter) SetLastResponse (response *APIResponse) {
	var usage []string
	if params != nil {
		usage = params.usage
	}
	backend.maybeEnqueueTelemetryMetrics(response.RequestID, response.duration, usage)
	s.LastResponseSetter.SetLastResponse(response)
}

func (s *BackendImplementation) CallRaw(method, path, key string, form *form.Values, params *Params, v LastResponseSetter) error {
	...
	s.Do(..., &responseSetter)
	responseSetter := lastResponseSetterWrapper{v, s, params}
}
```
I like reading the metrics-setting logic inline inside `CallRaw`, and more broadly I prefer general to specialized, at least for private methods/structs.

richardm-stripe:
> Should we target master first and then merge the infrastructure change into beta?
Yes. That's available now in https://github.com/stripe/stripe-go/pull/1780

stripe.go
pakrym-stripe: >Why? 
Mostly to get rid of layers of indirection. A callback calling another callback ...

>Do you mean
Yep, that.

pakrym-stripe (approved):

stripe.go
richardm-stripe: I made this change on the infrastructure branch!

PR #1786, opened 2023-12-15
Fix check_gofmt.sh

richardm-stripe:
`check-gofmt` checks for consistency with `gofmt -s` but make codegen-format run `go fmt ./...` which ultimately runs `gofmt` without the `-s` flag.

This PR changes that, plus it simplifies the check_gofmt.sh script.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,6 @@ check-api-clients:
 
 check-gofmt:
 	scripts/check_gofmt.sh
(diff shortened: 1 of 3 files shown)


pakrym-stripe:
Ugh, missed that this PR exists. Done similar fix in https://github.com/stripe/stripe-go/pull/1788

PR #1797, opened 2024-01-10
ChargeOutcome: Add NetworkAdviceCode and NetworkDeclineCode

shawnd:
Adds `NetworkAdviceCode` and `NetworkDeclineCode` to `ChargeOutcome` struct.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -535,6 +535,10 @@ type ChargeOutcomeRule struct {
 
 // Details about whether the payment was accepted, and why. See [understanding declines](https://stripe.com/docs/declines) for details.
 type ChargeOutcome struct {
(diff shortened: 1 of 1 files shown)


remi-stripe:
@shawnd The feature you're using that provides those properties is in private beta and isn't part of our public API. That's why it isn't in our stripe-go SDK (or any other SDK) and it won't be until the day this feature is released publicly and can be used by all developers.
This means we won't be able to merge this change for now unfortunately but it would be added to all SDKs if and when this is released publicly. In the meantime you can use [this approach](https://github.com/stripe/stripe-go/?tab=readme-ov-file#properties) to access those properties in your code.

PR #1836, opened 2024-03-27
Change the status on TerminalReader object from string to enum

ramya-stripe:
On Jan 31st, 2024 the API for Terminal Reader was changed as below

[Change type of Terminal.Reader.status from string to enum('offline'|'online')](https://docs.stripe.com/changelog#january-31,-2024)

We added overrides to suppress this change to not break users.
This PR removes that override


ramya-stripe:
Covered in #1837

PR #1837, opened 2024-03-27
Remove shared overrides added for backcompat, change TerminalReaderStatus to enum

ramya-stripe:
This PR contains the changes generated after removing the following from our code generator
- the shared overrides 
- the Go specific override to change TerminalReaderStatus from string to enum

TODO - Changelog entry


onsni:
Traore Ousseni

Issue #1894, opened 2024-07-24
GetBackendWithConfig overrides the config argument


Zyko0:
### Describe the bug

Currently, [`NewBackendsWithConfig`](https://github.com/stripe/stripe-go/blob/124246af3a3ba52c22b267248c79a115709ee482/stripe.go#L1186-L1192) does not allow a single config with no [`URL`](https://github.com/stripe/stripe-go/blob/124246af3a3ba52c22b267248c79a115709ee482/stripe.go#L251-L257) field supplied, because [`GetBackendWithConfig`](https://github.com/stripe/stripe-go/blob/124246af3a3ba52c22b267248c79a115709ee482/stripe.go#L1119-L1141) overrides the URL field if left empty.

This means that whenever the `Upload` backend is initialized the URL value can be set with the previously set APIBackend, and since the url differs for file management, Upload backend won't work at all.

I thought that the only use of NewBackendsWithConfig should be with an empty URL, since defining one would obviously apply to all backends, which does not seem to make sense.

### To Reproduce

- Create a BackendConfig without specifying a URL, so that default URL is picked
- Create Backends with config
- Try to use the UploadBackend

### Expected behavior

URL staying empty in the config, therefore filling the appropriate default URL for each respective backend.

### Code snippets

```Go
cfg := &stripe.BackendConfig{
   HTTPClient: myClient,
   LeveledLogger: myLogger,
   URL: nil,
}
backends := stripe.NewBackendsWithConfig(cfg)
// Create a new file => invalid api url
```

### OS

Linux

### Go version

Go 1.21.4

### stripe-go version

v.79

### API version

2024-06-20

### Additional context

[`NewBackends`](https://github.com/stripe/stripe-go/blob/124246af3a3ba52c22b267248c79a115709ee482/stripe.go#L1174-L1180) works because it's given a different config for each backend for example.

PR #1895, opened 2024-07-24
Fixed config override with GetBackendWithConfig

Zyko0:
Fixes https://github.com/stripe/stripe-go/issues/1894

Let me know if you see another way to write the test, or if you prefer another way to handle the config => cfg thing!


helenye-stripe:
Hi @Zyko0. Thanks so much for submitting this fix! Would it be possible to add a test to verify that using UploadBankend will correctly access the uploads API URL? It would be good to see this in use with a mock call to the API.

Zyko0:
Hey @helenye-stripe, I'm not sure to understand how to make it similar to the code you linked.
It seems `testBackend` does not do anything and prevents any URL testing.

As for the test, I just made something locally, but if I'm using `NewBackendsWithConfig` I have to hard-define the URL to the stripe-mock one, or to the originally set UploadBackend which kinda defeats the purpose.

Since this fix is meant to address the case when no URL is supplied 😬 

Not setting the URL in the backend config, makes it set by the stripe sdk to https://files.stripe.com which doesn't work locally (but at least this shows that the upload backend URL is correctly set even if it's set after the API backend with an empty URL in the config)

PR #1909, opened 2024-08-24
Add MsgCode to Error

wozaki:
### Background Summary

The current `Error` struct does not include a field to store the `message_code`. This requires writing additional code to extract the `message_code` from `RawJSON` when handling specific errors, such as checking if an invoice is already paid.

### Detailed Description

When executing `POST /v1/invoice/:id/pay` on an already paid invoice, the following error is returned:

```json
{
  "error": {
    "message": "foo",
    "message_code": "invoice_already_paid",
    "request_log_url": "https://dashboard.stripe.com/test/logs/foo",
    "type": "invalid_request_error"
  }
}
```

Currently, since the `Error` struct does not have a field that includes `message_code`, it is necessary to write code to extract `message_code` from `RawJSON`.

The example of Before and After of this modification is as follows:

```go
// Before
var stripeErr *stripe.Error
if errors.As(err, &stripeErr) {
    switch stripeErr.Type {
    case stripe.ErrorTypeInvalidRequest:
        if stripeErr.LastResponse != nil {
            var responseBody map[string]interface{}
            jsonErr := json.Unmarshal(stripeErr.LastResponse.RawJSON, &responseBody)
            if jsonErr != nil {
                return jsonErr
            }
            if errorObj, ok := responseBody["error"].(map[string]interface{}); ok {
                messageCode, _ := errorObj["message_code"].(string)
                if messageCode == "invoice_already_paid" {
                    // Process for already paid invoice
                }
            }
        }
    }
}

// After
var stripeErr *stripe.Error
if errors.As(err, &stripeErr) {
    switch stripeErr.Type {
    case stripe.ErrorTypeInvalidRequest:
        if stripeErr.MsgCode == "invoice_already_paid" {
            // Process for already paid invoice
        }
    }
}
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -268,6 +268,7 @@ type Error struct {
 
 	HTTPStatusCode    int               `json:"status,omitempty"`
 	Msg               string            `json:"message"`
+	MsgCode           string            `json:"message_code,omitempty"`
 	Param             string            `json:"param,omitempty"`
 	PaymentIntent     *PaymentIntent    `json:"payment_intent,omitempty"`
 	PaymentMethod     *PaymentMethod    `json:"payment_method,omitempty"`
--- a/error_test.go
+++ b/error_test.go
@@ -19,7 +19,7 @@ func TestErrorResponse(t *testing.T) {
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Add("Request-Id", "req_123")
 		w.WriteHeader(http.StatusUnauthorized)
-		fmt.Fprintln(w, `{"error":{"message":"bar","type":"`+ErrorTypeInvalidRequest+`"}}`)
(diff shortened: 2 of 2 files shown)


remi-stripe:
@wozaki The `message_code` property on the `Error` API resource is not a public feature. It does exist but it's purely internal to Stripe and shouldn't be exposed anywhere in the public API today (though you can sometimes see it in error messages on the Dashboard or via Workbench. This means we wouldn't add this feature to our public server-side SDKs.

Can you confirm that this maps to what you were seeing and that you do not see this property in our public API today?

wozaki:
@remi-stripe 
Thank you for your review. 
Upon further review, I confirmed that the `message_code` property only appears in error responses via the Dashboard.
I also understand that it should not be used in the public API at this time. 
I will go ahead and close this Pull Request.

PR #1924, opened 2024-09-27
add support for RawRequest

xavdid-stripe:
### Why?

As the v2 API is releasing soon, we want Go to be able to make requests against the new endpoints despite not having generated types.

### What

This is a copy/paste of https://github.com/stripe/stripe-go/pull/1648 with a few key differences:

- updated README with a more modern example
- changed the `APIMode` constants from `"preview"` and `"standard"` to `"v1"` and `"v2"`
- ❗inferred API mode from path prefix instead of letting user specify it
- removed some assertions in tests that cared about the API version

Because we no longer needed to supply default values for API mode, I could remove the function that handled that.  I also adjusted tests that assumed you could set `content-type` and `v1` / `v2` independently (given that all v1 requests are form encoded and all v2 request are json).

### Manual Testing

1. Create a new folder, henceforth `myfolder`
2. In `myfolder/go.mod`, paste:

```go
module xavd.id/test/v2

go 1.22.4

require github.com/stripe/stripe-go/v79 v79.0.0
// may need to adjust this relative path for your filesystem:
replace github.com/stripe/stripe-go/v79 v79.0.0 => ../../../stripe-go
```

3. In `myfolder/main.go` paste:

```go
package main

import (
	"encoding/json"
	"fmt"

	"github.com/stripe/stripe-go/v79"
	"github.com/stripe/stripe-go/v79/form"
	"github.com/stripe/stripe-go/v79/rawrequest"
)

func do_thing() error {
	stripe.Key = "sk_test_..."

	payload := map[string]interface{}{
		"event_name": "hotdogs_eaten", // FIXME
		"payload": map[string]string{
			"value":              "123", 
			"stripe_customer_id": "cus_Quq8itmW58RMet", // FIXME
		},
	}

	// for a v2 request, json encode the payload
	body, err := json.Marshal(payload)
	if err != nil {
		return err
	}

	v2_resp, err := rawrequest.Post("/v2/billing/meter_events", string(body), nil)
	if err != nil {
		return err
	}

	var v2_response map[string]interface{}
	err = json.Unmarshal(v2_resp.RawJSON, &v2_response)
	if err != nil {
		return err
	}
	fmt.Printf("%#v\n", v2_response)

	// for a v1 request, form encode the payload
	formValues := &form.Values{}
	form.AppendTo(formValues, payload)
	content := formValues.Encode()

	v1_resp, err := rawrequest.Post("/v1/billing/meter_events", content, nil)
	if err != nil {
		return err
	}

	var v1_response map[string]interface{}
	err = json.Unmarshal(v1_resp.RawJSON, &v1_response)
	if err != nil {
		return err
	}
	fmt.Printf("%#v\n", v1_response)

	return nil
}

func main() {
	err := do_thing()
	if err != nil {
		fmt.Printf("Error: %s\n", err)
	}
}
```

Running that (with valid customer id and event name) should print 2 response bodies

### See Also

- [DEVSDK-2187](https://jira.corp.stripe.com/browse/DEVSDK-2187)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Go Stripe
+
 [![Go Reference](https://pkg.go.dev/badge/github.com/stripe/stripe-go)](https://pkg.go.dev/github.com/stripe/stripe-go/v79)
 [![Build Status](https://github.com/stripe/stripe-go/actions/workflows/ci.yml/badge.svg?branch=master)](https://github.com/stripe/stripe-go/actions/workflows/ci.yml?query=branch%3Amaster)
 [![Coverage Status](https://coveralls.io/repos/github/stripe/stripe-go/badge.svg?branch=master)](https://coveralls.io/github/stripe/stripe-go?branch=master)
@@ -14,13 +15,13 @@ The official [Stripe][stripe] Go client library.
 Make sure your project is using Go Modules (it will have a `go.mod` file in its
 root if it already is):
 
-``` sh
+```sh
 go mod init
 ```
 
 Then, reference stripe-go in a Go program with `import`:
 
-``` go
+```go
 import (
 	"github.com/stripe/stripe-go/v79"
 	"github.com/stripe/stripe-go/v79/customer"
@@ -256,7 +257,7 @@ if err := i.Err(); err != nil {
 Use `LastResponse` on any `APIResource` to look at the API response that
 generated the current object:
 
-``` go
+```go
 c, err := coupon.New(...)
 requestID := coupon.LastResponse.RequestID
 ```
@@ -264,7 +265,7 @@ requestID := coupon.LastResponse.RequestID
 Similarly, for `List` operations, the last response is available on the list
 object attached to the iterator:
 
-``` go
+```go
 it := coupon.List(...)
 for it.Next() {
     // Last response *NOT* on the individual iterator object
@@ -354,7 +355,7 @@ full resource struct, but unless expansion is requested, only the `ID` field of
 that struct is populated. Expansion is requested by calling `AddExpand` on
 parameter structs. For example:
 
-``` go
+```go
 //
 // *Without* expansion
 //
@@ -373,11 +374,12 @@ c, _ = charge.Get("ch_123", p)
 c.Customer.ID    // ID is still available
 c.Customer.Name  // Name is now also available (if it had a value)
 ```
+
 ### How to use undocumented parameters and properties
 
 stripe-go is a typed library and it supports all public properties or parameters.
 
-Stripe sometimes launches private beta features which introduce new properties or parameters that are not immediately public. These will not have typed accessors in the stripe-go library but can still be used. 
+Stripe sometimes launches private beta features which introduce new properties or parameters that are not immediately public. These will not have typed accessors in the stripe-go library but can still be used.
 
 #### Parameters
 
@@ -412,13 +414,15 @@ secret_parameter, ok := rawData["secret_parameter"].(map[string]interface{})
 if ok {
 	primary := secret_parameter["primary"].(string)
 	secondary := secret_parameter["secondary"].(string)
-} 
+}
 ```
 
 ### Webhook signing
+
 Stripe can optionally sign the webhook events it sends to your endpoint, allowing you to validate that they were not sent by a third-party. You can read more about it [here](https://stripe.com/docs/webhooks/signatures).
 
 #### Testing Webhook signing
+
 You can use `stripe.webhook.GenerateTestSignedPayload` to mock webhook events that come from Stripe:
 
 ```go
@@ -477,10 +481,13 @@ config := &stripe.BackendConfig{
 To mock a Stripe client for a unit tests using [GoMock](https://github.com/golang/mock):
 
 1. Generate a `Backend` type mock.
+
 ```
 mockgen -destination=mocks/backend.go -package=mocks github.com/stripe/stripe-go/v79 Backend
 ```
+
 2. Use the `Backend` mock to initialize and call methods on the client.
+
 ```go
 
 import (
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -1,4 +1,5 @@
 # Go Stripe
+
```
xavdid-stripe, line 2: this got prettier-ed, but the only actual changes are the section I added

README.md
```diff
@@ -531,19 +538,66 @@ go get -u github.com/stripe/stripe-go/v79@v77.1.0-beta.1
...
+If you would like to send a request to an API that is:
+
+- not yet supported in stripe-go (like any `/v2/...` endpoints), or
+- undocumented (like a private beta), or
+- public, but you prefer to bypass the method definitions in the library and specify your request details directly
+
+You can use the `RawRequest` method on the `stripe` backend.
+
+```go
+import "github.com/stripe/stripe-go/v79"
```
xavdid-stripe, line 565: I may also add a simplified example here tomorrow before merge, but this example also works

params.go
```diff
@@ -172,6 +172,11 @@ type ListParamsContainer interface {
 	GetListParams() *ListParams
 }
 
+type APIMode string
+
+var V1APIMode APIMode = "v1"
```
xavdid-stripe, line 177: changed from

```go
var PreviewAPIMode APIMode = "preview"
var StandardAPIMode APIMode = "standard"
```

stripe_test.go
```diff
@@ -1280,6 +1280,213 @@ func TestBoolSlice(t *testing.T) {
...
+	params := &RawParams{Params: Params{}, APIMode: V2APIMode}
+	response, err := backend.RawRequest(http.MethodPost, "/v1/abcs", "sk_test_xyz", `{"foo":"myFoo","bar":{"baz":false}}`, params)
+	assert.NoError(t, err)
+	myABC := &MyABC{}
+	assert.Nil(t, params.Headers)
+	assert.Equal(t, `{"foo":"myFoo","bar":{"baz":false}}`, body)
+	assert.Equal(t, `/v1/abcs`, path)
+	assert.Equal(t, `POST`, method)
+	assert.Equal(t, `application/json`, contentType)
+	// assert.Equal(t, previewVersion, stripeVersion)
```
xavdid-stripe, line 1324: removed for now, but maybe useful later

api_version.go
```diff
@@ -7,5 +7,5 @@
 package stripe
 
 const (
-	apiVersion string = "2024-06-20"
+	apiVersion string = "2024-09-30.acacia"
```
xavdid-stripe, line 10: I changed this to get my local example working, but am not sure it should be included in this PR

ramya-stripe: change the base of the PR from master to sdk-release/next-major - that branch has this change

params.go
ramya-stripe: The readme is still referring to` stripe.PreviewAPIMode` though

api_version.go
xavdid-stripe: good call! Changed the base branch, so this will be a no-op. I'm not sure GH will reflect that though

README.md
```diff
@@ -531,19 +538,89 @@ go get -u github.com/stripe/stripe-go/v79@v77.1.0-beta.1
...
+		},
+	}
+
+	// for a v2 request, json encode the payload
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return err
+	}
+
+	v2_resp, err := rawrequest.Post("/v2/billing/meter_events", string(body), nil)
```
jar-stripe, line 591: do users need to specify Content-Type for this to work w/ v2 and JSON?

rawrequest/client_test.go
```diff
@@ -0,0 +1,130 @@
...
+	var path string
+	var method string
+	var contentType string
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		req, _ := ioutil.ReadAll(r.Body)
+		r.Body.Close()
+		body = string(req)
+		path = r.URL.RequestURI()
+		method = r.Method
+		contentType = r.Header.Get("Content-Type")
```
jar-stripe, line 40: may want to assert on contentType == application/json, or return a 400 if its not, or something

rawrequest/client_test.go
```diff
@@ -0,0 +1,130 @@
...
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte(`{"object": "abc", "xyz": {"def": "jih"}}`))
+	}))
+
+	stubAPIBackend(testServer)
+
+	var params *stripe.RawParams
+
+	_, err := Post("/v1/abc", `{"xyz": {"def": "jih"}}`, params)
```
jar-stripe, line 81: this isnt correct for a v1 api, right?  i.e. should be form encoded?

rawrequest/client_test.go
```diff
@@ -0,0 +1,130 @@
...
+	var stripeContext string
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		req, _ := ioutil.ReadAll(r.Body)
+		r.Body.Close()
+		body = string(req)
+		path = r.URL.RequestURI()
+		method = r.Method
+		contentType = r.Header.Get("Content-Type")
+		fooHeader = r.Header.Get("foo")
+		stripeContext = r.Header.Get("Stripe-Context")
```
jar-stripe, line 108: this just returns empty string if it is not found; we should assert against it and fail if its not right.

stripe.go
```diff
@@ -417,14 +421,78 @@ func (s *BackendImplementation) CallMultipart(method, path, key, boundary string
...
+
 // CallRaw is the implementation for invoking Stripe APIs internally without a backend.
 func (s *BackendImplementation) CallRaw(method, path, key string, form *form.Values, params *Params, v LastResponseSetter) error {
 	var body string
 	if form != nil && !form.Empty() {
 		body = form.Encode()
 
 		// On `GET`, move the payload into the URL
-		if method == http.MethodGet {
+		if method != http.MethodPost {
```
jar-stripe, line 495: do we ever issue put or patch calls? if so, we want to keep the body in the body for those

stripe_test.go
```diff
@@ -1280,6 +1280,213 @@ func TestBoolSlice(t *testing.T) {
...
+	var method string
+	var contentType string
+	var stripeVersion string
+	testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		req, _ := ioutil.ReadAll(r.Body)
+		r.Body.Close()
+		body = string(req)
+		path = r.URL.RequestURI()
+		method = r.Method
+		contentType = r.Header.Get("Content-Type")
```
jar-stripe, line 1382: adding this comment again: r.Header.Get (I _think_) will always "succeed" so we should have an assert or some kind of evaluation where we want to check that the header is of a specific value

README.md
jar-stripe: nm, answered my question as I went along; we use the uri prefix (v1 or v2) to determine content type, right?

stripe.go
jar-stripe: I think I answered my own question here; it looks like we restrict that above.  Given that, I am curious why this change was needed.  GETs are the only methods that cannot have a body.

jar-stripe (changes requested):
Left a few non-blocking comments, but it looks like this repo is not set up to prevent merges on open conversations, so I wanted to not approve it just yet.

README.md
xavdid-stripe: correct, they can't specify API mode or content type manually. there didn't seem to be a point if we could just do it for them

rawrequest/client_test.go
xavdid-stripe: we do farther down - this line stores the incoming value on the "server" and then we assert its value after the response finishes.

rawrequest/client_test.go
xavdid-stripe: the test is technically wrong, yeah. this would be a 400 on the real API, but the tests treat `content` as opaque strings. The important part is that content-type is set correctly and the response is decoded. 

That said, I can update if you'd like!

stripe.go
xavdid-stripe: That was copied from the original PR: https://github.com/stripe/stripe-go/pull/1648/files#diff-f34528b48beca228ad5851464b3e20496f8c34c1f7b663e7f109fa9f5ed02fc4R426

I guess richard wanted the behavior to apply to `DELETE` in addition to `GET`? I don't have strong feelings, but I mostly erred towards not changing what's been in the beta branch 🤷

rawrequest/client_test.go
xavdid-stripe: https://github.com/stripe/stripe-go/pull/1924/files#diff-d25e28f49be803aa8fb79967e7cacada37a257743d0d82e8c4585b5a26ab9e1fR56

jar-stripe: ah i see that now.  thanks!

rawrequest/client_test.go
jar-stripe: i think it makes sense to update just from a readability standpoint; this might confuse me if I were new or unfamiliar.

stripe.go
jar-stripe: thats fair. can you update the comment?

xavdid-stripe:
@jar-stripe updated the comment, test string body, and extra HTTP method validation in https://github.com/stripe/stripe-go/pull/1924/commits/a05ed2d66ee340ab603a045e3de9e98f2a522154! I noticed that `CallRaw` was actually a different call path and could use validation as well.

jar-stripe (approved):
looks good!

PR #1929, opened 2024-10-02
raw_request.Client + V2 UBB Examples

jar-stripe:
### Why?
Stripe SDK for go does not have directly support V2 APIs but it is still possible to access them for e.g. Usage Based Billing.  This PR contains examples that show how to use the rawrequest functionality to handle the new events and endpoints.  Because the high-throughput meter event stream requires changing both the backend and the API key, this PR also introduces a raw_request.Client struct that behaves like other services.

### What
- added raw_request.Client which accepts a RawRequestBackend and API key; this can be different than stripe.Key or stripe.APIBackend
- added support for a MeterEvents backend to `stripe`, to enable posting meter events
- added thinevent_webhook_handler.go which receives a thin event, gets the v2 Event and then the related object for a v1.billing.meter.error_report_triggered event
- added meter_event_stream.go which demonstrates how to utilize the meter event stream for high throughput meter events


ramya-stripe:
@jar-stripe Can you look at the test failures please?

rawrequest/client.go
```diff
@@ -7,12 +7,24 @@ import (
 	stripe "github.com/stripe/stripe-go/v80"
 )
 
+// Client is used to invoke /quotes APIs.
```
ramya-stripe, line 10: quotes API?

example/v2/thinevent_webhook_handler/thinevent_webhook_handler.go
```diff
@@ -0,0 +1,95 @@
...
+var webhookSecret = "{{WEBHOOK_SECRET}}"
+
+func main() {
+	b, err := stripe.GetRawRequestBackend(stripe.APIBackend)
+	if err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+
+	client := rawrequest.Client{B: b, Key: apiKey}
```
ramya-stripe, line 26: From what I understood, we introduce the client in the rawrequest package so that we can use a different backend than the default. So I get why we want to use it in the `meter_event_stream.go` example, but not why we would want to use it here

rawrequest/client.go
jar-stripe: thanks for catching this; it was a typo.  fixing!

example/v2/thinevent_webhook_handler/thinevent_webhook_handler.go
jar-stripe: I think there are a couple reasons to want to do this in the example here:

1. using rawrequest.Client here keeps the interface consistent across all raw request examples, and I think is what we want users to do longer term, because
2. this feels more akin to the StripeClient pattern but is also structured the same as other golang services.  so I think migration in the future will be simpler if we introduce StripeClient but it's not going to feel out of place in the mean timer

This change is also coupled with a change in rawrequest/client.go which removes Get, Post, and Delete.  My thought here was just retired the preview API (e.g. `Get`, `Post`, `Delete`) in other SDKs in lieu of a single `RawRequest`, so it additionally keeps things consistent with our other SDKs.

jar-stripe:
> @jar-stripe Can you look at the test failures please?

Tests pass, and I updated the title and PR description to better represent the scope of the PR.  If you agree with my reasoning above re: using rawrequest.Client in both examples, this should be good to go!

ramya-stripe (approved):

ramya-stripe:
Note: We will need to do a release of the Go SDK for the example to be usable

PR #1936, opened 2024-10-13
make optional event attributes optional

m-mattia-m:
When my API receives a webhook request from Stripe via the CLI, I get the error `expected required property account to be present` and the location attribute is: `body`. After creating a manual request with the same body and adding the 'account' attribute (which is an empty string), I got the same error for the 'previous_attributes' attribute in the 'body.data' location. After adding that element as well, it worked.

I checked the stripe docs, I saw that the `account' attribute is a `nullable string'. The `previous_attributes` is a `nullable map`. However, the stripe-go-sdk considers these attributes to be required attributes.

[Stripe Docs - The Event object](https://docs.stripe.com/api/events/object)


jar-stripe:
Hi @m-mattia-m!  Thanks for opening this PR!  The events.go file is actually generated from our API specification, so we'll need to look at what may be causing these fields to not be marked correctly as optional.  Would you mind also opening a GH issue that describes the issue?  If you can include an example snippet of code and commands to run to show the error, that would be awesome.  We can do some investigation internally, and then collaborate on the fix on this PR.

mbroshi-stripe:
This is being tracked in https://github.com/stripe/stripe-go/issues/1937, so closing out this PR.

PR #1941, opened 2024-10-25
Do not allow setting API Version directly 

ramya-stripe:
## Why?

When we introduced beta SDKs, we allowed users to directly update the global configuration for API Version since they needed to pass beta headers in #1529. We soon realized that was not safe and was error prone if users didnt pass in the right format and so introduced a helper method in #1819 and advertised that as the right way of doing things in the README

Proper deserialization of classes from Events can be guaranteed only when the Webhook API version matches the API version used to generate the SDKs. Therefore, in this PR we are dropping the ability to directly update the API version.

## What?

- Resolving merge conflicts coming from running the auto merge tool after #1940
- Rename the public variable `APIVersion` to a private variable `apiVersionWithBetaHeaders` and add a new constant for `APIVersion`
- Updated tests

## Changelog
* `stripe.APIVersion` is no longer settable. If you were using this to set the beta headers, use the helper method `stripe.AddBetaVersion()` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,7 @@ We would love for you to try these and share feedback with us before these featu
 To install a beta version of stripe-go use the commit notation of the `go get` command to point to a beta tag:
 
 ```
-go get -u github.com/stripe/stripe-go/v80@v77.1.0-beta.1
+go get -u github.com/stripe/stripe-go/v80@beta
 ```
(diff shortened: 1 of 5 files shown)


jar-stripe (approved):
LGTM

PR #1964, opened 2025-01-13
add justfile, update readme, remove coveralls

xavdid-stripe:
### Why?

In an effort to modernize and simplify our local tooling, we're moving our dev commands from makefiles to [justfiles](https://github.com/casey/just). This is intended to be mostly a drop-in replacement, but some command names may change per standardization efforts.

### What?

- add justfile
- tweak CI to use justfile
- remove coveralls from CI

### See Also

[DEVSDK-2325](https://go/j/DEVSDK-2325)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+      - uses: extractions/setup-just@v2
       - uses: actions/checkout@master
 
       - name: Setup go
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -34,42 +35,33 @@ jobs:
...
-       - uses: stripe/openapi/actions/stripe-mock@master
-       - name: Test
-         run: make ci-test
-       - name: Coveralls
-         run: make coverage && make coveralls
-         if: matrix.go == '1.16'
-         env:
-           COVERALLS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-           COVERALLS_FLAG_NAME: Go-${{ matrix.go }}
+    runs-on: ubuntu-latest
```
xavdid-stripe, line 41: this block is mostly the same except coveralls is gone and `just` is used in favor of `make`. Also, prettier shifted it over (it was intended by 5 spaces?)

justfile
```diff
@@ -0,0 +1,62 @@
...
+bench:
+    go test -race -bench . -run "Benchmark" ./form
+
+# called by tooling. It updates the package version in the `VERSION` file and `stripe.go`
+[private]
+update-version version: && _normalize-imports
+    echo "{{ version }}" > VERSION
+    perl -pi -e 's|const clientversion = "[.\d\-\w]+"|const clientversion = "{{ version }}"|' stripe.go
+
+# go imports use the package's major version in the path, so we need to update them
```
xavdid-stripe, line 52: This was the only especially tricky thing- we want `major_version` to be evaluated lazily, but also shared throughout the recipe. 

This comment is probably too detailed, but I wanted to remind myself why I did it exactly this way.

README.md
```diff
@@ -641,19 +641,13 @@ the following guidelines in mind:
...
 2. All types, structs and funcs should be documented.
-3. Ensure that `make test` succeeds.
+3. Ensure that `just test` succeeds.
 
 ## Test
 
-The test suite needs testify's `require` package to run:
-
-    github.com/stretchr/testify/require
-
```
jar-stripe, line 651: do we not need require anymore?

README.md
```diff
@@ -641,19 +641,13 @@ the following guidelines in mind:
...
-Before running the tests, make sure to grab all of the package's dependencies:
-
-    go get -t -v
```
jar-stripe, line 654: should we include a way to do this thru just?  or is that elsewhere in the doc?

jar-stripe (commented):
looks good overall; a couple of questions about things you removed from the README.md, but if those are covered elsewhere this is an easy 👍

README.md
xavdid-stripe: It's included in the `go.mod`, so it's automatically pulled by `go get` (or `go build`), so it doesn't need it's own mention in the readme, IMO

README.md
xavdid-stripe: Go calls this whenever you do a different action (`build`, `test`, etc) so this felt like extra in the README.

I verified this with:

```
git clone https://github.com/stripe/stripe-go build-test
cd build-test
go test ./...
```

and (assuming `stripe-mock` is running) everything works out of the box.

xavdid-stripe: but otherwise yeah, `just` bundles all this if you're doing it natively (just in case, I guess).

README.md
jar-stripe: ah ok, good enough for me!

README.md
jar-stripe: cool, thanks!

Issue #1205, opened 2020-10-11
Parse unix timestamps into times (time.Time)


msingleton:
It'd be nice to parse timestamps into their time.Time during unmarshaling. I'd expect most people will be operating with the values as times (could be wrong) so most people are having to do `time.Unix(invoice.PeriodEnd, 0)`

Obviously not a big deal, but nice to have!

richardm-stripe:
Hi @msingleton. I think this makes sense, and we have analogs in other libraries. We have plans to codegen this library in the not-too-distant future which would make a change like this a lot easier to do, so we will punt on this for now but possibly revisit after that lands.

msingleton:
👍 sounds great

mbroshi-stripe:
Hi @msingleton! This is an old issue, but I wanted to let you know we have implemented this feature in https://github.com/stripe/stripe-go/pull/1972. It's currently available only on the [beta branch](https://github.com/stripe/stripe-go/tree/beta) (since it's a breaking change), but we would love if you had any feedback regarding the change before it goes live.

mbroshi-stripe:
Belated update here:
* We reverted https://github.com/stripe/stripe-go/pull/1972 in https://github.com/stripe/stripe-go/pull/1981 since we decided this change would have a unnecessarily large backwards incompatible blast radius.
* We are using `time.Time` in all of our V2 endpoints.
* I will leave this issue open for now, as we may consider incorporating it into the new `stripe.Client` pattern that we introduced in https://github.com/stripe/stripe-go/releases/tag/v82.1.0

mbroshi-stripe:
Although I totally agree that using `time.Time` would be more ergonomic, we have decided to close this issue. The reason is that it would be a foundational breaking change that would break every single integration. The payoff from that breaking change does not seem to counterbalance the disruption it would cause.

msingleton:
@mbroshi-stripe Ah yea, that makes sense -- appreciate you all looking into it

PR #1972, opened 2025-01-22
Support time.Time instead of int64 for date fields

mbroshi-stripe:
### Why?

This change supports codegen changes that converts datetime fields from `int64` to `time.Time` as brought up in this issue:  https://github.com/stripe/stripe-go/issues/1205.

### What?

* The main change is in the form serialization in `form.go` for `time.Time` fields. Serialization is performed by reflection, and we do not want the default behavior of inspecting into a `time.Time` type, rather just convert it to an `int64`
* While I was in there, I also created convenience method `Time`, `TimeValue`, `UnixTime` and `UnixValue` to help in creating `time.Time` and `*time.Time` parameter fields
* I also updated unit tests expecting an `int64` type

### See Also

*  https://github.com/stripe/stripe-go/issues/1205

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -469,8 +469,8 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 }
 
 func makeTypeEncoder(t reflect.Type) encoderFunc {
-	// For time.Time, we want to encode it as a Unix timestamp,
-	// and don't want to recurse into it.
+	// For time.Time, we want to encode imediately it as a Unix timestamp,
+	// and don't want to inspect into it and encode it as a struct.
 	if t == reflect.TypeOf(time.Time{}) {
 		return timeEncoder
 	}
--- a/form/form.go
+++ b/form/form.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 )
 
 const tagName = "form"
@@ -325,6 +326,14 @@ func intEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero b
 	values.Add(FormatKey(keyParts), strconv.FormatInt(val, 10))
 }
 
+func timeEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero bool, _ *formOptions) {
+	val := v.Interface().(time.Time)
+	if val.IsZero() && !encodeZero {
+		return
+	}
+	values.Add(FormatKey(keyParts), strconv.FormatInt(val.Unix(), 10))
+}
+
 func interfaceEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero bool, _ *formOptions) {
 	// interfaceEncoder never encodes a `nil`, but it will pass through an
 	// `encodeZero` value into its chained encoder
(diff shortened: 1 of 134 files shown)


form/form.go
```diff
@@ -325,6 +326,14 @@ func intEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero b
 	values.Add(FormatKey(keyParts), strconv.FormatInt(val, 10))
 }
 
+func timeEncoder(values *Values, v reflect.Value, keyParts []string, encodeZero bool, _ *formOptions) {
+	val := v.Interface().(time.Time)
```
jar-stripe, line 330: what happens if this is called with a v other than a time.Time; does this return an error we need to look for, or will val just be Zero (and is that swallowing an error?)

form/form.go
```diff
@@ -461,6 +469,12 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 }
 
 func makeTypeEncoder(t reflect.Type) encoderFunc {
+	// For time.Time, we want to encode it as a Unix timestamp,
+	// and don't want to recurse into it.
```
jar-stripe, line 473: the first part of this comment makes sense, but I'm not seeing the recursion.  can you clarify?

issuing/dispute/client_test.go
```diff
@@ -35,15 +35,15 @@ func TestIssuingDisputeNew(t *testing.T) {
 		Evidence: &stripe.IssuingDisputeEvidenceParams{
 			Canceled: &stripe.IssuingDisputeEvidenceCanceledParams{
 				AdditionalDocumentation:    stripe.String("file_123"),
-				CanceledAt:                 stripe.Int64(1577836800),
+				CanceledAt:                 stripe.UnixTime(1528573382),
```
jar-stripe, line 38: lets walk thru these linting errors to see where they come from; I would expect the types to be right if e.g. CanceledAt is a time.Time

stripe.go
```diff
@@ -1381,6 +1381,32 @@ func StringSlice(v []string) []*string {
 	return out
 }
 
+// Time returns a pointer to the time.Time value passed in.
```
jar-stripe, line 1384: I'm not too familiar with the stripe-go layout; is this where these helper methods go?  or can we break out a time.go file?

helenye-stripe:
:lurk: I think the merge-base for this PR should also be the `beta` branch!

form/form.go
mbroshi-stripe: It panics. I think that's OK, but open for other thoughts. It's only called in one location surrounded by
```
if t == reflect.TypeOf(time.Time{}) {
	return timeEncoder
}
```
so logically, the type assertion `v.Interface().(time.Time)` must succeed. I would definitely rather fail fast here rather than silently. The alternative would be to bubble up an `error`, but it feels a little weird to return an `error` due to a bug.

mbroshi-stripe: In fact, I'm realizing we have the same `panic` behavior in the rest of type assertions in `form.go`, so this is really no different than current behavior (where we do a type assertion to an integer: https://github.com/stripe/stripe-go/blob/b384f33fa0506bc052f50bf8c8be01b109278d05/form/form.go#L321

jar-stripe: Cool, if its consistent with what we have I think this is good.  Maybe this is something we should look at separately, i.e. if a panic is the right response here (vs something more graceful) but that would be separate from this PR.

jar-stripe (approved):
LGTM!

joeriddles:
When does the beta branch get merged into master? Looks like this change is not in master yet. Curious what the rollout of beta->master looks like. Thanks!

mbroshi-stripe:
@joeriddles I just left a [new comment](https://github.com/stripe/stripe-go/issues/1205#issuecomment-2985516368) on https://github.com/stripe/stripe-go/issues/1205. In short, we decided to revert this change since it had a large, backwards incompatible user impact. Feel free to comment further in that issue, though, if you have more to share!

PR #1977, opened 2025-01-27
Upgrade to go 1.18

mbroshi-stripe:
### Why?

We are currently pinned to a very old version of `golang.org/x/net` which contains a number of vulnerabilities, of which `govulncheck` reports two. The minimal version that address the vulnerabilities is `golang.org/x/net@v0.23.0`. In order to upgrade to `golang.org/x/net@v0.23.0`, we must upgrade the Go version in `go.mod` to 1.18 (found by binary search).

### What?

- Upgrades Go runtime to `1.18`
- Upgrades `golang.org/x/net` to `v0.23.0`
- Only runs CI checks back to `go 1.18`.
- Removes special case code around running `http.Client` on older versions of Go
- Adds `govulncheck` CI test to `sdk-release/next-major` branch

### See Also

Output from `govulncheck` before change:
```
$ govulncheck ./...
=== Symbol Results ===

Vulnerability #1: GO-2024-2687
    HTTP/2 CONTINUATION flood in net/http
  More info: https://pkg.go.dev/vuln/GO-2024-2687
  Module: golang.org/x/net
    Found in: golang.org/x/net@v0.0.0-20210520170846-37e1c6afe023
    Fixed in: golang.org/x/net@v0.23.0
    Example traces found:
    ...

Vulnerability #2: GO-2023-1571
    Denial of service via crafted HTTP/2 stream in net/http and golang.org/x/net
  More info: https://pkg.go.dev/vuln/GO-2023-1571
  Module: golang.org/x/net
    Found in: golang.org/x/net@v0.0.0-20210520170846-37e1c6afe023
    Fixed in: golang.org/x/net@v0.7.0
    ....
```
after change:
```
$ govulncheck ./...
=== Symbol Results ===

No vulnerabilities found.
```

## Changelog
* Go version 1.18 or later is now required to address security vulnerabilities in Go <= 1.17. In particular, HTTP/2 is enabled now by default for all users (it was disabled for Go <= 1.14).


mbroshi-stripe:
@prathmesh-stripe Instead of rebasing, it was easier to create a new PR branched off of [sdk-release/next-major](https://github.com/stripe/stripe-go/tree/sdk-release/next-major)

.github/workflows/ci.yml
```diff
@@ -37,6 +37,21 @@ jobs:
         run: |
           just lint format-check
 
+  govulncheck:
+    name: govulncheck
+    runs-on: "ubuntu-24.04"
+    if: github.base_ref == 'beta'
```
ramya-stripe, line 43: we don't need this condition anymore right?

stripe.go
```diff
@@ -1456,44 +1455,8 @@ var encodedUserAgent string
...
-	//
-	// This is particularly problematic for this library because the empty body
-	// results in no parameters being sent, which usually results in a 400,
-	// which is a status code expressly not covered by retry logic.
-	//
-	// The bug seems to be somewhat tricky to fix and hasn't seen any traction
-	// lately, so for now we're mitigating by disabling HTTP/2 in stripe-go by
-	// default. Users who like to live dangerously can still re-enable it by
-	// specifying a custom HTTP client. When the bug above is fixed, we can
-	// turn it back on.
```
ramya-stripe, line 1478: So looks like the default behavior for users will change moving forward. We should add a changelog entry

.github/workflows/ci.yml
mbroshi-stripe: Hmm, good point. I'll need to put some condition since this check will fail on `master`. I guess I can put `sdk-release/next-major` there for now, and then change it to `master` after we merge/port this change to `master` in March.

stripe.go
mbroshi-stripe: Yes and no. The current behavior is that if you are on Go <= 1.14, HTTP/2 is disabled, but Go >= 1.15 it is enabled. This change will force folks to upgrade to Go >= 1.18, and in particular HTTP/2 will be enabled for them. I can emphasize that in the changelog, but technically it's not really change in behavior.

prathmesh-stripe (approved):

stripe.go
ramya-stripe: 🤦‍♀️ Should have read the comments completely

.github/workflows/ci.yml
ramya-stripe: But... won't this action be run only in the `sdk-release/next-major` branch regardless of the if check because the action does not exist in the ci.yml files of other branches?

mbroshi-stripe: Hah! My turn for a 🤦 ! I'll remove the branch restriction in a follow up PR

PR #1982, opened 2025-02-05
Add usage data for stripe client usage

mbroshi-stripe:
### Why?

We would like to get a better sense of how many users are using Stripe Clients (aka. `client.API`) vs the global client. This will help us better inform how to update documentation and consider possible changes to the Stripe Client going forward.

### What?

This sends the string `stripe_client` to telemetry statistics when they are enabled when a Stripe Client was used for that request. It does this by wrapping the existing `Backend`s with a `usageBackend` that sets the string `stripe_client` on requests. This string is not sent on the global client, so we can distinguish the two.

I also updated a few random tests to use the Stripe Client to add some additional test coverage for this change.

## Changelog
* Add telemetry for usage of the Stripe Client

### DIFF
--- a/account/client_test.go
+++ b/account/client_test.go
@@ -10,28 +10,28 @@ import (
 )
 
 func TestAccountDel(t *testing.T) {
-	sc := client.New(TestApiKey, nil)
+	sc := client.New(TestAPIKey, nil)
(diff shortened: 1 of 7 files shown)


helenye-stripe (commented):
just one small nit

helenye-stripe (approved):

PR #1985, opened 2025-02-07
Add Stripe Client to example tests

mbroshi-stripe:
### Why?

In order to have a stronger set of tests on the Stripe Client before any upcoming changes to support V2.

### What?

For each resource-based example test of the form `TestFoo`, this adds a corresponding service-based test of the form `TestFooService`.


mbroshi-stripe:
FYI, test failures  in `go-next` and `go-beta` are because https://github.com/stripe/stripe-go/pull/1982 has not yet merged into those branches.

xavdid-stripe (approved):

PR #1996, opened 2025-03-14
Replace "$resource$" with "Customer" in README code snippets

mbroshi-stripe:
### Why?

The README was confusing with `$resource$`.

### What?

- Replace `$resource$` with `Customer` in the code snippets in the README
- Also replace the explicit `client.API{}` and `sc.Init` call with `client.New()` in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,9 @@ import (
 )
 
 // Setup
-sc := client.New("sk_key", nil) // second argument is for passing in backends
+sc := client.New("sk_key", nil)
+// To set backends, e.g. for testing, use this instead:
+// sc := client.New("sk_key", backends)
 
 // Create
 c, err := sc.Customers.New(&stripe.CustomerParams{})
--- a/README.md
+++ b/README.md
@@ -226,11 +226,7 @@ import (
 )
 
 // Setup
-sc := client.New("sk_key")
-
-// To set backends, e.g. for testing:
-// sc := &client.API{}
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -226,11 +226,7 @@ import (
 )
 
 // Setup
-sc := client.New("sk_key")
-
-// To set backends, e.g. for testing:
-// sc := &client.API{}
-// sc.Init("sk_key", backends)
+sc := client.New("sk_key", nil) // second argument is for passing in backends
```
jar-stripe, line 229: I think the old comment was clearer here: what if we reverted line 229 and then added something like this immediately after:
```
// To set backends, e.g. for testing, use this instead:
// sc := client.New("sk_key", backends)
```

README.md
```diff
@@ -190,21 +190,22 @@ stripe.Key = "sk_key"
 // stripe.SetBackend("api", backend)
 
 // Create
-resource, err := $resource$.New(&stripe.$Resource$Params{})
+c, err := customer.New(&stripe.CustomerParams{})
```
jar-stripe, line 193: just confirming, did you verify that all of these compile?

jar-stripe (approved):
Looks good!  Approved, with a couple of comments/suggestions I would like to see addressed.

README.md
mbroshi-stripe: Yep, verified!

PR #2008, opened 2025-03-28
Add V2 support to beta

mbroshi-stripe:
### Why?

In preparation for GA of V2 suppot in `stripe-go`, we will be releasing V2 support on the `beta` branch.

### What?

- Adds Go V2 support to `beta`

## Changelog
- Add Go V2 support: V2 resources, services, events, and errors


amount.go
```diff
@@ -0,0 +1,7 @@
+package stripe
+
+// Amount describes a monetary amount in a specific currency in minor units.
+type Amount struct {
+	Currency Currency `json:"currency"`
+	Value    int64    `json:"value"`
+}
```
ramya-stripe, line 7: This is part of the breaking change right? Can we add a changelog entry in the PR description?

api_version.go
```diff
@@ -7,5 +7,6 @@
 package stripe
 
 const (
-	apiVersion string = "2025-03-31.basil"
+	apiVersion     string = "2025-03-31.basil"
+	previewVersion string = "2025-03-31.basil"
```
ramya-stripe, line 11: This constant should not be needed. Are we adding it in other languages. I know we have it in the next repo but we shouldnt need here in the beta or GA side of things

client/api.go
```diff
@@ -163,330 +163,409 @@ import (
...
+	// AppsSecrets is the client used to invoke /v1/apps/secrets APIs.
 	AppsSecrets *appssecret.Client
-	// Balance is the client used to invoke /balance APIs.
+	// Balance is the client used to invoke /v1/balance APIs.
 	Balance *balance.Client
-	// BalanceTransactions is the client used to invoke /balance_transactions APIs.
+	// BalanceTransactions is the client used to invoke /v1/balance_transactions APIs.
 	BalanceTransactions *balancetransaction.Client
-	// BankAccounts is the client used to invoke bankaccount related APIs.
+	// BankAccounts is the client used to invoke /v1/accounts/{account}/external_accounts APIs.
```
ramya-stripe, line 215: haha - nice! Was this manually done or generated?

client/event.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// ParseThinEvent parses a Stripe event from the payload and verifies its signature.
+// It returns a ThinEvent object and an error if the parsing or verification fails.
+func (a *API) ParseThinEvent(payload []byte, header string, secret string) (*stripe.ThinEvent, error) {
+	return a.ParseThinEventWithTolerance(payload, header, secret, webhook.DefaultTolerance)
+}
+
+// ParseThinEventWithTolerance parses a Stripe event from the payload and verifies its signature with a custom tolerance.
+// Generally, you should use ParseThinEvent, which uses the default tolerance.
```
ramya-stripe, line 18: Not in this PR, but as a follow up - we should add a note on what this tolerance is or add a link to some doc that talks about it. Check if other languages can also use the same treatment. Am guessing we need it for the snapshot event parsing as well

api_version.go
mbroshi-stripe: Good catch! I'll remove that

thinevent.go
```diff
@@ -0,0 +1,39 @@
...
+
+type RelatedObject struct {
+	ID   string `json:"id"`
+	Type string `json:"type"`
+	URL  string `json:"url"`
+}
+
+type ThinEventReason struct {
+	// Information on the API request that instigated the event
+	Request EventRequest `json:"request"`
```
ramya-stripe, line 36: Where is this EventRequest defined?

client/api.go
ramya-stripe: What I meant was that we replaced a generic statement like "blah related APIs" with the actual link - that is good

thinevent.go
```diff
@@ -0,0 +1,39 @@
...
+	// Livemode indicates if the event is from a production(true) or test(false) account
+	Livemode bool `json:"livemode"`
+	// Time at which the event was created
+	Created time.Time `json:"created"`
+	// [Optional] Object containing the reference to API resource relevant to the event
+	RelatedObject *RelatedObject `json:"related_object"`
+	// [Optional] Authentication context needed to fetch the event or related object
+	Context *string `json:"context"`
+	// [Optional] Reason for the event
+	Reason *ThinEventReason `json:"reason"`
```
ramya-stripe, line 25: Where is ThinEventReason defined? How is it different from V2EventReasonType defined in v2_event.go

iter.go
```diff
@@ -139,3 +127,100 @@ func reverse(a []interface{}) {
 		a[i], a[len(a)-i-1] = a[len(a)-i-1], a[i]
 	}
 }
+
+// Seq2 is the same as the iter.Seq2 type in Go 1.23+. It is used as the return type
```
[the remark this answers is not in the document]
mbroshi-stripe: It is literally the same type, but `iter.Seq2` is only available in Go 1.23+.

thinevent.go
mbroshi-stripe: It's in `event.go`: https://github.com/stripe/stripe-go/blob/master/event.go#L307

thinevent.go
mbroshi-stripe: Good catch! I should point instead to `V2EventReason`

thinevent.go
mbroshi-stripe: But, I should remove this and point to `V2EventReason`!

error.go
```diff
@@ -303,6 +379,48 @@ func (e *Error) Unwrap() error {
 	return e.Err
 }
 
+// canRetry implements the retrier interface.
+func (e *Error) canRetry() bool {
```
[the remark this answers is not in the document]
mbroshi-stripe: I needed to add a method so `stripe.go` could know when to retry an error. See https://github.com/stripe/stripe-go-next/pull/18/files#diff-f34528b48beca228ad5851464b3e20496f8c34c1f7b663e7f109fa9f5ed02fc4R1145-R1147

error.go
```diff
@@ -1,27 +1,59 @@
...
 
-// ErrorCode is the list of allowed values for the error's code.
-type ErrorCode string
+// errorTypes: The end of the section generated from our OpenAPI spec
 
 // DeclineCode is the list of reasons provided by card issuers for decline of payment.
 type DeclineCode string
 
+// ErrorCode is the list of allowed values for the error's code.
+type ErrorCode string
```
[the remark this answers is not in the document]
mbroshi-stripe: No, this was never generated. I just moved it next to the generated part.

error.go
```diff
@@ -276,7 +353,6 @@ type Error struct {
 
 	HTTPStatusCode    int               `json:"status,omitempty"`
 	Msg               string            `json:"message"`
-	DeveloperMsg      string            `json:"developer_message,omitempty"`
```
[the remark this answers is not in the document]
mbroshi-stripe: This was removed in `master`. I think it just got picked up in my PR.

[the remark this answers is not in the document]
ramya-stripe: Can you point to the PR where we removed this from master? I am looking at https://github.com/stripe/stripe-go/blob/master/error.go and it hasnt had a non generated commit in a long while. From what I remember, `developer_message` is a new property on error in v2 APIs

client/event.go
mbroshi-stripe: Created a Jira ticket to track this here: https://jira.corp.stripe.com/browse/DEVSDK-2470

ramya-stripe (approved):

PR #2014, opened 2025-04-02
Handle top-level External Account service

mbroshi-stripe:
### Why?

Stripe recently added a `/v1/external_accounts` endpoint that used to be under `/v1/account/%s/external_accounts`. The endpoint returns a union type of either `Card` or `BankAccount`. Go does not have a great way to deal with union types, so in the past for the child endpoints, we created separate `Card` and `BankAccount` services. For the top-level External Account service, we decided to have a single service with mirrored methods (e.g. both `GetCard` and `GetBankAccount`).

### What?

- Adds `externalaccount/client.go` with separate CRUDL operations for each of Card and BankAccount.

## Changelog
- Add support for top-level External Account CRUDL calls by adding separate `GetCard`/`GetBankAccount`, `NewCard`/`NewBankAccount`, etc. calls.

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -54,6 +54,7 @@ import (
 	entitlementsfeature "github.com/stripe/stripe-go/v82/entitlements/feature"
 	"github.com/stripe/stripe-go/v82/ephemeralkey"
 	"github.com/stripe/stripe-go/v82/event"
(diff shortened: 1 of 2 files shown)


jar-stripe (approved):
LGTM

PR #2015, opened 2025-04-02
require beta SDKs match event versions exactly when deserializing

xavdid-stripe:
### Why?

We decided that beta SDKs should exactly match the API version for events they're deserializing to minimize the chances of accidental breaking changes. This PR tweaks the version matching function to accommodate that change

### What?

- updated `isCompatibleAPIVersion` signature and implementation
- added tests

### See Also

- [RUN_DEVSDK-1556](https://go/j/RUN_DEVSDK-1556)

## Changelog

- ⚠️ Public Preview SDKs now require an exact API version match when deserializing webhooks

### DIFF
--- a/webhook/client.go
+++ b/webhook/client.go
@@ -187,7 +187,7 @@ type signedHeader struct {
 // Private functions
 //
 
-func isCompatibleAPIVersion(eventApiVersion string) bool {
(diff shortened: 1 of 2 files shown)


prathmesh-stripe (approved):
LGTM. We would also need this to be merged with the beta branch.

jar-stripe (approved):
👍

PR #2022, opened 2025-04-10
Add more stripe refs

mbroshi-stripe:
### Why?

I'm handling two separate issues in this one PR:
* To support V2 APIs, we will need the convenience method `stripe.Time`
* To better support setting enum params, we can extend `stripe.String` to turn a `string` enum into a `*string`.

### What?

* Adds `stripe.Time` and `stripe.TimeValue` functions which will be useful in setting `time.Time` values in V2 APIs
* Updated `stripe.String` to accept an enum, which makes setting enum params less verbose. For example:
```
stripe.PaymentIntentParams{
    Amount: stripe.Int64(2000),
    Currency: stripe.String(stripe.CurrencyUSD),
    ...
}
```
instead of 
```
stripe.PaymentIntentParams{
    Amount: stripe.Int64(2000),
    Currency: stripe.String(string(stripe.CurrencyUSD)),
    ...
}
```

## Changelog
* You can now pass `string` enums into `stripe.String`. For example, `stripe.String(stripe.CurrencyUSD)` instead of `stripe.String(string(stripe.CurrencyUSD))`


stripe.go
```diff
@@ -1413,6 +1414,20 @@ func StringSlice(v []string) []*string {
 	return out
 }
 
+// Time returns a pointer to the time.Time value passed in.
+func Time(v time.Time) *time.Time {
```
jar-stripe, line 1418: why is this preferable to just taking the address of the time value in the calling code?

mbroshi-stripe: If the time value comes from a function, you cannot directly take the address. For example, when I was generating code snippets, I wanted to use the `time.Now()` literal, but in Go you cannot take the address of the return value of a function (i.e., `&time.Now()` does not compile). This function allows me to do `stripe.Time(time.Now())`.

mbroshi-stripe: Discussed offline, and we decided this is safe from a memory management point of view since Go escape analysis will not garbage collect the pointer returned from `Time`.

jar-stripe (approved):
LGTM!

PR #2024, opened 2025-04-12
Add V2 support

mbroshi-stripe:
### Why?

Release GA support for V2. This is the `master`-branch analog of https://github.com/stripe/stripe-go/pull/2008

### What?

- Adds Go V2 support

## Changelog
- Add first-class Go V2 support: V2 resources, services, events, and errors.


example/v2/recipient_payout/recipient_payout.go
jar-stripe, line 1: just to confirm, are we removing this example because this is GA?

also, separate, but we have some standard examples in other repos that were launched around ubb/2024-09-30: meter event stream, and then a thinevent web hook example.  can we add those for go before we go live?

jar-stripe (approved):
Looks good!

example/v2/recipient_payout/recipient_payout.go
mbroshi-stripe: I'll fix the example and re-submit with the newer V2 pattern (along with other examples) in a separate PR

PR #2031, opened 2025-04-17
Add auto-pagination for search

mbroshi-stripe:
### Why?

This is a follow-up to https://github.com/stripe/stripe-go/pull/2027 in which we laid the groundwork for cleaner `List` pagination for V1 services. This PR does the same for `Search` pagination.

### What?

- Adds an unexported (i.e., internal-only) type `v1SearchList` with an `All `method for pagination. This is not currently used anywhere, but will be leveraged by `stripe.Client` in a subsequent PR.
- Also made a minor simplification to `iter.go` that I noticed while re-implementing the same logic in `search_iter.go`. That is, I replaced `new(T)` with `nil` when calling `yield` on an `error`.

### DIFF
--- a/search_iter.go
+++ b/search_iter.go
@@ -124,14 +124,11 @@ func GetSearchIter(container SearchParamsContainer, query SearchQuery) *SearchIt
 	return iter
 }
 
-// SearchIter provides a convenient interface
-// for iterating over the elements
-// returned from paginated search API calls.
-// Successive calls to the Next method
-// will step through each item in the search results,
-// fetching pages of items as needed.
-// Iterators are not thread-safe, so they should not be consumed
-// across multiple goroutines.
+// v1SearchList provides a convenient interface for iterating over the elements
+// returned from paginated list API calls. It is meant to be an improvement
+// over the SearchIter type, which was written before Go introduced generics and iter.Seq2.
+// Calling the `All` allows you to iterate over all items in the list,
+// with automatic pagination.
 type v1SearchList[T any] struct {
 	cur             *T
 	err             error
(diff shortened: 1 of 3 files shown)


search_iter.go
```diff
@@ -123,3 +123,98 @@ func GetSearchIter(container SearchParamsContainer, query SearchQuery) *SearchIt
...
+	}
+}
+
+// Next advances the SearchIter to the next item in the search results,
+// which will then be available
+// through the Current method.
+// It returns false when the iterator stops
+// at the end of the search results.
+func (it *v1SearchList[T]) next() bool {
+	if len(it.values) == 0 && it.meta.HasMore && !it.searchParams.Single {
```
xavdid-stripe, line 167: The structure of this block reads a little confusingly. Like, it _seems_ like we check `if len(it.values) == 0` twice, so those could maybe be combined? We've also got `if it.meta.NextPage != nil`, so it's not clear why that's not included in the top-level `if`. There's probably a good reason, but a comment that really explains the flow of logic here would go a long way.

(I see the comment on the whole method, but that doesn't seem to explain the actual logical steps)

search_iter.go
```diff
@@ -123,3 +123,98 @@ func GetSearchIter(container SearchParamsContainer, query SearchQuery) *SearchIt
...
+
+// GetSearchIter returns a new SearchIter for a given query and its options.
+func newV1SearchList[T any](container SearchParamsContainer, query v1SearchQuery[T]) *v1SearchList[T] {
+	var searchParams *SearchParams
+	formValues := &form.Values{}
+
+	if container != nil {
+		reflectValue := reflect.ValueOf(container)
+
+		// See the comment on Call in stripe.go.
```
xavdid-stripe, line 201: IMO it's worth duplicating that comment here for readability.

xavdid-stripe (approved):
Had some nits about comments but nothing blocking. Also there's a lint failure that's blocking the merge

search_iter.go
mbroshi-stripe: FWIW, this is literally a copy-paste of the existing search iter code, which I presume has gotten a lot of usage and any edge cases have been addressed. Not that comments are bad! Just that I figured I didn't want to break anything.

mbroshi-stripe: I combined the `if it.meta.NextPage != nil` into the parent `if`. We can't combine `if len(it.values) == 0` calls without having nested `if`'s, which are not idiomatic Go.

PR #2039, opened 2025-04-22
Fix event data unmarshaling

mbroshi-stripe:
### Why?

Currently, we are not correctly unmarshaling V2 event `Data`, and the resulting event will have an empty `Data` struct.

### What?

- Fixes V2 event data unmarshaling

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,7 @@
 {
   // Show the repo name in the top window bar.
-  "window.title": "${rootName}${separator}${activeEditorMedium}"
+  "window.title": "${rootName}${separator}${activeEditorMedium}",
+  "go.alternateTools": {
+      "go": "~/.local/share/mise/installs/go/latest/bin/go"
+  }
 }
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,7 @@
 {
   // Show the repo name in the top window bar.
-  "window.title": "${rootName}${separator}${activeEditorMedium}"
-}
+  "window.title": "${rootName}${separator}${activeEditorMedium}",
+  "go.alternateTools": {
+      "go": "~/.local/share/mise/installs/go/latest/bin/go"
+  }
```
mbroshi-stripe, line 6: I needed to add this to get Go working in VS code after I upgraded my OS 🤷

xavdid-stripe (approved):
good catch!

.vscode/settings.json
mbroshi-stripe: One change I made was removing the `mise` setup stuff from my `~/.bashrc` because starting a new shell took forever. Could that be the issue?

PR #2040, opened 2025-04-23
Add related object to ping test

mbroshi-stripe:
### Why?

In an upcoming change, the ping event will get its own `Event` object, which will have a related object. The `TestEventDestinationPing` will fail once we pull that change.

### What?

- Adds `RelatedObject` data to the `TestEventDestinationPing` unit test


jar-stripe (approved):
👍

PR #2042, opened 2025-04-23
Embed `V2BaseEvent` in V2 events instead of `V2RawEvent`

mbroshi-stripe:
### Why?

Tests are failing on `master` because we have a few conflicting in-flight changes for Events coming from codegen. This PR picks out the changes that are needed to get tests passing again.

### What?

- Updates events to embed `V2BaseEvent` instead of `V2RawEvent`
- Adds `json` tags to V2 events
- Updates a test to reflect the above

### DIFF
--- a/v2/core/event/client_test.go
+++ b/v2/core/event/client_test.go
@@ -18,15 +18,10 @@ func TestEventGet(t *testing.T) {
 	params := stripe.V2CoreEventParams{}
 	testServer, sc := mock.Server(t, string(http.MethodGet), "/v2/core/events/evt_123", nil, func(p *stripe.V2CoreEventParams) []byte {
 		data, err := json.Marshal(stripe.V1BillingMeterErrorReportTriggeredEvent{
-			V2RawEvent: stripe.V2RawEvent{
-				V2BaseEvent: stripe.V2BaseEvent{
-					Created: timeNow,
-					ID:      "evt_123",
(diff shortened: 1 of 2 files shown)


v2/core/event/client_test.go
```diff
@@ -18,15 +18,10 @@ func TestEventGet(t *testing.T) {
 	params := stripe.V2CoreEventParams{}
 	testServer, sc := mock.Server(t, string(http.MethodGet), "/v2/core/events/evt_123", nil, func(p *stripe.V2CoreEventParams) []byte {
 		data, err := json.Marshal(stripe.V1BillingMeterErrorReportTriggeredEvent{
-			V2RawEvent: stripe.V2RawEvent{
-				V2BaseEvent: stripe.V2BaseEvent{
-					Created: timeNow,
-					ID:      "evt_123",
-					Type:    "v1.billing.meter.error_report_triggered",
-				},
-				RelatedObject: &stripe.RelatedObject{
```
jar-stripe, line 27: v1.billing.meter.error_report_triggered should have a related object, right?

mbroshi-stripe: Yep! This removes the related object from the embedded `RawEvent`, which no longer exists.

jar-stripe: oh i see, it was already in the event subtype; github just hid that because it wasn't changing (right?)

ramya-stripe (approved):

PR #2047, opened 2025-04-25
Add stripe.Client

mbroshi-stripe:
### Why?

To support a much cleaner, less error-prone, and more idiomatic way to interact with the Stripe API, we are introducing the `stripe.Client` type to replace the previous `client.API`.

### What?

- Adds `stripe.Client`, which delegates out to the individual Stripe services.

These code snippets give a sense for the updated behavior:

**Before**:
```go
sc := client.New(stripeKey, nil)
sc.Customers.New(&stripe.CustomerParams{
	Name: stripe.String("Jenny Rosen"),
	Params: stripe.Params{
		Context: ctx,
	},
}
```

**After**
```go
sc := stripe.NewClient(stripeKey)
sc.V1Customers.Create(ctx, &stripe.CustomerCreateParams{
	Name: stripe.String("Jenny Rosen"),
})
```

Some of the improvements include:
1. Better, more descriptive naming `stripe.Client` instead of `client.API`
2. Methods take method-specific params (`CustomerCreateParams` instead of `CustomerParams`) to avoid runtime errors for sending the wrong fields
3. Methods take `context.Context` as the first argument instead of buried in `Params`
4. Methods are named like the API for more clarity (`Create` instead of `New`, `Retrieve` instead of `Get`, and `Delete` instead of `Del`)
5. All services are namespaced instead of just V2 services for better symmetry

## Changelog
* Introduces the `stripe.Client` to provide a more idiomatic, more consistent, and less error-prone pattern to call Stripe APIs. See the README for details.


scripts/check_api_clients/main.go
mbroshi-stripe, line 1: This only made sense when we were manually maintaining the client files. They're all autogenerated now.

webhook_service.go
mbroshi-stripe, line 1: This is mostly a copy-paste from`webhook/client.go`

oauth_service.go
mbroshi-stripe, line 1: This is mostly a copy-paste from `oauth/client.go`

webhook_service.go
helenye-stripe: In other languages, we did not end up adding a WebhookService, just a `ConstructEvent` method, since a webhook itself isn't actually an object in the API. Did we decide that this would make more sense for Go?

helenye-stripe (commented):
Could you also add a test to show how this is meant to look in the SDK itself?

webhook_service.go
mbroshi-stripe: Maybe I'll rename the file then. There's no actual `WebhookService`, just a collection of functions (like `ConstructEvent`) that I had to lift from the `webhook` package up into the `stripe` package to avoid a circular dependency caused by `ParseThinEvent`. I followed the same name convention I was using (and was honestly confused why the old one was called `client.go`)

helenye-stripe:
The usage! But I realized that this is coming in the generated examples PR, so you can disregard.

webhooks.go
```diff
@@ -0,0 +1,331 @@
...
+// signature doesn't match, or if the timestamp for the signature is older than
+// DefaultTolerance.
+//
+// NOTE: Stripe will only send Webhook signing headers after you have retrieved
+// your signing secret from the Stripe dashboard:
+// https://dashboard.stripe.com/webhooks
+//
+// This will return an error if the event API version does not match the
+// APIVersion constant.
+func ConstructEvent(payload []byte, header string, secret string) (Event, error) {
```
helenye-stripe, line 66: Is this callable from the client? This is what I was talking about before. (I also realize that this applies to the client.API as well.

helenye-stripe (approved):
Approving since we can have the webhook changes as additive changes, but we do definitely want parity between legacy and client-based!

webhooks.go
mbroshi-stripe: It's not, but I'm adding function on the client as we speak!

PR #2076, opened 2025-06-09
Switch to use generated API versions and add major/monthly constants

helenye-stripe:
### What?

Export the variables from the generated `api_version.go`, for easier future iteration. This should be a no-op for users.

## Changelog
* Export constants for the major and monthly API versions
  * e.g. `2025-05-28.basil` has major version `basil` and monthly version `2025-05-28`

### DIFF
--- a/api_version.go
+++ b/api_version.go
@@ -7,5 +7,7 @@
 package stripe
 
 const (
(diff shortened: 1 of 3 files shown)


mbroshi-stripe (approved):
Thanks for updating this!

PR #2082, opened 2025-06-30
Add form information to amount

helenye-stripe:
### Why?

Bug fix, this was causing panic on v2 endpoints that use this type.


mbroshi-stripe (approved):
Great catch! Thanks!

PR #2091, opened 2025-07-28
removed useless nil checks

XerWoho:
### Why?
There were some nil == nil, checks which were annoying, especially because the go compiler did not stop crying about it in my IDE.

### What?
Removed 3 useless nil checks from 3 files (paymentsource_service.go, feerefund_service.go, cashbalance_service.go).  These nil checks were checking if the params was nil, even though it already checked it previously.

### See Also
/

### DIFF
--- a/cashbalance_service.go
+++ b/cashbalance_service.go
@@ -37,9 +37,7 @@ func (c v1CashBalanceService) Update(ctx context.Context, params *CashBalanceUpd
 		return nil, fmt.Errorf(
 			"params cannot be nil, and params.Customer must be set")
 	}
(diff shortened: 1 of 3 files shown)


jar-stripe:
Hi @XerWoho , thanks for submitting this!  Unfortunately we cannot accept this as is because it changes files produced using a code generator.  Can you file this as an issue and we can take a look at fixing the generator for these files?

Please also take a look at our contributor guidelines in https://github.com/stripe/stripe-go/blob/master/CONTRIBUTING.md for more information about our codebase and how to contribute.  Thanks again!

Issue #2093, opened 2025-07-28
Unable to create BankAccount


mberezhnoy:
### Describe the bug

Error on create BankAccount: `parse \"https://api.stripe.com/v1/accounts/btok_456/external_accounts%!(EXTRA string=, string=acct_123)\": invalid URL escape \"%!(\"\n`

The OpenAPI–generated [code](https://github.com/stripe/stripe-go/blob/master/bankaccount_service.go#L28-L30) for creating bank accounts builds a URL with one `%s` but supplies three value, resulting in an invalid escape sequence in the path

### To Reproduce

call `client.V1BankAccounts.Create(ctx, params)` with correct data

### Expected behavior

create new BankAccount

### Code snippets

```Go
// var client *stripe.Client
params := &stripe.BankAccountCreateParams{
  Account: stripe.String("acct_123"),
  Token:   stripe.String("btok_456"),
}
_, err := client.V1BankAccounts.Create(ctx, params)
```

### OS

macOS

### Go version

go1.23.2 darwin/arm64

### stripe-go version

v82.2.1

### API version

2025-05-28

### Additional context

It can still be created via the deprecated client.API
```
import (
	"github.com/stripe/stripe-go/v82"
	"github.com/stripe/stripe-go/v82/client"
)
var stripeClient *client.API
.......
params := &stripe.BankAccountParams{
	Account: stripe.String(accountId),
	Token:   stripe.String(token),
}
ba, err := stripeClient.BankAccounts.New(params)

```

mbroshi-stripe:
Thanks for reporting this issue! And sorry to hear about the issues you're facing with `stripe.Client`. We have a fix in the works, and in the meantime you can continue to use `client.API` (even though it's marked as deprecated).

mbroshi-stripe:
@mberezhnoy Thanks again for filing this issue. We have addressed it in the latest `stripe-go` release: https://github.com/stripe/stripe-go/releases/tag/v82.4.1

PR #2098, opened 2025-08-01
Mbroshi/run devsdk 1828/fix payment method services

mbroshi-stripe:
### Why?

In https://github.com/stripe/stripe-go/issues/2093, it was brought to our attention that there was a bug in the `stripe.Client.V1BankAccounts` service's `Create` method. It turns out there were actually a few bugs in both the `V1BankAccounts` and `V1Cards` services, which are non-standard Stripe services that we did not correctly port over to the new `stripe.Client` pattern.

### What?

- Adds custom parameter serialization to `BankAccountCreateParams` and `CardCreateParams`
- Fixes `Create` and `List` methods in `V1BankAccounts` and `V1Cards` services.
- Adds `nil`-checking to `Retrieve`, `Update`, and `Delete` methods  in `V1BankAccounts` and `V1Cards` services.
- Adds unit tests for both services

### See Also

* https://github.com/stripe/stripe-go/issues/2093

## Changelog
* Fixes bugs in `Create` and `List` methods in `V1BankAccounts` and `V1Cards` services in the `stripe.Client` pattern.

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -8,8 +8,9 @@ package stripe
 
 import (
 	"encoding/json"
-	"github.com/stripe/stripe-go/v82/form"
 	"strconv"
+
+	"github.com/stripe/stripe-go/v82/form"
 )
 
 // The type of entity that holds the account. This can be either `individual` or `company`.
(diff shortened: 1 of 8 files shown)


xavdid-stripe (approved):
PR looks good!

Could you be a little more specific than "fixes bugs" in your changelog? Namely, we want to communicate to a reader if they would have been affected and any action they need to take

mbroshi-stripe:
> Could you be a little more specific than "fixes bugs" in your changelog? Namely, we want to communicate to a reader if they would have been affected and any action they need to take

Fair point. In this case, these methods were basically broken, so any user attempting to use them would have received an error like that reported in https://github.com/stripe/stripe-go/issues/2093. Not sure how much gory detail to add there, but I will mention that they were previously broken altogether.

PR #2107, opened 2025-08-20
add missing expand param in `get /v1/accounts/{account}/external_accounts`

xavdid-stripe:
### Why?

We didn't support an `expand` parameter where there should have been done.

### What?

- codegen for missing param

### See Also

RUN_DEVSDK-1881

### DIFF
--- a/bankaccount.go
+++ b/bankaccount.go
@@ -407,6 +407,8 @@ type BankAccountListParams struct {
 	// The identifier of the parent account under which the bank accounts are
 	// nested. Either Account or Customer should be populated.
 	Account *string `form:"-"` // Included in URL
+	// Specifies which fields in the response should be expanded.
+	Expand []*string `form:"expand"`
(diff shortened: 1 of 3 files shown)


xavdid-stripe:
The PR title will be used as the changelog entry. Is there additional context you think I should provide?

prathmesh-stripe (approved):

prathmesh-stripe:
@xavdid-stripe I was thinking of adding something like the opposite of a [changelog entry](https://github.com/stripe/stripe-go/blob/master/CHANGELOG.md#76130---2024-01-18) we have for `Expand`:
```
Add support for `Expand` on `BankAccountListParams` and `CardListParams`
```

This is cleaner than putting in the whole URL

xavdid-stripe:
sure, done! Odd that it got removed back then. I wonder if that was something to do with the overrides being wrong?

PR #2121, opened 2025-09-19
⚠️ Add strongly typed EventNotifications

xavdid-stripe:
### Why?

As part of the Next-Gen Event Handling project, we're adding fully typed classes to represent all the ~~thin events~~ Event Notifications a user could receive. There are also helper methods that help with the event handling experience.

### What?

- (see changelog below for user-facing changes)
- add tests
- update examples

### See Also

[DEVSDK-2662](https://go/j/browse/DEVSDK-2662)

## Changelog

We've overhauled how V2 Events are handled in the SDK! This approach should provide a lot more information at authoring and compile time, leading to more robust integrations. As part of this process, there are a number of changes to be aware of.

- ⚠️ Renamed the `ThinEvent` struct to `EventNotification`
- add the `UnknownEventNotification ` struct
- ⚠️ Removed the `relatedObject` property from `EventNotification`. It's on `UnknownEventNotification` instead.
- ⚠️ Renamed `Client.ParseThinEvent` to `Client.ParseEventNotification` and changed its method signature accordingly
- ⚠️ Removed `API.parseThinEvent`. Use `Client` instead (referring to the [migration guide](https://github.com/stripe/stripe-go/wiki/Migration-guide-for-Stripe-Client) if necessary).
- Added matching `EventNotification` structs for every v2 `Event`. For example, there's now a `V1BillingMeterErrorReportTriggeredEventNotification` to match the existing `V1BillingMeterErrorReportTriggeredEvent`. Each of these implements a `FetchEvent()` method to retrieve its corresponding event. For events with related objects, there's a `FetchRelatedObject()` method that performs the API call and casts the response to the correct type.


v2_events.go
```diff
@@ -219,3 +296,50 @@ func ConvertRawEvent(event *V2RawEvent, backend Backend, key string) (V2Event, e
 		return event, nil
 	}
 }
+
+// V2Events: The end of the section generated from our OpenAPI spec
+
+// EventNotificationFromJSON turns a JSON payload into the corresponding EventNotification, if possible.
+// WARN: Not meant for production use! It doesn't do any signature validation. It's useful as a test util though.
+func EventNotificationFromJSON(payload []byte, client Client) (EventNotificationContainer, error) {
```
jar-stripe, line 305: why make this public this if its not meant for production use?  do we need to reference it from another location internally?

client/event.go
```diff
@@ -9,20 +9,31 @@ import (
 )
 
 // ParseThinEvent parses a Stripe event from the payload and verifies its signature.
-// It returns a ThinEvent object and an error if the parsing or verification fails.
-func (a *API) ParseThinEvent(payload []byte, header string, secret string) (*stripe.ThinEvent, error) {
+// It returns a union of all possible event notification types that implement EventNotificationContainer.
+func (a *API) ParseThinEvent(payload []byte, header string, secret string) (stripe.EventNotificationContainer, error) {
```
jar-stripe: should be ParseEventNotification right?

client/event.go
```diff
@@ -9,20 +9,31 @@ import (
...
 	return a.ParseThinEventWithTolerance(payload, header, secret, webhook.DefaultTolerance)
 }
 
 // ParseThinEventWithTolerance parses a Stripe event from the payload and verifies its signature with a custom tolerance.
 // Generally, you should use ParseThinEvent, which uses the default tolerance.
-func (a *API) ParseThinEventWithTolerance(payload []byte, header string, secret string, tolerance time.Duration) (*stripe.ThinEvent, error) {
+func (a *API) ParseThinEventWithTolerance(payload []byte, header string, secret string, tolerance time.Duration) (stripe.EventNotificationContainer, error) {
```
jar-stripe: should be ParseEventNotificationWithTolerance?

client/event.go
```diff
@@ -9,20 +9,31 @@ import (
...
 	}
-	var event stripe.ThinEvent
+	var event stripe.V2UnknownEventNotification
 	if err := json.Unmarshal(payload, &event); err != nil {
 		return nil, err
 	}
-	return &event, nil
+
+	switch event.Type {
+	case "v1.billing.meter.error_report_triggered":
```
jar-stripe: why is this special cased?  and do we need to generate these special cases?

jar-stripe (commented):
couple of errant "ThinEvent*" methods and other comments to address

client/event.go
xavdid-stripe: yep, good call

v2_events.go
xavdid-stripe: Users may also find it helpful to generate event notifications. e.g. they have a function that takes a `V1BillingMeterErrorReportTriggeredEventNotification` and want to generate them in tests. Convenient to not have to go sign a payload. Not super important though

xavdid-stripe: Now that i'm saying it out loud, it may need to be public anyway to be used in our tests since they're in a different package.

client/event.go
xavdid-stripe: actually this whole method may need to come out? I was going to ask michael. It duplicates the method on the client, but doesn't supply a client. I think I accidentally left it half finished, but events returned from this method won't be able to make requests.

v2_events.go
jar-stripe: thats fair; as an alternative, would it be any better to provide a sentinal dummy signature value so users can use the production-ready method?  or, can we preface this with whatever is standard in go for "UNSAFE" or private or whatever?  i am concerned that folks will miss the comment and see the name and think that's the blessed way to parse event notifications.

client/event.go
xavdid-stripe: ok, confirmed the whole thing should come out. It's on the legacy approach and parsing thin events is meant to be new stuff only. i'll note in changelog.

```go
sc := &client.API{} // deprecated
sc.Init("sk_test", nil)
sc.ParseThinEvent("payload", "", "")
```

xavdid-stripe:
(ignore red CI; docker hub is having a problem so all tests are down. Everything besides compat was good before i merged master back into this and trigger the CI run.)

v2_events.go
jar-stripe: just to confirm, we decided to rol with this as stated right?  if so, we should fix the comment

jar-stripe (approved):
Couple small nits but overall looks good!

v2_events.go
xavdid-stripe: that's correct. We'll ship it as is. i'll update comment.

v2_events.go
```diff
@@ -68,18 +69,65 @@ type V1BillingMeterErrorReportTriggeredEvent struct {
...
+// Occurs when a Meter has invalid async usage events.
+type V1BillingMeterErrorReportTriggeredEventNotification struct {
+	V2EventNotification
+	RelatedObject RelatedObject `json:"related_object"`
+}
+
+// GetEventNotification ensures we conform to `EventNotificationContainer`.
+func (en *V1BillingMeterErrorReportTriggeredEventNotification) GetEventNotification() *V2EventNotification {
+	return &en.V2EventNotification
+}
```
mbroshi-stripe, line 87: See comment above. You shouldn't need to generate these getters if you inherit the `GetEventNotification` on the embedded `V2EventNotification`.

event_notification.go
```diff
@@ -0,0 +1,79 @@
...
+type V2UnknownEventNotification struct {
+	V2EventNotification
+
+	// [Optional] Object containing the reference to API resource relevant to the event
+	RelatedObject *RelatedObject `json:"related_object"`
+}
+
+func (en *V2UnknownEventNotification) GetEventNotification() *V2EventNotification {
+	return &en.V2EventNotification
+}
```
mbroshi-stripe, line 62: The pattern we've followed in the past is to put this `Get` method on `V2EventNotification` itself so that any type embedding it will "inherit" that method and not need to re-implement it.

event_notification.go
```diff
@@ -0,0 +1,79 @@
...
+
+type V2UnknownEventNotification struct {
+	V2EventNotification
+
+	// [Optional] Object containing the reference to API resource relevant to the event
+	RelatedObject *RelatedObject `json:"related_object"`
+}
+
+func (en *V2UnknownEventNotification) GetEventNotification() *V2EventNotification {
```
mbroshi-stripe, line 60: Nit: Function receivers generally just use a single letter (`e` or `n`).

v2_events.go
```diff
@@ -219,3 +296,51 @@ func ConvertRawEvent(event *V2RawEvent, backend Backend, key string) (V2Event, e
 		return event, nil
 	}
 }
+
+// V2Events: The end of the section generated from our OpenAPI spec
+
+// EventNotificationFromJSON is a helper for constructing an Event Notification. Doesn't perform signature validation,
+// so you should use [Client.ParseEventNotification] instead for initial handling.
+// This is useful in unit tests and working with EventNotifications that you've already validated the authenticity of.
+func EventNotificationFromJSON(payload []byte, client Client) (EventNotificationContainer, error) {
```
mbroshi-stripe, line 305: Do we need make this function exported? You explain its usage in the comment, but I feel it might cause confusion

xavdid-stripe: We do, yes. It's useful as a test util for both us and others. 

I can't make it unexported because `stripe_test` and `stripe` are different packages and we're using it in our own tests. It's a useful util for users as well. Jesse and I talked through it in a different PR (Java, IIRC)

event_notification.go
xavdid-stripe: are changing this and removing the extra `GetEventNotification` breaking changes? If not, I'll probably circle back after release.

mbroshi-stripe: Nope, those are not breaking changes. They can wait.

xavdid-stripe: perfect, i'll make a note. thanks for looking!

PR #2133, opened 2025-09-26
Add `StripeContext` object

xavdid-stripe:
### Why?

As the `stripe-context` header is evolving, we want to give better tools for creating and managing contexts. This PR adds a new class, `StripeContext`. It can be used anywhere the context option is supplied and gets serialized to a string when making requests. It's also found on the `EventNotification.context` property.

### What?

- add `StripeContext` class
- allow it to be supplied for RequestOptions.context; serialize it into strings
- add tests

## Changelog

- Add the `stripe.Context` struct
- ⚠️ Change `EventNotification` (formerly known as `ThinEvent`)'s `context` property from `string` to `stripe.Context`


params.go
```diff
@@ -298,6 +298,13 @@ func (p *Params) SetStripeContext(val string) {
 	p.StripeContext = &val
 }
 
+// SetStripeContextObject sets a value for the Stripe-Context header using a StripeContext object.
+func (p *Params) SetStripeContextObject(val *StripeContext) {
+	if val != nil {
+		p.StripeContext = val.String()
+	}
+}
+
```
mbroshi-stripe: I might have left a similar comment in Java, but did we want to have separate methods `SetStripeContext` and `SetStripeContextObject`? I would prefer just the single method. Also, I think it should take a `StripeContext` not a `*StripeContext` (so you don't need a `nil`-check), but don't feel too strongly there

event_notification.go
```diff
@@ -57,23 +60,21 @@ type V2UnknownEventNotification struct {
...
-func (en *V2UnknownEventNotification) FetchRelatedObject(ctx context.Context) (*APIResource, error) {
-	if en.RelatedObject == nil {
+func (n *V2UnknownEventNotification) FetchRelatedObject(ctx context.Context) (*APIResource, error) {
+	if n.RelatedObject == nil {
 		return nil, nil
 	}
 
 	// TODO: usage?
 	obj := &APIResource{}
+	params := &eventNotificationParams{}
```
mbroshi-stripe: Looks like you lost setting the `context.Context`. Was that intentional?

stripe_context.go
```diff
@@ -0,0 +1,109 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// StripeContext provides an immutable container and convenience methods for interacting with the `Stripe-Context` header. All methods return a new instance of StripeContext.
+// You can use it whenever you're initializing a `StripeClient` or sending `stripe_context` with a request. It's also found in the `EventNotification.context` property.
+type StripeContext struct {
```
mbroshi-stripe: See https://go.dev/wiki/CodeReviewComments#package-names. The recommendation is to name the struct simply `Context` to avoid "stutter". Our users would then always refer to it as `stripe.Context` instead of the redundant `stripe.StripeContext`. That would also cascade to other names, e.g. `ParseStripeContext` --> `ParseContext`, etc.

stripe_context.go
```diff
@@ -0,0 +1,109 @@
...
+	if len(c.segments) == 0 {
+		return nil, fmt.Errorf("cannot pop from empty context")
+	}
+
+	return NewStripeContext(c.segments[:len(c.segments)-1]), nil
+}
+
+// String returns the string representation of the StripeContext.
+// Segments are joined with "/" as the separator.
+func (c *StripeContext) String() *string {
```
mbroshi-stripe: I would change the return type to `string`, which would then satisfy the `Stringer` interface: https://pkg.go.dev/fmt#Stringer

v2_events.go
```diff
@@ -81,25 +82,22 @@ type V1BillingMeterErrorReportTriggeredEventNotification struct {
 	RelatedObject RelatedObject `json:"related_object"`
 }
 
-// GetEventNotification ensures we conform to `EventNotificationContainer`.
-func (en *V1BillingMeterErrorReportTriggeredEventNotification) GetEventNotification() *V2EventNotification {
-	return &en.V2EventNotification
-}
-
```
mbroshi-stripe, line 88: Thanks for removing these extra methods!

stripe_context.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// StripeContext provides an immutable container and convenience methods for interacting with the `Stripe-Context` header. All methods return a new instance of StripeContext.
+// You can use it whenever you're initializing a `StripeClient` or sending `stripe_context` with a request. It's also found in the `EventNotification.context` property.
+type StripeContext struct {
+	segments []string
```
mbroshi-stripe: OOC, is there a reason you made this field unexported? Usually in Go, we just make the data exported, rather than having a separate `Segments()` method

stripe_context.go
```diff
@@ -0,0 +1,109 @@
...
+	return len(c.segments) == 0
+}
+
+// Size returns the number of segments in the StripeContext.
+func (c *StripeContext) Size() int {
+	return len(c.segments)
+}
+
+// Clone creates a deep copy of the StripeContext.
+func (c *StripeContext) Clone() *StripeContext {
```
mbroshi-stripe: I don't recall these extra methods in Java. Is there a reason you added them in Go?

params.go
xavdid-stripe: The goal is to support both inputs, so absent method overloading (or a union parameter) this seemed cleanest.

event_notification.go
xavdid-stripe: nope, good catch! Should probably have a test that caught that...

mbroshi-stripe (approved):
Looks great!

PR #2157, opened 2025-10-02
Improve event notification example

xavdid-stripe:
### Why?

I'm going back and sprucing up the new event notification handling examples. This will also go onto docs.stripe.com.

### DIFF
--- a/example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
+++ b/example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
@@ -46,26 +46,37 @@ func main() {
 		// Unmarshal the event data into an appropriate struct depending on its Type
 		switch evt := eventNotification.(type) {
 		case *stripe.V1BillingMeterErrorReportTriggeredEventNotification:
-			fmt.Printf("Got an event related to Meter ID: %s\n", evt.RelatedObject.ID)
+			// there's basic info about the related object in the notification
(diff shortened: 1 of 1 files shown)


example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
```diff
@@ -46,26 +46,39 @@ func main() {
...
+			// or you can fetch the full object form the API for more details
 			meter, err := evt.FetchRelatedObject(context.TODO())
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error fetching related object: %v\n", err)
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
-			fmt.Printf("Meter ID %s uses aggregation %s.", meter.ID, meter.DefaultAggregation)
+			sb := fmt.Sprintf("Meter %s (%s) had a problem", meter.DisplayName, meter.ID)
+			fmt.Println(sb)
```
mbroshi-stripe: I know it's just an example, but how about just `fmt.Printf("...\n")` here?

mbroshi-stripe (approved):

example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
xavdid-stripe: yeah fair!

Issue #2159, opened 2025-10-04
webhook.ConstructEvent fails for events where created is a ISO-8601 string


ReeceJones:
### Describe the bug

Calls to `webhook.ConstructEvent` fail, because of the following JSON unmarshalling error:
```
Failed to parse webhook body json: json: cannot unmarshal string into Go struct field Event.created of type int64
```
This is because the event provides a ISO-8601 date string, but stripe-go attempts to parse the string as if it were unix time.

### To Reproduce

1. Listen to events using the stripe cli
2. Consume the events using a local server
3. Process the event by calling `webhook.ConstructEvent`
4. Observe that the call to `webhook.ConstructEvent` will fail

### Expected behavior

`webhook.ConstructEvent` should successfully parse the provided event by correctly parsing the timestamp.

### Code snippets

```Go
Example event

{"id":"evt_test_123","object":"v2.core.event","type":"v2.core.account[configuration.merchant].capability_status_updated","livemode":false,"created":"2025-10-04T15:13:32.232083866Z","related_object":{"id":"acct_123","type":"v2.core.account","url":"/v2/core/accounts/acct_123?include=configuration.merchant"}


Example failing code

body = `{"id":"evt_test_123","object":"v2.core.event","type":"v2.core.account[configuration.merchant].capability_status_updated","livemode":false,"created":"2025-10-04T15:13:32.232083866Z","related_object":{"id":"acct_123","type":"v2.core.account","url":"/v2/core/accounts/acct_123?include=configuration.merchant"}`
stripeSignature := ""
stripeWebhookSecret := "whsec_123"

event, err := webhook.ConstructEvent(body, stripeSignature, stripeWebhookSecret)
```

### OS

macOS

### Go version

go1.24.4 darwin/arm64

### stripe-go version

v83.1.0-beta.1

### API version

2025-09-30.clover

### Additional context

I am unable to set the stripe cli to use the preview API version when listening to events. I am not sure if this is contributing to the issue.

mbroshi-stripe:
👋 Hi @ReeceJones ! Thanks for reporting this issue, and apologies for the confusion! The error you are getting is caused by trying to construct a snapshot event from a thin event. To get an overview of how thin events work, you can read https://docs.stripe.com/event-destinations. 

[...]

PR #2161, opened 2025-10-06
Return clearer error when calling `ConstructEvent` on an event notification

mbroshi-stripe:
### Why?

In https://github.com/stripe/stripe-go/issues/2159, a user got a confusing JSON unmarshaling error when trying to parse an event notification using the `ConstructEvent` function, which is only meant for snapshot events. Instead, we should return a clearer error that points the user to the correct function.

### What?

- Returns a clearer error when calling `ConstructEvent` on an event notification. This uses the fact that snapshot events have an `api_version` field, but event notifications do not.
- Adds tests for the new functionality.

## Changelog
- Returns a clearer error when calling `ConstructEvent` on an event notification.

### DIFF
--- a/example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
+++ b/example/v2/event_notification_webhook_handler/event_notification_webhook_handler.go
@@ -46,26 +46,38 @@ func main() {
 		// Unmarshal the event data into an appropriate struct depending on its Type
 		switch evt := eventNotification.(type) {
 		case *stripe.V1BillingMeterErrorReportTriggeredEventNotification:
-			fmt.Printf("Got an event related to Meter ID: %s\n", evt.RelatedObject.ID)
+			// there's basic info about the related object in the notification
+			fmt.Printf("Meter w/ id %s had a problem\n", evt.RelatedObject.ID)
+
+			// or you can fetch the full object form the API for more details
 			meter, err := evt.FetchRelatedObject(context.TODO())
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error fetching related object: %v\n", err)
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
-			fmt.Printf("Meter ID %s uses aggregation %s.", meter.ID, meter.DefaultAggregation)
+			fmt.Fprintf(os.Stdout, "Meter %s (%s) had a problem\n", meter.DisplayName, meter.ID)
 
+			// And you can always fetch the full event:
 			event, err := evt.FetchEvent(context.TODO())
 			if err != nil {
 				fmt.Fprintf(os.Stderr, "Error fetching event: %v\n", err)
 				w.WriteHeader(http.StatusInternalServerError)
 				return
 			}
-			fmt.Printf("Event data reason: %+v\n", event.Data.Reason)
-
+			fmt.Printf("More info: %s\n", event.Data.DeveloperMessageSummary)
 		case *stripe.UnknownEventNotification:
-			fmt.Printf("Received an event the SDK doesn't have types for: %s\n", evt.Type)
-			fmt.Printf("Event ID: %s; has related object? %t\n", evt.ID, evt.RelatedObject != nil)
+			// Events that were introduced after this SDK version release are
+			// represented as `UnknownEventNotification`s.
+			// They're valid, the SDK just doesn't have corresponding classes for them.
+			// You must match on the "type" property instead.
+			switch evt.Type {
+			case "some.new.event":
+				// you can still `.FetchEvent()` and `.FetchRelatedObject()`, but the latter may
+				// return `nil` if that event type doesn't have a related object.
+				return
+			}
+
 		default:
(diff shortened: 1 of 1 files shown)


webhook/client.go
```diff
@@ -218,6 +218,10 @@ func constructEvent(payload []byte, sigHeader string, secret string, options Con
 		return e, err
 	}
 
+	if err := checkThinEvent(payload); err != nil {
```
jar-stripe: should be EventNotification

webhook/client.go
```diff
@@ -227,7 +231,22 @@ func constructEvent(payload []byte, sigHeader string, secret string, options Con
...
+	e := struct {
+		APIVersion string `json:"api_version"`
+	}{}
+	if err := json.Unmarshal(payload, &e); err != nil {
+		return fmt.Errorf("Failed to parse webhook body json: %s", err.Error())
+	}
+
+	// Snapshot events have an API version field, but thin events do not
+	if e.APIVersion == "" {
+		return fmt.Errorf("Received thin event; use ParseEventNotification instead")
```
jar-stripe: should be "Received event notification; ..."

webhook/client.go
```diff
@@ -227,7 +231,22 @@ func constructEvent(payload []byte, sigHeader string, secret string, options Con
...
+
+func checkThinEvent(payload []byte) error {
+	e := struct {
+		APIVersion string `json:"api_version"`
+	}{}
+	if err := json.Unmarshal(payload, &e); err != nil {
+		return fmt.Errorf("Failed to parse webhook body json: %s", err.Error())
+	}
+
+	// Snapshot events have an API version field, but thin events do not
```
jar-stripe: thin events -> event notifications

webhook/client_test.go
```diff
@@ -325,3 +332,14 @@ func TestApiVersionCompatibility(t *testing.T) {
 		}
 	}
 }
+
+func TestConstructEvent_ErrorOnThinEvent(t *testing.T) {
```
jar-stripe: ThinEvent -> EventNotification

webhooks.go
```diff
@@ -168,6 +168,10 @@ func constructEvent(payload []byte, sigHeader string, secret string, cfg webhook
 		return e, err
 	}
 
+	if err := checkThinEvent(payload); err != nil {
```
jar-stripe: same comments as above re ThinEvent -> EventNotification, but also this looks like duplicated code.  can one file import the other?

webhooks_test.go
```diff
@@ -323,3 +331,14 @@ func TestApiVersionCompatibility(t *testing.T) {
 		}
 	}
 }
+
+func TestConstructEvent_ErrorOnThinEvent(t *testing.T) {
```
jar-stripe: ThinEvent -> EventNotification

webhooks.go
```diff
@@ -177,7 +181,22 @@ func constructEvent(payload []byte, sigHeader string, secret string, cfg webhook
 	}
 
 	return e, nil
+}
+
+func checkThinEvent(payload []byte) error {
```
jar-stripe: same comments as above

webhooks.go
mbroshi-stripe: It is duplicate code. Technically I'm putting in a fix to the deprecated `client.API` in this PR, but it's a small enough lift that it was worth it

webhook/client.go
```diff
@@ -227,7 +231,22 @@ func constructEvent(payload []byte, sigHeader string, secret string, options Con
...
+func checkEventNotification(payload []byte) error {
+	e := struct {
+		APIVersion string `json:"api_version"`
+	}{}
+	if err := json.Unmarshal(payload, &e); err != nil {
+		return fmt.Errorf("Failed to parse webhook body json: %s", err.Error())
+	}
+
+	// Snapshot events have an API version field, but event notifications do not
+	if e.APIVersion == "" {
```
xavdid-stripe: the `object` field is the canonical way to tell objects apart from one another. So while it's true `api_version` happens to work for v1 event vs v2 event today, the check should really be `object != "event"`

webhook/client.go
```diff
@@ -227,7 +231,22 @@ func constructEvent(payload []byte, sigHeader string, secret string, options Con
...
+		return fmt.Errorf("Received event notification; use ParseEventNotification instead")
```
xavdid-stripe: nit: `Received event notification` isn't super clear without context. Maybe something like 

> Unable to use `ConstructEvent` to parse an `EventNotification`. Use `client.ParseEventNotification` instead

xavdid-stripe (changes requested):
one quick thing about the discriminator, but looks good otherwise!

webhook/client.go
mbroshi-stripe: OK, thanks! I asked the events team and they suggested this check, but I'll change to checking on `object` instead.

webhook/client.go
mbroshi-stripe: Fair point--I'll use the suggested text. In general in Go, it's the caller's job to add the context, but I agree in this case it's probably best to include it.

xavdid-stripe:
thank you!

PR #2178, opened 2025-10-12
Add context-aware logging interface and update logger usage

### LINKED ISSUE #1281: Distributed tracing and other context in logging
The current logging interface does not seem to allow callers to inject structured fields, such as trace ids, into the log lines. While custom loggers are acceptable, only global instances are acceptable, and the logging interface only passes Stripe's log messages as arguments.

In my setup, I use `context.Context` to pass request ids into my logger, which in turn includes these request ids in the structured log lines that are written to stderr. Systems like OpenCensus/OpenTelemetry/OpenTracing also use context to pass correlation tokens through.

It would be nice if the Stripe SDK had support for this, so that I could more easily correlate Stripe's logging with incoming requests. While it is possible to do this to a certain extent by overriding the `http.Transaport`, I'd like to take full advantage of the diagnostic logging that is already built into this library.

I am curious if this has already been thought through, or if others have solutions. If not, it would be great to add support for this into the logging interfaces.

***

The current API, for reference:

https://github.com/stripe/stripe-go/blob/ebbff369b03d0cf6475352e7ae94b5684bcf0288/stripe.go#L181-L195

https://github.com/stripe/stripe-go/blob/ebbff369b03d0cf6475352e7ae94b5684bcf0288/log.go#L124-L142

lmcrean:
Adds `ContextLeveledLoggerInterface` to enable distributed tracing integration.

## Summary

- Adds new `ContextLeveledLoggerInterface` with context-aware logging methods
- Backend checks interface type on each log call and passes context when supported
- Fully backward compatible with existing `LeveledLoggerInterface` implementations

## Usage

```go
type MyLogger struct{}

func (l *MyLogger) Debugf(ctx context.Context, format string, v ...interface{}) {
    traceID := ctx.Value("trace_id")
    log.Printf("[%s] "+format, append([]interface{}{traceID}, v...)...)
}
// ... implement Infof, Warnf, Errorf similarly

stripe.DefaultLeveledLogger = &MyLogger{}
```

## Testing

```bash
go test ./...  # All pass
```

## Breaking Changes

None. Existing `LeveledLoggerInterface` implementations work unchanged.

Fixes #1281


mbroshi-stripe:
@lmcrean Ditto to my other comment: Thanks for this PR! 🙏 High-level, your approach looks great, but I haven't had a chance to give a careful read-through yet. I'll get to that later this week (on both PRs)

stripe.go
```diff
@@ -1623,6 +1693,18 @@ func newBackendImplementation(backendType SupportedBackend, config *BackendConfi
...
+	// Note: A type can only implement ONE of these interfaces due to conflicting
+	// method signatures (Debugf, etc. have different parameters)
+	var contextLogger ContextLeveledLoggerInterface
+	var regularLogger LeveledLoggerInterface
+
+	if cl, ok := config.LeveledLogger.(ContextLeveledLoggerInterface); ok {
+		contextLogger = cl
+	} else if rl, ok := config.LeveledLogger.(LeveledLoggerInterface); ok {
+		regularLogger = rl
+	}
```
mbroshi-stripe, line 1687: Do you think you should `else {panic()}`? If Go supported union types, this case wouldn't even compile, so it feels reasonable to me to `panic` here.

stripe.go
```diff
@@ -274,7 +279,9 @@ type BackendImplementation struct {
 	Type              SupportedBackend
 	URL               string
 	HTTPClient        *http.Client
-	LeveledLogger     LeveledLoggerInterface
+
+	// LeveledLogger holds the configured logger (either LeveledLoggerInterface or ContextLeveledLoggerInterface)
+	LeveledLogger     interface{}
```
mbroshi-stripe, line 282: We don't have control over how folks might be using this field. It seems possible someone might create a `BackendImplementation` "by hand" instead of using `NewBackends` or `NewBackendsWithConfig`, in which case their `LeveledLogger` will never be used. Can you be more defensive of that case in your `logf`, `warnf`, etc. methods?

stripe.go
lmcrean: Done. Added panic clause that fires when `config.LeveledLogger` is non-nil but doesn't implement either interface.

Won't panic for manual `BackendImplementation` instantiation (handled defensively in log methods).

Thanks for the suggestion.

stripe.go
lmcrean: Added fallback type assertions to all 4 logging methods (`logDebugf`, `logInfof`, `logWarnf`, `logErrorf`). Thanks again.

mbroshi-stripe:
@lmcrean I ran the CI checks, and there are some linter failures. You can run `just lint` locally to reproduce (which is basically just a wrapper around `staticcheck`). Can you address those failures?

lmcrean:
thanks @mbroshi-stripe  -- I've fixed. `just lint` is passing on local machine, can you re-run the CI please?

context_logger_test.go
```diff
@@ -149,11 +157,10 @@ func TestHelperMethodsWithNilContext(t *testing.T) {
...
 	backend := newBackendImplementation(APIBackend, config).(*BackendImplementation)
 
-	// Call with nil context - should not panic
-	backend.logInfof(nil, "test message")
+	// Call with context.TODO() instead of nil
+	backend.logInfof(context.TODO(), "test message")
 
-	assert.Equal(t, "[INFO] test message\n", stdout.String(), "Should handle nil context gracefully")
-	assert.Nil(t, contextLogger.lastContext, "Nil context should be passed as-is")
+	assert.Equal(t, "[INFO] test message\n", stdout.String(), "Should handle context.TODO() gracefully")
```
mbroshi-stripe: This test case doesn't make any sense anymore. I saw there a linter error. Is there a way to suppress that and keep the test as it was?

log_test.go
```diff
@@ -219,12 +226,11 @@ func TestContextLeveledLoggerNilContext(t *testing.T) {
 		},
 	}
 
-	// Should handle nil context gracefully
-	logger.Infof(nil, "test message")
+	// Should handle context.TODO() gracefully
+	logger.Infof(context.TODO(), "test message")
```
mbroshi-stripe: Ditto to previous comment

lmcrean:
appreciate the candid review, i was being overly optimistic about the refactor, let me revert a step and see if i can improve the design while maintaining backwards compatibility

lmcrean:
@mbroshi-stripe  Thanks for your notes on the nil context handling.

I've refactored the logging methods to address your concerns; hopefully the design is now improved. 

Key points at a glance:

- The log helper methods (`logDebugf`, `logInfof`, etc.) now trust the initialization-time detection instead of re-checking types on every call. This removes 16 lines of defensive code.
- Defined a custom `contextKey` type in tests to address SA1029 warnings.
- Kept `contextLogger` and `regularLogger` fields for Performance - detecting the logger type once at initialization is faster than type assertions on every log call. The initialization panic catches invalid loggers anyway.

Lint status:

```
$ go vet ./...     ✓
$ staticcheck      ✓
```

Let me know if this approach makes more sense

lmcrean:
@mbroshi-stripe Ready for another look; thanks for your patience. 

The design has been simplified by removing the cached `contextLogger`/`regularLogger` fields. This addresses your valid concern about complexity not being justified.

Log methods now do runtime type assertions on each call (no stored pointers). The defensive fallback for manual `BackendImplementation` instantiation is also preserved.

All checks pass locally on my end (`gofmt`, `go vet`, `staticcheck`, tests).

log_test.go
```diff
@@ -106,6 +107,127 @@ func TestLeveledLoggerErrorf(t *testing.T) {
 	}
 }
 
+//
+// ContextLeveledLogger
+//
+
+// testContextLogger is a test implementation of ContextLeveledLoggerInterface
+type testContextLogger struct {
```
mbroshi-stripe, line 115: I don't think these tests provide anything, and can be removed. The existing tests in this file test an exported type (`stripe.LeveledLogger`), but these new tests just test the functionality of the `testContextLogger`

stripe.go
```diff
@@ -997,7 +1042,7 @@ func (s *BackendImplementation) UnmarshalJSONVerbose(statusCode int, body []byte
 
 		newErr := fmt.Errorf("Couldn't deserialize JSON (response status: %v, body sample: '%s'): %v",
 			statusCode, bodySample, err)
-		s.LeveledLogger.Errorf("%s", newErr.Error())
+		s.logErrorf(context.Background(), "%s", newErr.Error())
```
mbroshi-stripe, line 1049: I think this is fine for now. I'll make a backlog item to pass `context` into `UnmarshalJSONVerbose` in our next major (it's technically breaking) so we log the request's context in this situation

stripe.go
```diff
@@ -619,7 +628,7 @@ func (s *BackendImplementation) maybeSetTelemetryHeader(req *http.Request) {
 			if err == nil {
 				req.Header.Set("X-Stripe-Client-Telemetry", string(metricsJSON))
 			} else {
-				s.LeveledLogger.Warnf("Unable to encode client telemetry: %v", err)
+				s.logWarnf(context.Background(), "Unable to encode client telemetry: %v", err)
```
mbroshi-stripe: Can you use the `req.Context()` here?

log_test.go
lmcrean: Done. Kept the `testContextLogger` type + its 4 methods (implements `ContextLeveledLoggerInterface`, used by `context_logger_test.go`) and removed the test functions.

stripe.go
lmcrean: OK -- good to know

lmcrean:
@mbroshi-stripe Thanks again for your review, I have addressed those tweaks that were requested. It's ready for another look

stripe.go
```diff
@@ -766,21 +811,21 @@ func (s *BackendImplementation) logError(statusCode int, err error) {
...
-		s.LeveledLogger.Errorf("Error decoding error from Stripe: %v", err)
+		s.logErrorf(ctx, "Error decoding error from Stripe: %v", err)
 	}
 }
 
 func (s *BackendImplementation) handleResponseBufferingErrors(res *http.Response, err error) (io.ReadCloser, error) {
 	// Some sort of connection error
 	if err != nil {
-		s.LeveledLogger.Errorf("Request failed with error: %v", err)
+		s.logErrorf(context.Background(), "Request failed with error: %v", err)
```
mbroshi-stripe: Actually, are you able to use `res.Request.Context()` here? If so, can you add a test around that?

stripe.go
```diff
@@ -797,7 +842,7 @@ func (s *BackendImplementation) handleResponseBufferingErrors(res *http.Response
 	if err == nil {
 		err = s.ResponseToError(res, resBody)
 	} else {
-		s.logError(res.StatusCode, err)
+		s.logError(context.Background(), res.StatusCode, err)
```
mbroshi-stripe: Ditto.

lmcrean:
@mbroshi-stripe Thanks.

`res.Request.Context()` Implemented at both locations. Note that the nil checks are needed because `res` can be nil on connection errors (see existing test `TestHandleResponseBufferingErrors_NilResponse`).

Added tests verifying context propagation and the fallback behavior. All pass.

This is the first use of `res.Request.Context()` in this codebase, it appears a good decision that provides more context than `context.Background()`

stripe.go
```diff
@@ -825,7 +825,11 @@ func (s *BackendImplementation) logError(ctx context.Context, statusCode int, er
 func (s *BackendImplementation) handleResponseBufferingErrors(res *http.Response, err error) (io.ReadCloser, error) {
 	// Some sort of connection error
 	if err != nil {
-		s.logErrorf(context.Background(), "Request failed with error: %v", err)
+		if res != nil && res.Request != nil {
```
mbroshi-stripe: Nit: `res.Request != nil` check is redundant since `context.Background()` is the default value: https://github.com/golang/go/blob/master/src/net/http/request.go#L352-L357

stripe.go
```diff
@@ -842,7 +846,11 @@ func (s *BackendImplementation) handleResponseBufferingErrors(res *http.Response
 	if err == nil {
 		err = s.ResponseToError(res, resBody)
 	} else {
-		s.logError(context.Background(), res.StatusCode, err)
+		if res.Request != nil {
```
mbroshi-stripe: Ditto: `nil`-check is not necessary here.

lmcrean:
> Nit: res.Request != nil check is redundant since context.Background() is the default value: https://github.com/golang/go/blob/master/src/net/http/request.go#L352-L357

This makes sense. I've removed them.

mbroshi-stripe (commented):
Looks good now! Please fix the linter error

lmcrean:
> Please fix the linter error

Fixed. `bash scripts/gofmt.sh check` passing locally.

lmcrean:
@mbroshi-stripe I am pleased to hear that! Thank you for the guidance!

Issue #2168, opened 2025-10-08
Using stripe.Client, List methods no longer expose pagination related properties


chanioxaris:
### Describe the bug

For the List methods, when using the new `stripe.Client` instead of the now-deprecated resource-specific clients (e.g., `customer.Client{}`), the SDK no longer appears to expose certain pagination related properties, such as `has_more`

### To Reproduce

Now, doesn't seem to be a way to access the pagination related properties but only iterate over the results

```go
params := &stripe.CustomerListParams{
	ListParams: stripe.ListParams{
		Limit:  stripe.Int64(20),
		Single: true, // fetch single page and not iterate
	},
}
for customer, err := range sc.V1Customers.List(context.TODO(), params) {
	// handle err
	// handle customer
}

// there is currently no way to access the 'has_more' property
```

### Expected behavior

I would expect a way to have access to the pagination related properties, such as `has_more`

### Code snippets

Previously, it was possible to access the pagination related properties via the Meta() method

```Go
params := &stripe.CustomerListParams{
	ListParams: stripe.ListParams{
		Limit: stripe.Int64(20),
	},
}
i := sc.Customers.List(params)
for _, customer := range i.CustomerList().Data {
	// handle customer
}
if err := i.Err(); err != nil {
	// handle err
}

// there used to be a way to access the 'has_more' property
print(i.Meta().HasMore)
```

### OS

macOS

### Go version

Go 1.25.0

### stripe-go version

v82.4.0

### API version

2025-07-30.basil

mbroshi-stripe:
👋 Hi @chanioxaris ! Thanks for bring up this issue! You are right that there is currently no way to access `HasMore` through `stripe.Client`. Can you tell me more about your use case and how you are you using that property?

chanioxaris:
Hey @mbroshi-stripe, as part of our company pricing, we fetch and preview the invoices for each customer in our Dashboard. So we need a way to let the UI know whether there is more data to fetch or not, a pretty straightforward pagination use case

mbroshi-stripe:
Yep, makes a lot of sense! Thanks for added context! We are working on a fix for this and I'll circle back if we need any additional info. In the meantime, although it is marked as `deprecated`, a workaround is to use the old `client.API` pattern.

mbroshi-stripe:
[...]

PR #2179, opened 2025-10-14
[Breaking] Update `List` and `Search` methods with `stripe.Client` to return a `struct`

mbroshi-stripe:
### Why?

This PR address issues like https://github.com/stripe/stripe-go/issues/2168 where users want to manually paginate instead of auto-paginate after a `List` call using `stripe.Client`.  Although there are possible workarounds with the current design, they are quite hacky  (e.g. requesting an additional item from the API to determine if `HasMore` is true).

After much internal debate, we decided the best path forward to support both the manual and automatic pagination use cases is to introduce a breaking change where `List` returns a `struct` instead of a `func`. You can access the underlying data on the `struct` with `Data()` and `Meta()` methods, or call `.All(ctx)` to get back to the auto-pagination behavior. This is in line with how our other language SDKs handle `List`.

So, automated pagination will look like this:
```go
for c, err := range sc.V1Customers.List(ctx, nil).All(ctx) {
	// handle err
	// do something
}
```
and manual pagination will look like:
```go
list := sc.V1Customers.List(ctx, nil)
if list.Err() {
	// handle err
}
doSomething(list.Data(), list.Meta())
```
### What?

- ⚠️ `List` and `Search` methods using `stripe.Client` now return a `struct` instead of `Seq2`. 
- All of the logical changes are in `iter.go` and `search_iter.go`, where the new types and methods are introduced, and internal pagination logic depends on `page()` instead of `next()`.
- Exports the type `v1List` --> `V1List` with a `v1Page` struct that represents the response from the API.
- Deprecates the `NewV2List` function and `Fetch` type which were really meant to be used only internally, and not exported.

## Changelog
- ⚠️ `List` and `Search` methods using `stripe.Client` now return a `struct` instead of `Seq2`. This is a backwards incompatible change, and you will need to add an additional call to `.All(ctx)` in your `for` loop. E.g.
```diff
-for c, err := range sc.V1Customers.List(ctx, nil) {
+for c, err := range sc.V1Customers.List(ctx, nil).All(ctx) {
	// handle err
	// do something
}
```
For manual pagination use cases, you can access the API call's `error` by calling `list.Err()`, a page's data using `list.Data()`, and its metadata by calling `list.Meta()`.

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -902,24 +902,25 @@ func (s *BackendImplementation) Do(req *http.Request, body *bytes.Buffer, v Last
 	resBody := result.([]byte)
 	s.logDebugf(req.Context(), "Response: %s", string(resBody))
 
-	err = s.UnmarshalJSONVerbose(res.StatusCode, resBody, v)
+	err = s.unmarshalJSONVerbose(req.Context(), res.StatusCode, resBody, v)
 	v.SetLastResponse(newAPIResponse(res, resBody, requestDuration))
 	return err
 }
 
 // ResponseToError converts a stripe response to an Error.
 func (s *BackendImplementation) ResponseToError(res *http.Response, resBody []byte) error {
+	ctx := contextFromResponse(res)
 	var raw rawError
 	if s.Type == ConnectBackend {
 		// If this is an OAuth request, deserialize as Error because OAuth errors
 		// are a different shape from the standard API errors.
 		var topLevelError Error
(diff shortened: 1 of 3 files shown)


prathmesh-stripe:
Is it possible to do a diff like statement in the changelog: That would be easier to understand for people going through the changelog to understand the required change. 

```diff
-for c, err := range sc.V1Customers.List(ctx, nil) {
+for c, err := range sc.V1Customers.List(ctx, nil).All(ctx) {
```

Issue #2181, opened 2025-10-16
`V2CoreAccounts.Retreive` does not properly serialize `Include` parameter


ReeceJones:
### Describe the bug

When using the `sc.V2CoreAccounts.Retreive` function, the `Include` attribute of the `params` parameter is not properly serialized. This causes the API request to be made without the `include` query parameters, which results in none of the `included` resources being returned by the API.

This appears to be a bug specific to how the request parameters are encoded for 2 reasons:
1. In the stripe dashboard, the request shows an empty array passed for the `included` parameter.
2. I have verified that as a workaround, `params.AddExtra` can be used to ensure that the request is properly made to the stripe API.

### To Reproduce

The following code makes a request to the Stripe API without the `include` parameter:
```go
account, err := stripeClient.V2CoreAccounts.Retrieve(context, accountID, &stripeGo.V2CoreAccountRetrieveParams{
	Include: []*string{
		stripeGo.String("configuration.customer"),
		stripeGo.String("configuration.merchant"),
		stripeGo.String("configuration.recipient"),
		stripeGo.String("requirements"),
	},
})
```

### Expected behavior

It is expected that when the `Include` parameter is provided, the client properly serializes the request and the API returns the included resources. However, the request is not properly serialized and so the included resources are not returned.

### Code snippets

```Go

```

### OS

macOS

### Go version

1.24.4

### stripe-go version

v83.1.0-beta.2

### API version

2025-09-30.preview

### Additional context

As a workaround, you can use the `params.AddExtra` method to construct a request that is properly serialized:
```go
params := &stripeGo.V2CoreAccountRetrieveParams{}
params.AddExtra("include", "configuration.customer")
params.AddExtra("include", "configuration.merchant")
params.AddExtra("include", "configuration.recipient")
params.AddExtra("include", "requirements")

account, err := stripeClient.V2CoreAccounts.Retrieve(context, accountID, params)
```

helenye-stripe:
Hi! @prathmesh-stripe  worked on fixing this. Thanks for the report and we'll report back when it has been released.

prathmesh-stripe:
@ReeceJones The fix for this has been released as part of [GA](https://github.com/stripe/stripe-go/releases/tag/v83.0.2), [Beta](https://github.com/stripe/stripe-go/releases/tag/v83.1.0-beta.3), [Private Preview](https://github.com/stripe/stripe-go/releases/tag/v83.1.0-alpha.5) releases this week.

PR #2183, opened 2025-10-20
Added flat_array structtag to V2 APIs to fix query param encoding

prathmesh-stripe:
### Why?
See V2 form encoding problem in Go reported in https://github.com/stripe/stripe-go/issues/2181

### What?
- In V2(url query params) we want the arrays to be encoded flat as `foo=bar&foo=baz`. 
- In V1, arrays are still index-encoded as `foo[0]=bar&foo[1]=baz`.

### Test Plan
- [x] All changes in PR are covered by tests
- [ ] Failures and edge cases tested
- [x] Manually tested any SDK changes

### See Also
https://github.com/stripe/stripe-go/issues/2181

## Changelog
* Fixes URL serialization for array query parameters in V2 APIs.

### DIFF
--- a/form/form.go
+++ b/form/form.go
@@ -418,7 +418,6 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 	// `form`
 	se := &structEncoder{}
 
-	// validation step
 	for i := 0; i < t.NumField(); i++ {
 		reflectField := t.Field(i)
 		tag := reflectField.Tag.Get(tagName)
@@ -441,6 +440,7 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 		fldTyp := reflectField.Type
 		fldKind := fldTyp.Kind()
 
+		// validate that fields are of expected types and have expected annotations
 		if Strict && options != nil {
 			if options.Empty && fldKind != reflect.Bool {
 				panic(fmt.Sprintf(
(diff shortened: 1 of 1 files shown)


form/form.go
```diff
@@ -408,6 +418,7 @@ func makeStructEncoder(t reflect.Type) *structEncoder {
 	// `form`
 	se := &structEncoder{}
 
+	// validation step
```
mbroshi-stripe: Couple of things:
1. Did you remove the extra validation we added for `flat_array` on purpose?
2. This comment belongs above line 444. Can you move it there, and expand on this comment a little? Maybe a sentence like `// validate that fields are of expected types and have expected annotations`

prathmesh-stripe: 1. Yes, it was also running on v1 APIs which is not we want. 
2. Sure

prathmesh-stripe: 1. Added extra validation back. The test failure was due to option nil check.

v2/core/eventdestination/client_test.go
```diff
@@ -90,7 +90,7 @@ func TestEventDestinationGet(t *testing.T) {
 		assert.NoError(t, err)
 		return data
 	}, func(t *testing.T, r *http.Request) {
-		assert.Equal(t, r.URL.Query().Get("include[0]"), "webhook_endpoint.url")
+		assert.Equal(t, r.URL.Query().Get("include"), "webhook_endpoint.url")
```
prathmesh-stripe, line 93: This was changed because of a failing test.

prathmesh-stripe:
> This looks like it has some unrelated generated changes pulled in; mostly descriptions and tests, but is that intentional?
@jar-stripe 

Those were unintentional but they were generated using the last GA release spec(v2028). They were comment changes so didn't look too closely at them.

PR #2187, opened 2025-10-20
Generated code post merge from master

prathmesh-stripe:
### Why?
Regenerated code to add flat_array fix to beta

### DIFF
--- a/v2billing_collectionsetting_params.go
+++ b/v2billing_collectionsetting_params.go
@@ -13,7 +13,7 @@ type V2BillingCollectionSettingListParams struct {
 	Limit *int64 `form:"limit" json:"limit,omitempty"`
 	// Only return the settings with these lookup_keys, if any exist.
 	// You can specify up to 10 lookup_keys.
-	LookupKeys []*string `form:"lookup_keys" json:"lookup_keys,omitempty"`
+	LookupKeys []*string `form:"lookup_keys,flat_array" json:"lookup_keys,omitempty"`
 }
 
 // Controls emails for when the payment is due. For example after the invoice is finilized and transition to Open state.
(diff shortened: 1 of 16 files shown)


v2billing_collectionsetting_params.go
```diff
@@ -412,17 +412,17 @@ type V2BillingCollectionSettingUpdatePaymentMethodOptionsSEPADebitParams struct{
 // Provide filters for the linked accounts that the customer can select for the payment method.
 type V2BillingCollectionSettingUpdatePaymentMethodOptionsUSBankAccountFinancialConnectionsFiltersParams struct {
 	// The account subcategories to use to filter for selectable accounts.
-	AccountSubcategories []*string `form:"account_subcategories" json:"account_subcategories,omitempty"`
+	AccountSubcategories []*string `form:"account_subcategories,flat_array" json:"account_subcategories,omitempty"`
```
prathmesh-stripe, line 415: @mbroshi-stripe What do you think about this?

mbroshi-stripe: This looks right to me. Do you have any worries about it?

prathmesh-stripe: Sorry this looks fine. Let me tag you on another place

v2coreaccounts_person_params.go
```diff
@@ -858,12 +863,13 @@ type V2CoreAccountsPersonUpdateScriptNamesParams struct {
...
-	Params    `form:"*"`
+	Params `form:"*"`
+	// The Account the Person is associated with.
 	AccountID *string `form:"-" json:"-"` // Included in URL
 	// Additional addresses associated with the person.
-	AdditionalAddresses []*V2CoreAccountsPersonUpdateAdditionalAddressParams `form:"additional_addresses" json:"additional_addresses,omitempty"`
+	AdditionalAddresses []*V2CoreAccountsPersonUpdateAdditionalAddressParams `form:"additional_addresses,flat_array" json:"additional_addresses,omitempty"`
 	// Additional names (e.g. aliases) associated with the person.
-	AdditionalNames []*V2CoreAccountsPersonUpdateAdditionalNameParams `form:"additional_names" json:"additional_names,omitempty"`
+	AdditionalNames []*V2CoreAccountsPersonUpdateAdditionalNameParams `form:"additional_names,flat_array" json:"additional_names,omitempty"`
```
prathmesh-stripe, line 872: @mbroshi-stripe What do you think about this? Non-string type getting typed as flat-array? It is part of UpdateParams so it will never be form encoded.

mbroshi-stripe: Interesting. I guess the most careful change would be to only add `flat_array` for `GET` methods params. And, TBH, I'm not sure about non-string types. You can either ask about it in the Slack channel, and/or check how other languages handle non-primitive array types.

mbroshi-stripe: Discussed offline. This will never be `form`-encoded so it's a no-op and fine to leave in

Issue #2211, opened 2025-11-12
Segmentation fault in `handleResponseBufferingErrors`


daniel-weisse:
### Describe the bug

The `RawRequest` method of `BackendImplementation` uses the `handleResponseBufferingErrors` method to inspect errors returned by the Stripe backend to the client.
The inputs to `handleResponseBufferingErrors` are a `*http.Response` and `error` directly returned by `http.Client.Do`.
In case `Do` returns an error, `*http.Response` may be nil.
Since `handleResponseBufferingErrors` tries to return `*http.Response.Body` in case `error` is not `nil`, this causes a nil pointer dereference:

https://github.com/stripe/stripe-go/blob/5d995960d20e34474f00cd2ce7a6871180c4d04c/stripe.go#L780-L785

See the stack trace for more details:
```
[ERROR] Request failed with error: Post "https://meter-events.stripe.com/v2/billing/meter_event_stream": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x1f2ca96]

goroutine 160 [running]:
github.com/stripe/stripe-go/v83.(*BackendImplementation).handleResponseBufferingErrors(0xc0000e9aa0, 0x0, {0x2a50c20, 0xc002bae600})
        github.com/stripe/stripe-go/v83@v83.1.0/stripe.go:784 +0xf6
github.com/stripe/stripe-go/v83.(*BackendImplementation).RawRequest.func1(0x12a065d144?, {0x2a50c20?, 0xc002bae600?})
        github.com/stripe/stripe-go/v83@v83.1.0/stripe.go:500 +0x2d
github.com/stripe/stripe-go/v83.(*BackendImplementation).requestWithRetriesAndTelemetry(0xc0000e9aa0, 0xc00045e280, 0xc004520c90, 0xc00105f4d0)
        github.com/stripe/stripe-go/v83@v83.1.0/stripe.go:726 +0x465
github.com/stripe/stripe-go/v83.(*BackendImplementation).RawRequest(0xc0000e9aa0, {0x26f5a35, 0x4}, {0x2722a0b, 0x1e}, {0xc000188580, 0x143}, {0xc0001863c0, 0x136}, 0xc001180c00)
        github.com/stripe/stripe-go/v83@v83.1.0/stripe.go:504 +0x426
[...]

PR #2212, opened 2025-11-12
Fix segfault in `BackendImplementation.handleResponseBufferingErrors`

daniel-weisse:
### Why?
`BackendImplementation.handleResponseBufferingErrors` may segfault due to a nil pointer dereference.

### What?
- Fix a segmentation fault in `BackendImplementation.handleResponseBufferingErrors` by returning `io.NopCloser(&bytes.Buffer{})` in case `res` is nil

### See Also
- See https://github.com/stripe/stripe-go/issues/2211

### DIFF
--- a/rawrequest/client_test.go
+++ b/rawrequest/client_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
 	assert "github.com/stretchr/testify/require"
 	stripe "github.com/stripe/stripe-go/v83"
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -781,7 +781,10 @@ func (s *BackendImplementation) handleResponseBufferingErrors(res *http.Response
 	// Some sort of connection error
 	if err != nil {
 		s.LeveledLogger.Errorf("Request failed with error: %v", err)
-		return res.Body, err
+		if res != nil {
+			return res.Body, err
+		}
+		return io.NopCloser(&bytes.Buffer{}), nil
```
mbroshi-stripe: I appreciate the very careful coding here, but I double-checked that it's safe to just do the simpler:
```go
if err != nil {
    s.LeveledLogger.Errorf("Request failed with error: %v", err)
    return nil, err
}
```

mbroshi-stripe:
@daniel-weisse I also wrote some unit tests locally. Do you mind if I push them to your PR?

daniel-weisse:
>Do you mind if I push them to your PR?

Please go ahead

PR #2218, opened 2025-11-18
Devsdk 2822 again

xavdid-stripe:


xavdid-stripe:
thought this would fix some CI weirdness, but it didn't, so i'm closing

PR #2240, opened 2025-12-18
Add HTTP response code to V2 Error

mbroshi-stripe:
### Why?

We did an audit of V2 error-handling across our SDKs, and realized that in V2 APIs, Go users do not receive the HTTP Status Code like they do with V1 APIs. That makes error handling harder (e.g. handling a 429 status code).

### What?

- Adds `HTTPStatusCode` and `RequestID` to generic errors in V2.
- Slight improvement to the error message if the error message body from the server is not in the expected shape.

## Changelog
- Adds `HTTPStatusCode` and `RequestID` fields to `V2RawError`

### DIFF
--- a/error.go
+++ b/error.go
@@ -453,9 +453,9 @@ type V2RawError struct {
 	Type       *ErrorType `json:"type,omitempty"`
 	Message    string     `json:"message"`
 	UserMesage *string    `json:"user_message,omitempty"`
-	// HTTPStatusCode is the HTTP status code of the response.
+	// HTTPStatusCode is the HTTP status code of the response returned by Stripe.
 	HTTPStatusCode int `json:"status,omitempty"`
-	// RequestID is the request ID of the response.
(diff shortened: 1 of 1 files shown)


error.go
```diff
@@ -453,6 +453,10 @@ type V2RawError struct {
 	Type       *ErrorType `json:"type,omitempty"`
 	Message    string     `json:"message"`
 	UserMesage *string    `json:"user_message,omitempty"`
+	// HTTPStatusCode is the HTTP status code of the response.
+	HTTPStatusCode int `json:"status,omitempty"`
+	// RequestID is the request ID of the response.
```
prathmesh-stripe: This phrase is a bit confusing to read. What do you think about: 

https://github.com/stripe/stripe-java/blob/ecb791f567f85377f60b272eb9f7a4d91bfb67b7/src/main/java/com/stripe/net/AbstractStripeResponse.java#L68

PR #2241, opened 2025-12-25
Copy API_VERSION to CODEGEN_VERSION

ramya-stripe:
### Why?
The API_VERSION file stores the commit details of the codegen repo which was used to generate v2 APIs in the SDK. Until October this year, this file stored the API version instead. To avoid misleading users and having them depend on this file in any manner, we want to rename this file.

### What?
- Copy API_VERSION to CODEGEN_VERSION instead of renaming just yet
- After our codegen automations have been updated to use the new file, the old one will be deleted


CODEGEN_VERSION
```diff
@@ -0,0 +1 @@
+ee56c4b73d856e48409065f8d7e7c82c2e805c38
```
prathmesh-stripe, line 1: This SHA is same as [.NET](https://github.com/stripe/stripe-dotnet/pull/3279/files) but different from others:
* https://github.com/stripe/stripe-java/pull/2132/files
* https://github.com/stripe/stripe-node/pull/2529/files
* https://github.com/stripe/stripe-ruby/pull/1750/files

ramya-stripe: This file is meant to be a copy of API_VERSION - It is entirely possible for different versions of the codegen to have produced the code in different languages based on whether any language specific fix was made in codegen

PR #2254, opened 2026-01-13
Remove golang.org/x/net/http2 dependency

mbroshi-stripe:
### Why?

We currently depend on `golang.org/x/net/http2`, which has some open CVEs. This dependency was only used for testing, and is no longer needed since the required HTTP2 configuration is available in the standard library.

### What?

- Removes the dependency on `golang.org/x/net/http2`

## Changelog
- Removes the dependency on `golang.org/x/net/http2`

### DIFF
--- a/testing/testing.go
+++ b/testing/testing.go
@@ -69,6 +69,13 @@ func init() {
 			"it running? Please see README for setup instructions.\n", port, err)
 		os.Exit(1)
 	}
+
+	if resp.ProtoMajor != 2 {
+		fmt.Fprintf(os.Stderr, "Expected HTTP/2 connection to stripe-mock, but got %s. "+
+			"This may indicate ForceAttemptHTTP2 is not set correctly.\n", resp.Proto)
+		os.Exit(1)
+	}
+
 	version := resp.Header.Get("Stripe-Mock-Version")
 	if version != "master" && compareVersions(version, MockMinimumVersion) > 0 {
 		fmt.Fprintf(os.Stderr, "Your version of stripe-mock (%s) is too old. The "+


testing/testing.go
```diff
@@ -52,24 +51,13 @@ func init() {
...
-	//
-	// See the incorrectly closed bug report here:
-	//
-	//     https://github.com/golang/go/issues/20645
-	//
-	err := http2.ConfigureTransport(transport)
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Failed to initialize HTTP/2 transport: %v\n", err)
-		os.Exit(1)
+		ForceAttemptHTTP2: true,
```
jar-stripe, line 59: just a sanity check, that ticket looks like it targets an ancient version of go and it looks like its been closed (maybe in error?).  Is this still needed, and is there any cost to doing it?

mbroshi-stripe: You're 100% right. That was the ticket in the original comment, and I accidentally kept it. I updated the comment.

mbroshi-stripe:
> but I would imagine our normal http requests have the same issue, right?

No, this only affects HTTP transport when TLS is disabled. It's only disabled in testing against stripe-mock. For backwards compatibility, the Go standard library forces you to explicitly enable HTTP/2 when TLS is disabled. In our regular HTTP request we are of course not disabling TLS, so HTTP/2 is on by default.

jar-stripe:
> > but I would imagine our normal http requests have the same issue, right?
> 
> No, this only affects HTTP transport when TLS is disabled. It's only disabled in testing against stripe-mock. For backwards compatibility, the Go standard library forces you to explicitly enable HTTP/2 when TLS is disabled. In our regular HTTP request we are of course not disabling TLS, so HTTP/2 is on by default.

ohhh that makes sense, and explains why the issue was an issue 😅

PR #2263, opened 2026-01-29
Add helpers in preparation for BatchJobs public preview support

mbroshi-stripe:
### Why?
Stripe's V2 Batch Jobs API allows performing bulk operations (e.g., updating thousands of customers or subscriptions) in a single API call. The client needs to serialize individual request params into JSONL lines locally before submitting them. Additionally, `BatchJobEndpointPath` is a `StringEnum` — the first `StringEnum` type used in V2 JSON-serialized requests — which requires dedicated JSON converters to serialize as a plain string rather than an object. While this feature is still only in public preview, we need certain prerequisite changes to be made in GA to handle public and private preview features.

### What?
- Adds `newUUIDV4` method to create unique IDs for batch jobs. This can eventually be replaced by Go's stdlib `uuid` once that's introduced (Go 1.27?)
- Adds `stripe.Pointer` and `stripe.PointerValue` generic counterparts to `stripe.String`, `stripe.Int64`, etc. These can all be removed once we only support Go 1.26+, which supports `new`


batch_helpers.go
```diff
@@ -0,0 +1,23 @@
+//
+//
+// File generated from our OpenAPI spec
```
xavdid-stripe: i'm not sure there's much advantage to this being a generated file, but it's not a hill I'd die on.

mbroshi-stripe: I agree. I'll move it out.

PR #2271, opened 2026-02-13
extend ChargePaymentMethodDetailsCardPresent with reader and location

artur-chopikian:
### Why?
When processing tap-to-pay (or physical terminal) payments, we need to identify which terminal reader and location were used for each transaction. Currently, our ChargePaymentMethodDetailsCardPresent struct does not include these fields, making it impossible to trace a payment back to a specific terminal device.

### What?
- Added Reader and Location fields to ChargePaymentMethodDetailsCardPresent to capture the terminal reader ID

### See Also
No more info, the PR is pretty simple.

### DIFF
--- a/charge.go
+++ b/charge.go
@@ -1226,6 +1226,8 @@ type ChargePaymentMethodDetailsCardPresent struct {
 	IncrementalAuthorizationSupported bool `json:"incremental_authorization_supported"`
 	// The last four digits of the card.
 	Last4 string `json:"last4"`
+	// ID of the [location](https://docs.stripe.com/api/terminal/locations) that this transaction's reader is assigned to.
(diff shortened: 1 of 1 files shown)


jar-stripe:
Hi @artur-chopikian , thank you for sending this!  Unfortunately we cannot accept this PR because it makes changes to a generated file.

If this field is in the Stripe API but missing from our SDK, can you refile this as an [GitHub issue](https://github.com/stripe/stripe-go/issues/new/choose)?  We can then look at why this field isn't being generated properly.

artur-chopikian:
@jar-stripe, thank you! I created an issue for that: https://github.com/stripe/stripe-go/issues/2274

PR #2281, opened 2026-02-27
Removed impossible nil checks

CR2D2:


xavdid-stripe:
Hi @CR2D2, thanks for the PR!

Unfortunately, the files you edited are part of our generated code, which means your changes would be overwritten next time our SDKs were generated.

We can take a look at extraneous nil checks in our generation code, but sometimes they're applicable for some services but not all and it's more trouble than its worth to pull them.

If this is something you feel strongly about, we'd appreciate an issue detailing what problems the checks are causing.

Thanks!

CR2D2:
Understood. Generating incorrect code is a bug, not a feature.

PR #2352, opened 2026-05-05
Add EventNotificationHandler (private preview)

xavdid:
### Why?

The code for `EventNotificationHandler` has been in the beta branch since late last year (added in https://github.com/stripe/stripe-go/pull/2209), but hasn't been available in private-preview. This PR rectifies that.

Note: there's no net-new code in this PR. Just files / changes copied from already-approved sources.

### What?

- generate SDK to copy new static files into private-preview (adds "is generated" header to existing files)
- do manual tweaks to wire last bits together (basically copied from original beta PR)

### See Also

- https://github.com/stripe/stripe-go/pull/2209

### DIFF
--- a/event_notification.go
+++ b/event_notification.go
@@ -1,3 +1,9 @@
+//
+//
+// File copied from our code generator; changes here will be overwritten.
+//
+//
+
 package stripe
 
 import (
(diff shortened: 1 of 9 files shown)


mbroshi-stripe (approved):
There are linter failures that I think you'll have to fix in codegen.

PR #2356, opened 2026-05-18
Handle nil params in usage backend raw calls

oab24413gmai:
UsageBackend.CallRaw and CallMultipart can receive nil Params from callers using raw request paths. This avoids dereferencing nil while preserving usage propagation when params are present, and adds focused regression coverage.

Verification:
- go test . -run TestUsageBackend_NilParams -count=1


mbroshi-stripe:
@oab24413gmai please sign our CLA

PR #2360, opened 2026-05-27
Convert linting from staticcheck to golangci-lint

mbroshi-stripe:
### Why?

golangci-lint bundles staticcheck, go vet, ineffassign, and unused into a single tool with a declarative config file. It also provides a maintained GitHub Action that handles caching automatically, to greatly speed up linting time in CI.

* [Attempt 1](https://github.com/stripe/stripe-go/actions/runs/26521248267/attempts/1?pr=2360): 4m10s
* [Attempt 2](https://github.com/stripe/stripe-go/actions/runs/26521248267/attempts/2?pr=2360): 36s

### What?
- Replace staticcheck with golangci-lint
- golangci-lint includes some new linters, so to keep this PR scoped:
  - Disables `errcheck` for now (production cases in `stripe.go`/`log.go` need per-case review — tracked as follow-up)
  - Globally disables `ST1005`/`ST1020`/`ST1021` (comment-format rules; all violations are in generated resource files that cannot be edited)
  - Exact-file exclusions for `SA1019` (deprecated client package, 12 files), `ST1000`/`ST1001`/`ST1003` (4 files), and `QF` quickfix hints (6 files) — each with a TODO comment pointing to follow-up work
  - Per-file allowlisting means any _new_ file violating a suppressed check will still fail CI
- Fix three pre-existing lint issues surfaced by the new linters:
  - `ineffassign`: dead `err` assignment in `v2/core/eventdestination/client_test.go`
  - `unused`: two unused consts in `file/client_test.go`
  - `unused`: unused `parseID` var in `example/deserialization_test.go`

Follow-up work tracked via TODO comments in `.golangci.yml`:
- Re-enable `errcheck` after reviewing `stripe.go` and `log.go` call sites
- Re-enable `ST1005`/`ST1020`/`ST1021` once code generator emits compliant comments
- Migrate `SA1019` files from deprecated `client.New` to `stripe.NewClient`
- Fix `QF` quickfix suggestions (strings.ReplaceAll, time.Equal, etc.)


jar-stripe:
Nice one!  It was't fully clear from your PR description: can you confirm that the linters that you disabled are all net new?  Or does this leave us temporarily with fewer lint protections than before?  I think the latter is fine to merge with the follow up, it would just be good to call out where we are after the PR merges vs before the PR.

mbroshi-stripe:
> Nice one! It was't fully clear from your PR description: can you confirm that the linters that you disabled are all net new? Or does this leave us temporarily with fewer lint protections than before? I think the latter is fine to merge with the follow up, it would just be good to call out where we are after the PR merges vs before the PR.

Confirmed. This adds only net-new checks, but disables some of the new checks (e.g. `errcheck`) that are failing at this time. I intend to follow-up with some lint fixes so we can enable the ones I'm disabling there.

.github/workflows/ci.yml
```diff
@@ -39,9 +39,26 @@ jobs:
         with:
           go-version: "1.26"
 
+      - name: Cache Go tools
+        id: cache-go-tools
+        uses: actions/cache@v5
+        with:
+          path: ~/go/bin
+          key: go-tools-${{ runner.os }}-goimports-v0.39.0
```
jar-stripe: small suggestion to add an "us" controlled value here, e.g. v1-go-tools-${{ runner.os }}-goimports-v0.39.0 or something like that.  I'm not sure about Actions but CI cache in other systems is designed to be immutable and so it can be useful to have a mechanism to force a cache clear by changing v1 to v2 in the the key.

jar-stripe (approved):

PR #2361, opened 2026-05-28
Unsuppress golangci-lint checks

mbroshi-stripe:
### Why?
Follow-up to #2360 (golangci-lint migration), which replaced staticcheck with golangci-lint. The latter has a superset of linters, some of which were failing so we suppressed them temporarily. This PR unsuppresses most of the new checks.

### What?
Unsuppresses and fixes almost all golangci-lint checks:
- **ST1000**: Add missing package comments to `form`, `mock`, `testing`, and `webhook` packages
- **ST1001**: Replace dot import in `mock/mock.go` with explicit `stripetest` alias
- **ST1003**: Rename `sdkApiVersion`/`eventApiVersion` → `sdkAPIVersion`/`eventAPIVersion` in `webhook/` to comply with Go initialisms convention
- **QF**: Apply staticcheck quickfix suggestions across 6 files:
  - `strings.ReplaceAll` instead of `strings.Replace(..., -1)`
  - De Morgan's law on double-negation in `stripe.go`
  - `fmt.Fprintf` instead of `Write([]byte(fmt.Sprintf(...)))`
  - `time.Time.IsZero()` instead of `== time.Time{}`
  - Tagged `switch` on `r.URL.RawQuery` in `search_iter_test.go`
  - Remove redundant embedded-field qualifier in `v2_events.go` (regenerated via codegen fix)
- `errcheck`: unchecked errors in `stripe.go`, `log.go`, and test files
- `SA1019`: deprecated `client.New` usage in some files
- `ST1005`: capitalized error strings

Remaining checks left suppressed
- `ST1020`/`ST1021`: Our type and function comments are derived from the OpenAPI spec (same as the descriptions found in https://docs.stripe.com/api). These comments do not adhere to the Go standard of starting comments with the name of the type/function
- `SA1019`: Some tests purposefully depend on the deprecated `client.API`. Once we remove that deprecated type, we can remove those deprecated call sites.

## Changelog
- Lint compliance for golangci-lint, which includes uncapitalizing some `Error` strings. If you are handling errors by matching on error strings, we recommend switching to use `error` types like `CardError` instead.

### DIFF
--- a/v2/core/event/client_test.go
+++ b/v2/core/event/client_test.go
@@ -1,6 +1,7 @@
 package event_test
 
 import (
+	"context"
 	"encoding/json"
 	"net/http"
 	"testing"
@@ -15,7 +16,7 @@ import (
 
 func TestEventGet(t *testing.T) {
 	timeNow := time.Now()
-	params := stripe.V2CoreEventParams{}
+	params := stripe.V2CoreEventRetrieveParams{}
 	testServer, sc := mock.Server(t, string(http.MethodGet), "/v2/core/events/evt_123", nil, func(p *stripe.V2CoreEventParams) []byte {
 		data, err := json.Marshal(stripe.V1BillingMeterErrorReportTriggeredEvent{
 			V2BaseEvent: stripe.V2BaseEvent{
@@ -35,7 +36,7 @@ func TestEventGet(t *testing.T) {
 	})
 	defer testServer.Close()
 
-	event, err := sc.V2CoreEvents.Get("evt_123", &params)
+	event, err := sc.V2CoreEvents.Retrieve(context.TODO(), "evt_123", &params)
 	assert.Nil(t, err)
 	v1BillingEvent, ok := event.(*stripe.V1BillingMeterErrorReportTriggeredEvent)
 	assert.True(t, ok)
@@ -92,7 +93,7 @@ func TestEventAll(t *testing.T) {
 	})
 	defer testServer.Close()
 	cnt := 1
-	sc.V2CoreEvents.All(params)(func(event stripe.V2CoreEvent, err error) bool {
+	sc.V2CoreEvents.List(context.TODO(), params).All(context.TODO())(func(event stripe.V2CoreEvent, err error) bool {
 		assert.Nil(t, err)
 		assert.NotNil(t, event)
 		if cnt == 1 {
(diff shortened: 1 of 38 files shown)


v2/core/event/client_test.go
```diff
@@ -35,7 +36,7 @@ func TestEventGet(t *testing.T) {
 	})
 	defer testServer.Close()
 
-	event, err := sc.V2CoreEvents.Get("evt_123", &params)
+	event, err := sc.V2CoreEvents.Retrieve(context.TODO(), "evt_123", &params)
```
jar-stripe, line 39: sanity check: did we rename a method here?  or are both available but Get is deprecated?

v2/core/eventdestination/client_test.go
```diff
@@ -11,25 +12,24 @@ import (
...
 	stripe "github.com/stripe/stripe-go/v85"
-	"github.com/stripe/stripe-go/v85/client"
 	"github.com/stripe/stripe-go/v85/mock"
 	. "github.com/stripe/stripe-go/v85/testing"
 )
 
 func TestEventDestinationNew(t *testing.T) {
 	timeNow := time.Now()
-	params := &stripe.V2CoreEventDestinationParams{
+	params := &stripe.V2CoreEventDestinationCreateParams{
```
jar-stripe, line 21: similar question to above; are we renaming params or just updating tests to use nondeprecated modern params?

v2/core/event/client_test.go
jar-stripe: Oh, is this
> SA1019: deprecated client.New usage in some files

v2/core/eventdestination/client_test.go
jar-stripe: Is this covered by 

> SA1019: deprecated client.New usage in some files

jar-stripe (approved):
Please check the sanity checks before merging; otherwise, looks good!

v2/core/event/client_test.go
mbroshi-stripe: Yep, I updated `mock.go` to use `stripe.Client`, which forced me to also use it in the v2 test files.

v2/core/eventdestination/client_test.go
mbroshi-stripe: Ditto. It was because of fixing `mock.go`

PR #2370, opened 2026-06-03
Add "source" field to user-agent header

xavdid:
### Why?

Earlier this year, we removed the output of `uname` from the user-agent header for privacy reasons. It over-collected data we usually didn't need.

Since then, our tech support team has reached out asking us to re-add some of the information to help with complex debugging tasks: sometimes they need to help users identify exactly which machine produced an API call.

Rather than just revert the previous change and reintroduce the same privacy issues that caused its removal in the first place, we're taking a more privacy-focused approach. Rather than send all the PII in plaintext, we hash it first. This way it can still be used to confirm the source of an API call if a user generates the hashes themselves, but isn't sensitive in and of itself.

### What?

- add `source` key to user-agent hash. It's contents are an md5 hash of `uname -a`
- add tests

### See Also

- [DEVSDK-3131](https://go/j/DEVSDK-3131)

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -4,7 +4,9 @@ package stripe
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -1810,6 +1812,7 @@ type stripeClientUserAgent struct {
 	Language        string   `json:"lang"`
 	LanguageVersion string   `json:"lang_version"`
 	Platform        string   `json:"platform,omitempty"`
+	Source          string   `json:"source,omitempty"`
 }
 
 // requestMetrics contains the id and duration of the last request sent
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -1877,6 +1881,15 @@ func init() {
 	initUserAgent()
 }
 
+func init() {
+	parts := []string{runtime.GOOS, runtime.GOARCH, runtime.Version()}
+	if h, err := os.Hostname(); err == nil {
+		parts = append(parts, h)
+	}
+	hash := md5.Sum([]byte(strings.Join(parts, " ")))
```
jar-stripe, line 1889: is there any reason to believe this wont be available for some installations?

jar-stripe (approved):
looks good, small question about hash function availability

stripe.go
xavdid: nope!

PR #2371, opened 2026-06-03
Add "source" field to user-agent header

johnsaurabh:
### Why?

The \`uname\` output was previously removed from the user-agent header for
privacy reasons. Stripe's support team has since asked for a way to
identify which machine produced an API call for complex debugging tasks.

Rather than reverting the original change, this takes a privacy-preserving
approach: send an MD5 hash of \`uname -a\` instead of the plaintext output.
Users can reproduce the hash locally to confirm the source of a call without
the SDK ever exposing raw machine information.

### What?

- Add \`source\` field to \`stripeClientUserAgent\`, containing an MD5 hash of \`uname -a\`
- Falls back to hashing \`runtime.GOOS + " " + runtime.GOARCH\` on platforms where \`uname\` is unavailable
- Gated on \`enableTelemetry\`, consistent with the existing \`platform\` field
- Add tests

### See Also

- #2370

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -4,7 +4,9 @@ package stripe
 import (
 	"bytes"
 	"context"
+	"crypto/md5"
 	"crypto/x509"
+	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


jar-stripe:
Closed, duplicate of https://github.com/stripe/stripe-go/pull/2370.  If you would like, please add comments to that PR instead of opening a new one.  Thanks!

PR #2399, opened 2026-07-24
add/adjust event parsing helpers

xavdid:
### Why?

Historically, our webhook parsing logic has been closely tied to signature validation (because the expectation is that users verify webhooks' authenticity before use). But, as the webhook handling experience has grown in complexity over time, our helper methods haven't kept up.

In addition to the classic parse+verify & handle, we also support / encourage :

- using cloud providers (like [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid)), which don't require validation at all
- validating & ACKing an event quickly, then queuing it up for later processing
- unit tests w/ signed test payloads

This gives us 4 use cases that each need a subset of helper methods:

| case                                  | `verify` | `parse` | `parseAndVerify` | `signTestPayload` |
| ------------------------------------- | -------- | ------- | ---------------- | ----------------- |
| process immediately                   |          |         | ✅               | ✅                |
| verify, queue, process later          | ✅       | ✅      |                  | ✅                |
| process immediately (cloud provider)  |          | ✅      |                  |                   |
| queue, process later (cloud provider) |          | ✅      |                  |                   |

To support these workflows, we need all of these methods publicly available in every SDK:

1. `constructEvent(payload, signature, ...)` (parse + verify)
2. `parseEventNotification(payload, signature, ...)` (parse + verify)
3. `constructEventWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
4. `parseEventNotificationWithoutVerification(payload)` (parse, supports both direct webhooks and cloud provider envelopes)
5. `verify(payload, signature, secret, ...)`
6. `signTestPayload(payload, secret)`

The exact naming & arguments vary by language, but each SDK should provide roughly the same capabilities.

> [!IMPORTANT]
> For reviewers: the exact error messages are a little inconsistent between SDKs right now. Everything throws in the right place, but we're still iterating on some of the naming. We're trying out "thin event notification" as a descriptor on some of these surfaces. We're planning to have this all cleaned up by the GA launch of interop events; ignore exact messages for now.

### What?

> [!NOTE]
> these notes generic across all the SDKs; see `Changelog` below for the SDK-specific list of changes

- add missing event handling methods to `StripeClient` & `Webhook` classes, where relevant.
- Add an internal-facing `maybe_extract_from_cloud_provider_envelope` for handling wrapped events without the user having to think about it
- centralize some event parsing logic; adjusting accepted types for some methods
- adjust some error messages
- add tests
- update docstrings
- go: reordered some client methods without changing them

### See Also

- [stripe/stripe-java#2201](https://github.com/stripe/stripe-java/issues/2201)
- [DEVSDK-3085](https://go/j/DEVSDK-3085)
- [DEVSDK-3210](https://go/j/DEVSDK-3210)

## Changelog

- Added methods that return their respective `Event`/`EventNotification` structs without verifying authenticity. Use them when you've previously verified an event (e.g. you verified, put the event in a queue, and are now processing). Supports events from [AWS EventBridge](https://docs.stripe.com/event-destinations/eventbridge) and [Azure Event Grid](https://docs.stripe.com/event-destinations/eventgrid) natively.
  - `stripe.ConstructEventWithoutVerification(payload, opts ...WebhookOption)`
  - `Client.ConstructEventWithoutVerification(payload, opts ...WebhookOption)`
  - `Client.ParseEventNotificationWithoutVerification(payload)`


webhooks.go
```diff
@@ -52,18 +52,17 @@ func ComputeSignature(t time.Time, payload []byte, secret string) []byte {
...
-// the Stripe-Signature header using the specified signing secret. Returns an error
+// ConstructEvent constructs a [snapshot event] from an incoming webhook after
+// verifying its authenticity. To work with a webhook that has already been
+// verified (i.e. one from a cloud provider, an asynchronous queue, or during
+// testing), see [ConstructEventWithoutVerification]. Returns an error
 // if the body or Stripe-Signature header provided are unreadable, if the
 // signature doesn't match, or if the timestamp for the signature is older than
 // DefaultTolerance.
-//
-// NOTE: Stripe will only send Webhook signing headers after you have retrieved
```
jar-stripe, line 61: Was this intentionally removed?  This feels like good context for folks receiving webhooks directly from Stripe

jar-stripe (approved):

webhooks.go
xavdid: It was, but mostly for standardization reasons. happy to re-add (or add everywhere) but was trying to keep this docstring focused.

PR #2415, opened 2026-08-22
Add `.PreHandle` method to `EventNotificationHandler`

xavdid:
### Why?

In final testing, users have expressed interest in being able to centrally run code before any other handler executes. It can be used for centralized logging, event deduplication, and more. I don't want to go full middleware, but this felt like a reasonable way to dip our toe into the waters.

Because users are responsible for [deduplicating events](https://docs.stripe.com/webhooks#handle-duplicate-events), this pre-handle hook (if present) should return a `bool`, signifying whether handling should continue.

If the function returns `true` (or is missing entirely), the handler proceeds like normal (running at-most-one other callback). If it returns `false`, no further callback is run and the original `.handle()` call finishes cleanly.

### What?

- add the `.pre_handle` method
- & associated tests
- update some error messages with correct verbs & consistency
- refactored some internal error handling
- updated example

### See Also

- http://go/j/DEVSDK-3249

### DIFF
--- a/stripe_event_notification_handler.go
+++ b/stripe_event_notification_handler.go
@@ -65,7 +65,7 @@ func NewEventNotificationHandler(client *Client, webhookSecret string, fallbackC
 // synchronously on startup, so it'll only be read after it's done being written.
 func (h *eventNotificationHandlerBase) assertCanRegister() error {
 	if h.hasHandledEvent {
-		return fmt.Errorf("cannot register new event handlers after handling an event. This is indicative of a bug.")
+		return fmt.Errorf("cannot register new callbacks after an event has been handled. This is indicative of a bug.")
 	}
 
 	return nil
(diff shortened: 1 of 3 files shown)


stripe_event_notification_handler.go
```diff
@@ -53,21 +57,47 @@ func NewEventNotificationHandler(client *Client, webhookSecret string, fallbackC
...
 }
 
+// PreHandle registers a hook that runs after Handle() parses the payload but before any other callback fires. Returning `false` from the hook stops handling: no callback runs. Returning an error aborts handling and is returned from the original Handle() call.
+func (h *eventNotificationHandlerBase) PreHandle(callback PreHandleFunc) error {
+	if err := h.assertCanRegister(); err != nil {
+		return err
+	}
+
+	if h.preHandleFunc != nil {
+		return fmt.Errorf("a PreHandle callback is already registered")
```
jar-stripe, line 100: just a small gut check.  Should this be an error or a panic?  How much signal do we want to give a user who tries to call PreHandle multiple times?

jar-stripe (approved):
Looks good!  Left a small comment to consider for error reporting

stripe_event_notification_handler.go
xavdid: My understanding is that you basically never want to panic on behalf of a user- they can panic with an error there if they want, or they can ignore it. Plus all the other `on_*` methods return an error, so it would be weird if this one worked differently

PR #2424, opened 2026-08-28
Harden API requestor code against malicious URLs

xavdid:
### Why?

If an attacker got access to a user's webhook secret (or the user wasn't validating incoming webhook signatures) _and_ a user was calling the built-in `event_notification.fetch_related_object()` method, a malicious webhook payload could cause a user integration to make an authenticated request to an attacker-controlled domain (leaking the secret key)

To solve, we're being more careful during url construction to ensure user input can't be used to send requests to non-`stripe.com` domains (on a per-request basis).

### What?

- add path validation method and call it before the final URL is built
- add tests

### See Also

- [RUN_DEVSDK-2808](https://go/j/RUN_DEVSDK-2808)

### DIFF
--- a/stripe.go
+++ b/stripe.go
@@ -653,16 +653,13 @@ func (s *BackendImplementation) CallMultipart(method, path, key, boundary string
 //
 // The absolute URL is built by concatenating a base URL onto this path, and no
 // base URL ends in a slash (normalizeURL actively strips a trailing one). A path
-// like "@evil.example/v1/x" or ".evil.example/v1/x" would therefore land inside
-// the authority component and send the request -- Authorization header included
-// -- to a host of the path's choosing. Some request paths originate in remote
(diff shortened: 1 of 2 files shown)


stripe.go
```diff
@@ -648,6 +648,25 @@ func (s *BackendImplementation) CallMultipart(method, path, key, boundary string
 	return nil
 }
 
+// validatePath ensures a request path is origin-relative: that it
+// begins with a single "/".
+//
+// The absolute URL is built by concatenating a base URL onto this path, and no
+// base URL ends in a slash (normalizeURL actively strips a trailing one). A path
```
jar-stripe, line 655: this feels like a lot of exposition for validatePath.  is there a better place to put this?

jar-stripe (approved):

PR #2441, opened 2026-09-18
Represent free-form objects as dynamic maps

jar-stripe:
### Why?

Free-form API objects were generated as named empty structs, preventing integrations from reading or sending their arbitrary keys. This updates the affected private-preview fields to use dynamic Go maps.

### What?

- Changes Coupon script configuration request and response fields to `map[string]any`.
- Changes `CryptoOnrampTransactionLimits.Limits` to `map[string]any`.
- Changes PaymentIntent and SetupIntent next-action SDK dictionaries to `map[string]any`.
- Extends the existing Coupon JSON unmarshalling test with nested arbitrary-key coverage.
- Adds JSON unmarshalling coverage for transaction limits.
- Adds a major-level changefile documenting the user-facing Coupon and transaction-limit type changes.

### See Also

- The generated field comments describe the free-form dictionary behavior.

## Configuration

- [ ] `skip-changefile`: This PR is not a user-facing change, so there's no changefile.


coupon_test.go
```diff
@@ -18,12 +18,13 @@ func TestCoupon_UnmarshalJSON(t *testing.T) {
...
 	{
-		v := Coupon{ID: "25OFF"}
+		v := Coupon{ID: "25OFF", Script: &CouponScript{Configuration: map[string]any{"nested": map[string]any{"key": "value"}}}}
 		data, err := json.Marshal(&v)
 		assert.NoError(t, err)
 
 		err = json.Unmarshal(data, &v)
 		assert.NoError(t, err)
 		assert.Equal(t, "25OFF", v.ID)
+		assert.Equal(t, "value", v.Script.Configuration["nested"].(map[string]interface{})["key"])
```
mark-kogan-stripe: Should we use "any" here instead of interface{}?

.hark/changes/2026-09-18_stripe_dynamic-map-types.change.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Represent dynamic objects as maps
+pr_url: https://github.com/stripe/stripe-go/pull/2441
+semver_level: major
+---
+
+* ⚠️ Change type of `CouponScriptParams.Configuration`, `CouponCreateScriptParams.Configuration`, `CouponScript.Configuration`, and `CryptoOnrampTransactionLimits.Limits` from an empty struct to `map[string]any`
```
mark-kogan-stripe, line 7: PaymentIntentNextAction.UseStripeSDK and SetupIntentNextAction.UseStripeSDK were also changed, should we include those here?

crypto_onramptransactionlimits_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	assert "github.com/stretchr/testify/require"
+)
+
+func TestCryptoOnrampTransactionLimits_UnmarshalJSON(t *testing.T) {
+	var limits CryptoOnrampTransactionLimits
+	err := json.Unmarshal([]byte(`{"limits":{"usd":{"card":{"standard":{"remaining":1250}}}}}`), &limits)
+	assert.NoError(t, err)
+
+	usd := limits.Limits["usd"].(map[string]interface{})
```
mark-kogan-stripe, line 15: same question as above (any instead of interface{} ?)

coupon_test.go
jar-stripe: good catch; updated!

mark-kogan-stripe (approved):
LGTM! I had two other questions/nitpicks which I left as comments

